PR #7, opened 2023-08-10
Update OWNERS files

ashwindasr:
For our use case, we need only one OWNERS file in the root of art-tools repo.


ashwindasr:
Doozer and Elliott OWNERS files are referenced in code, so I'm not sure if I should remove them, since its not needed for prow.

Two other files are [rhel-8](https://github.com/openshift-eng/art-tools/blob/main/doozer/doozerlib/cli/ci_transforms/rhel-8/ci-build-root/OWNERS) and[ rhel-9](https://github.com/openshift-eng/art-tools/blob/main/doozer/doozerlib/cli/ci_transforms/rhel-8=9/ci-build-root/OWNERS) ci-build-root OWNERS. Not sure if I should remove that as well.

ashwindasr:
Ahh looks like the [usage](https://github.com/openshift-eng/art-tools/blob/c99b3dec1b93f0c413ee273724dfb1ffbb207e14/doozer/doozerlib/cli/images_streams.py#L975) refers to the repos that we are cloning. So looks like we can safely delete the OWNERS of doozer and elliott

thegreyd (approved):

thegreyd (approved):

PR #8, opened 2023-08-10
[ART-7412] Update CI

ashwindasr:
Migrating unittests and removing Synk tests. synk is being run with org account directly

### Test Run
- [Ran](https://github.com/ashwindasr/art-tools/pull/6) when there was a change in doozer. [Did not run](https://github.com/ashwindasr/art-tools/pull/7) when there was a change in the root level dir.


thegreyd:
Let's have 2 separate workflows like doozer-tests and elliott-tests, which run only if files in their respective dir change

ashwindasr:
Leveraged
```yaml
pull_request:
    paths:
      - 'doozer/**'
```
and similarily for elliott to run only if their code base changes.

thegreyd (approved):

PR #11, opened 2023-08-14
[ART-4761] mirror builder image streams of ocp-private namespace

Ximinhan:
if the upstream_image exists in stream config, mirror it to ocp private by config upstream_image_mirror


sosiouxme (approved):

sosiouxme:
/lgtm
i would refactor out the common code but that can wait for a refactor

PR #12, opened 2023-08-14
[ART-7462] set alignment bug's depends

Ximinhan:
re: [ART-7462](https://issues.redhat.com/browse/ART-7462)
accoring to https://github.com/openshift/cluster-update-keys/pull/47#issuecomment-1664730578
after creating an alignment bug CI will check if it has a bug to depend on if it's not non-GA version
we need to check if a similar bug in higher version exist and try to link it

### DIFF
--- a/doozer/doozerlib/cli/images_streams.py
+++ b/doozer/doozerlib/cli/images_streams.py
@@ -748,6 +748,10 @@ Jira mapping: https://github.com/openshift-eng/ocp-build-data/blob/main/product.
             issue = jira_client.create_issue(
                 fields
             )
+            # check depend issues and set depend to a higher version issue if ture
+            depend_issues = search_issues(f"project={project} AND summary ~ 'Update {major}.{minor+1} {image_meta.name} image to be consistent with ART'")
+            if depend_issues:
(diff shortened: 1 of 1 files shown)


doozer/doozerlib/cli/images_streams.py
```diff
@@ -748,6 +748,10 @@ def search_issues(query):
             issue = jira_client.create_issue(
                 fields
             )
+            # check depend issues and set depend to a higher version issue if ture
+            depend_issues = search_issues(f"Update {major}.{minor+1} {image_meta.name} image to be consistent with ART")
```
vfreex, line 752: Shouldn't this be a jql? Did you test this?

vfreex (changes requested):

doozer/doozerlib/cli/images_streams.py
Ximinhan: updated, test with
```
> jira_client.search_issues("project=OCPBUGS AND summary ~ 'Update 4.14 ose-apiserver-network-proxy image to be consistent with ART'")
[<JIRA Issue: key='OCPBUGS-10187', id='15143318'>]
```

vfreex (approved):
/lgtm

PR #14, opened 2023-08-14
Set target_version value as string

ashwindasr:
Fixes error `response text = {"errorMessages":[],"errors":{"customfield_12323140":"expected 'name' property to be a string"}}`


ashwindasr:
in favor of https://github.com/openshift-eng/art-tools/pull/15

PR #17, opened 2023-08-17
pass build_info into db

Ximinhan:
get build_id from task may failed because the task can be failed but build is success, we can directly pass in the build id if we already have it.


Ximinhan:
testcomment

PR #22, opened 2023-08-29
[ART-7318][ART-5526] introduce doozer operator_index_mode for pre-releases

sosiouxme:
Introduce logic to label bundles differently depending on the group release mode, which can make bundles available either to future releases or in pre-release indexes.

[This hack run](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/lmeyer/job/lmeyer-dev/job/dev-build%252Folm_bundle/38/console) against [this config](https://github.com/openshift-eng/ocp-build-data/pull/3424/files) produced [local-storage-operator-metadata-container-v4.14.0.202308281027.p0.g089e32e.assembly.stream-2](https://brewweb.engineering.redhat.com/brew/buildinfo?buildID=2655730) which I added to a [test advisory](https://errata.devel.redhat.com/advisory/119637) along with dependencies so we can really exercise this new capability.

### DIFF
--- a/doozer/doozerlib/olm/bundle.py
+++ b/doozer/doozerlib/olm/bundle.py
@@ -557,14 +557,27 @@ class OLMBundle(object):
         return files
 
     @property
-    def redhat_delivery_tags(self):
-        versions = '=v{MAJOR}.{MINOR}'.format(**self.runtime.group_config.vars)
+    def operator_index_mode(self):
+        mode = self.runtime.group_config.operator_index_mode or 'ga'  # default when missing
+        if mode in {'pre-release', 'ga', 'ga-plus'}:
+            # pre-release: label for pre-release operator index (unsupported)
+            # ga: label for only this release's operator index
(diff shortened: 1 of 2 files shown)


doozer/doozerlib/olm/bundle.py
```diff
@@ -556,15 +556,28 @@ def list_of_manifest_files_to_be_copied(self):
...
 
+    @property
+    def operator_index_mode(self):
+        mode = self.runtime.group_config.operator_index_mode or 'ga'  # default when missing
+        if mode in {'pre-release', 'ga', 'ga-plus'}:
+            # pre-release: label for pre-release operator index (unsupported)
+            # ga: label for only this release's operator index
+            # ga-plus: label for this release's operator index and future release indexes as well
+            return mode
+        self.runtime.logger.warning(f'{mode} is not a valid group_config.operator_index_mode')
```
Ximinhan, line 567: should we raise an error here? although it was come from group.yaml and unlikely to be invalid

PR #24, opened 2023-08-30
ART-7629 early kernel: only tag when kernel and kernel-rt are of same version

thegreyd:
## Summary
https://issues.redhat.com/browse/ART-7629
Make sure kernel and kernel-rt have the same version before tagging them.

## Test
- Checkout this branch
- `./doozer -g openshift-4.14 config:tag-rpms --dry-run --json`

### DIFF
--- a/doozer/doozerlib/cli/config_tag_rpms.py
+++ b/doozer/doozerlib/cli/config_tag_rpms.py
@@ -101,7 +101,7 @@ class TagRPMsCli:
         koji_api = self._runtime.build_retrying_koji_client()
         koji_api.gssapi_login()
         MAX_BUILDS = 10  # getting 10 latest builds per package should be more than enough
-        builds_to_tag: Dict[str, Set[str]] = {}  # target_tag_name -> set of NVRs
+        builds_to_tag: Dict[str, Dict[str]] = {}  # target_tag_name -> dict of NVRs {nvr_string: build_object}
         builds_to_untag: Dict[str, Set[str]] = {}  # target_tag_name -> set of NVRs
         for entry in rpm_deliveries:
             # For each package, look at builds in stop-ship tag and integration tag,
@@ -110,7 +110,7 @@ class TagRPMsCli:
                 logger.warning("RPM delivery config for package(s) %s doesn't define a target tag. Skipping...",
                                ", ".join(entry.packages))
                 continue
-            builds_to_tag.setdefault(entry.target_tag, set())
+            builds_to_tag.setdefault(entry.target_tag, dict())
             builds_to_untag.setdefault(entry.target_tag, set())
             # Get all builds in stop-ship tag
             logger.info("Getting tagged builds in stop-ship tag %s...", entry.stop_ship_tag)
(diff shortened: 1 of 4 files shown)


thegreyd:
I've tweaked the logic but tests still need updating

doozer/doozerlib/cli/config_tag_rpms.py
```diff
@@ -150,9 +151,34 @@ async def run(self):
...
+
+            for tag, nvr_dict in builds_to_tag.items():
+                if not nvr_dict:
+                    continue
+
+                package_names = {b['name'] for b in nvr_dict.values()}
+                if package_names != {'kernel', 'kernel-rt'}:
+                    raise ValueError(f"Unexpected packages found with enforce_same_version: {package_names}")
+                if len(nvr_dict) != 2:
+                    raise ValueError(f"More than 2 kernel builds found {nvr_dict.keys()}")
```
thegreyd, line 168: I didn't know what to do in this case, do we expect builds > number of packages in each rpm_delivery entry?
I don't think so right @vfreex ?

doozer/doozerlib/cli/config_tag_rpms.py
```diff
@@ -150,9 +151,34 @@ async def run(self):
...
+            if not entry.enforce_same_version:
+                continue
+
+            for tag, nvr_dict in builds_to_tag.items():
+                if not nvr_dict:
+                    continue
+
+                package_names = {b['name'] for b in nvr_dict.values()}
+                if package_names != {'kernel', 'kernel-rt'}:
+                    raise ValueError(f"Unexpected packages found with enforce_same_version: {package_names}")
```
thegreyd, line 166: There can be case when one of them (either kernel or kernel-rt) is missing, so this would change to "Need 1 build of kernel and 1 build of kernel-rt" or something like that

doozer/doozerlib/cli/config_tag_rpms.py
```diff
@@ -150,9 +151,34 @@ async def run(self):
...
+                if len(nvr_dict) != 2:
+                    raise ValueError(f"More than 2 kernel builds found {nvr_dict.keys()}")
+
+                kernel_build = next(b for b in nvr_dict.values() if b['name'] == 'kernel')
+                kernel_rt_build = next(b for b in nvr_dict.values() if b['name'] == 'kernel-rt')
+
+                # e.g. kernel-5.14.0-284.28.1.el9_2, kernel-rt-5.14.0-284.28.1.rt14.313.el9_2
+                kernel_version = f"{kernel_build['version']}-{split_el_suffix_in_release(kernel_build['release'])[0]}"
+                kernel_rt_version = f"{kernel_rt_build['version']}-{kernel_rt_build['release']}"
+                if kernel_version not in kernel_rt_version:
```
vfreex, line 176: The release parts of kernel and kernel-rt are not the same:
kernel: 284.28.1.el9_2
kernel-rt: 284.28.1.rt14.313.el9_2

This is what we do in gen-payload: https://github.com/openshift-eng/doozer/blob/15a672d3dd127feefaa99bbb293b9d20ee53fda1/doozerlib/cli/release_gen_payload.py#L1348

thegreyd: Right I'm not matching them exactly but getting the non-el part of the release (split_el_suffix_in_release) from kernel "284.28.1" and checking if that exists somewhere in kernel_rt release string "284.28.1.rt14.313.el9_2" which works in this case. The gen-payload logic seems more brittle than this? but don't have a strong opinion, I can use that logic if you still think that's better

thegreyd: Updated tests and code

doozer/doozerlib/cli/config_tag_rpms.py
thegreyd: Updated

doozer/doozerlib/cli/config_tag_rpms.py
vfreex: Ah, you made a good point. We don't have to check if their version-releases match or not.

vfreex (approved):
/lgtm
/approve

PR #25, opened 2023-08-31
[ART-7056] deprecate placeholder bug

Ximinhan:
remove the deprecated use of placeholder bug


elliott/elliottlib/cli/create_textonly_cli.py
```diff
@@ -108,10 +108,6 @@ def create_textonly(runtime, errata_type, date, assigned_to, manager, package_ow
         erratum.commit()
         green_prefix("Created new text only advisory: ")
         click.echo(str(erratum))
-        bug = bug_tracker.create_textonly(bug_title, bug_description)
```
vfreex, line 111: This line shouldn't be removed.

locriandev (commented):
/lgtm

elliott/elliottlib/cli/create_textonly_cli.py
Ximinhan: text-only advisory is also RHOSE product which doesn't need a placeholder bug

vfreex: ack

PR #31, opened 2023-09-01
[build-microshift] - Don't skip advisory prep and other job enhancements

thegreyd:
# Summary
- Don't skip advisory prep if build is pinned or PR is not created - do it again (reentrant behavior)
- Check if expected builds are pinned - if not create PR
- Add wait and retry for moving advisory to QE
- Refactor / simplify slack thread communication

Test run: https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fbuild-microshift/579/console

Slack thread: https://redhat-internal.slack.com/archives/C048BG69N8H/p1694115877311999

### DIFF
--- a/pyartcd/pyartcd/pipelines/build_microshift.py
+++ b/pyartcd/pyartcd/pipelines/build_microshift.py
@@ -17,6 +17,7 @@ from doozerlib.util import (brew_arch_for_go_arch,
 from ghapi.all import GhApi
 from ruamel.yaml import YAML
 from semver import VersionInfo
+from tenacity import retry, stop_after_attempt, wait_fixed
 
 from pyartcd import constants, exectools, oc, util, jenkins
 from pyartcd.cli import cli, click_coroutine, pass_runtime
@@ -48,6 +49,8 @@ class BuildMicroShiftPipeline:
         self.force = force
         self._logger = logger or runtime.logger
         self._working_dir = self.runtime.working_dir.absolute()
+        self.slack_client = None
+        self.slack_thread = None
 
         # determines OCP version
         match = re.fullmatch(r"openshift-(\d+).(\d+)", group)
@@ -71,7 +74,6 @@ class BuildMicroShiftPipeline:
         # Make sure our api.ci token is fresh
         await oc.registry_login(self.runtime)
 
-        slack_client = None
         assembly_type = AssemblyTypes.STREAM
         major, minor = util.isolate_major_minor_in_group(self.group)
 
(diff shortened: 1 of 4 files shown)


pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -165,12 +166,13 @@ async def run(self):
                 await self._change_advisory_status()
                 await slack_client.say("Complete preparing microshift advisory.", slack_thread)
         except Exception as err:
-            error_message = f"Error building microshift: {err}\n {traceback.format_exc()}"
-            if assembly_type != AssemblyTypes.STREAM:
-                error_message += "\n@release-artists"
+            slack_message = f"Error preparing microshift advisory: {err}"
```
Ximinhan, line 169: the exception may come from build failure, preparing the advisory is only for the release build, I don't think it needs to be modified here

thegreyd: Ah good catch, made a mistake

thegreyd: Reverted to previous wording

pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -116,15 +118,13 @@ async def run(self):
...
+                await self.slack_say(f":construction: Microshift prep for assembly {self.assembly} :construction:")
+
             if nvrs:
                 self._logger.info("Builds already exist: %s", nvrs)
             else:
-                # Rebases and builds microshift
-                if assembly_type is not AssemblyTypes.STREAM:
-                    slack_client = self.runtime.new_slack_client()
-                    slack_client.bind_channel(self.group)
-                    slack_response = await slack_client.say(f":construction: Build microshift for assembly {self.assembly} :construction:")
```
thegreyd, line 126: Sometimes we were creating slack_thread and sometime not, other times we were relying on slack_thread to indicate something else (like if assembly is stream). So simplify this branching, always create a top level slack message for named assemblies for each run

pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -138,41 +138,61 @@ async def run(self):
...
-                # Create a PR to pin microshift build
+                # Nothing to do for stream builds after this
+                if assembly_type is AssemblyTypes.STREAM:
+                    return
+
+                message = f"microshift for assembly {self.assembly} has been successfully built."
+                await self.slack_say(message)
+
+            # Check if we need create a PR to pin the build
+            pinned_microshift_rpms = util.get_rpm_if_pinned(releases_config, self.assembly, 'microshift')
```
Ximinhan, line 149: we should do the check earlier, if build exist don't need to rebase and build again

pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -138,41 +138,61 @@ async def run(self):
...
+                self._logger.info("Creating PR to pin microshift build: %s", diff)
                 assembly_basis = get_assembly_basis(releases_config, self.assembly)
                 if assembly_basis.brew_event:
                     pr = await self._create_or_update_pull_request(nvrs)
 
-            # Sends a slack message
-            message = f"microshift for assembly {self.assembly} has been successfully built."
-            if pr:
-                message += f"\nA PR to update the assembly definition has been merged: {pr.html_url}"
+                message = (f"PR to pin microshift build to the {self.assembly} assembly has been merged:"
```
Ximinhan, line 157: need to under the if to make sure pr exist

pyartcd/pyartcd/pipelines/build_microshift.py
thegreyd: Good point, I'll update

thegreyd: Updated

pyartcd/pyartcd/pipelines/build_microshift.py
thegreyd: if diff exists -> pr should always be created , so if pr doesn't exist then something went wrong - i'm happy to let it fail at that point

Ximinhan: then why need `if assembly_basis.brew_event:`

Ximinhan:
And would you mind rebase your commits?

pyartcd/pyartcd/pipelines/build_microshift.py
thegreyd: Good point, I've removed that now - since we are returning early for stream assembly it's not needed. I would like to refactor this in another pr - so we have 2 different funcs for stream and named assemblies so we don't have to do if-else for every step

Ximinhan:
/lgtm
generally lgtm but by the original design this microshift job should not be complicated, if anything wrong we can always start from the beginning with a new build, If we want to post advisory-related steps can be rerun, we can simplify by adding a param to just do that

Ximinhan:
/lgtm
Generally lgtm but by the original design this microshift job should not be complicated, if anything wrong we can always start from the beginning with a new build, If we want to post advisory-related steps can be rerun, we can simplify by adding a param to just do that

PR #32, opened 2023-09-04
gen-payload: respect exempted_packages option

vfreex:
If a package is already listed in exempted_packages, don't make a noise in assembly_issues.


locriandev (commented):
/lgtm

PR #33, opened 2023-09-04
Train Doozer to auto-create missing distgit branches

vfreex:
Updates gen-payload and images:rebase commands to auto-create missing distgit branches.
This is a bit hacky and doesn't work for rpms.


locriandev (commented):
/lgtm

PR #37, opened 2023-09-06
[ocp4_scan] Wait for redis lock manager connection destroy op

thegreyd:
[ocp4_scan](https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Focp4_scan/517575/console) job has a warning `Task was destroyed but it is pending!` after releasing the lock

I believe this should fix it


locriandev (commented):
/lgtm
/approve

PR #38, opened 2023-09-07
ART-7628 Store Jenkins build path as lock identifier

locriandev:
Must go with https://github.com/openshift-eng/aos-cd-jobs/pull/3959

The actual change is represented by setting identifiers for all locks created by `ocp4` equal to the Jenkins build path.
The build path is the build URL minus the host name; so given a build URL as https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Focp4/46870/, all locks created within that run will be identified as `job/aos-cd-builds/job/build%252Focp4/46870`.

This identifier will be then used by a scheduled lock cleanup procedure, which will release all the locks who are identified by a build that's no longer running.

A [test ocp4 build](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/dpaolell/job/ocp4/238/) created `build-lock-4.10` with value `job/hack/job/dpaolell/job/ocp4/238`

The cleanup procedure was then run by a [hack space run](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/dpaolell/job/cleanup-locks/39/console), that correctly deleted only that lock, while keeping everything else intact. More details about testing in https://github.com/openshift-eng/aos-cd-jobs/pull/3959

### DIFF
--- a/pyartcd/pyartcd/jenkins.py
+++ b/pyartcd/pyartcd/jenkins.py
@@ -5,6 +5,7 @@ import time
 from enum import Enum
 from typing import Optional
 
+import requests
 from jenkinsapi.jenkins import Jenkins
 from jenkinsapi.job import Job
 from jenkinsapi.queue import QueueItem
@@ -136,6 +137,21 @@ def set_build_description(build: Build, description: str):
         data="",
         valid=[200]
     )
+
+
+def is_build_running(build_path: str) -> bool:
+    init_jenkins()
+
+    response = requests.get(f'{constants.JENKINS_SERVER_URL}/{build_path}/api/json')
+    if response.status_code != 200:
+        logger.error('Could not fetch data for build %s', build_path)
+        raise ValueError
+
+    build_data = response.json()
+    logger.info('Build %s %s in progress',
+                f'{constants.JENKINS_UI_URL}/{build_path}',
+                'is' if build_data['inProgress'] else 'is not')
+    return build_data['inProgress']
 
 
 @check_env_vars
--- a/pyartcd/pyartcd/jenkins.py
+++ b/pyartcd/pyartcd/jenkins.py
@@ -70,6 +70,12 @@ def get_build_id() -> str:
 
 
 def get_build_id_from_url(build_url: str) -> int:
+    """
+    Examples:
+    - https://buildvm.hosts.prod.psi.bos.redhat.com:8443/job/aos-cd-builds/job/build%252Focp4/46870/ => 46870
+    - https://buildvm.hosts.prod.psi.bos.redhat.com:8443/job/aos-cd-builds/job/build%252Focp4/46870 => 46870
+    """
+
     return int(list(filter(None, build_url.split('/')))[-1])
 
 
@@ -140,6 +146,17 @@ def set_build_description(build: Build, description: str):
 
 
 def is_build_running(build_path: str) -> bool:
+    """
+    Fetches build data using API endpoint {JENKINS_SERVER_URL}/{BUILD_PATH}/api/json
+    E.g. https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Focp4/46902/api/json
+
+    The resulting JSON has a field called "inProgress" that is true if the build is still ongoing
+
+    Build paths examples are:
+    - job/aos-cd-builds/job/build%252Focp4/46902
+    - job/hack/job/dpaolell/job/ocp4/238/
+    """
+
     init_jenkins()
 
     response = requests.get(f'{constants.JENKINS_SERVER_URL}/{build_path}/api/json')
--- a/pyartcd/pyartcd/jenkins.py
+++ b/pyartcd/pyartcd/jenkins.py
@@ -15,8 +15,8 @@ from pyartcd import constants
 
 logger = logging.getLogger(__name__)
 
-current_build_url = None
-current_job_name = None
+current_build_url: Optional[str] = None
+current_job_name: Optional[str] = None
 jenkins_client: Optional[Jenkins] = None
 
 
@@ -52,6 +52,30 @@ def init_jenkins():
     logger.info('Connected to Jenkins %s', jenkins_client.version)
 
 
+def get_build_url():
+    url = os.environ.get("BUILD_URL")
+    if not url:
+        return None
+    return f"{url.rstrip('/')}"
+
+
+def get_build_path():
+    url = get_build_url()
+    return '/'.join(url.split('/')[3:]) if url else None
+
+
+def get_build_id() -> str:
+    return os.environ.get("BUILD_ID")
+
+
+def get_build_id_from_url(build_url: str) -> int:
+    return int(list(filter(None, build_url.split('/')))[-1])
+
+
+def get_job_name():
+    return os.environ.get("JOB_NAME")
+
+
 def check_env_vars(func):
     """
     Enforces that BUILD_URL and JOB_NAME are set
@@ -60,8 +84,8 @@ def check_env_vars(func):
     @functools.wraps(func)
     def wrapped(*args, **kwargs):
         global current_build_url, current_job_name
-        current_build_url = current_build_url or os.environ.get('BUILD_URL', None)
-        current_job_name = current_job_name or os.environ.get('JOB_NAME', None)
+        current_build_url = current_build_url or get_build_url()
+        current_job_name = current_job_name or get_job_name()
 
         if not current_build_url or not current_job_name:
             logger.error('Env vars BUILD_URL and JOB_NAME must be defined!')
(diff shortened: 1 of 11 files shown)


pyartcd/pyartcd/pipelines/ocp4.py
```diff
@@ -928,6 +930,13 @@ async def run(self):
 async def ocp4(runtime: Runtime, version: str, assembly: str, data_path: str, data_gitref: str, pin_builds: bool,
                build_rpms: str, rpm_list: str, build_images: str, image_list: str, skip_plashets: bool,
                mail_list_failure: str, ignore_locks: bool):
+
+    try:
+        job_id = os.environ['JOB_ID']
```
vfreex, line 935: instead of using env var JOB_ID, I think it is better to use something like `Runtime.get_job_run_name` so our code would be more independent from Jenkins environment.

locriandev: As pointed out [here](https://github.com/openshift-eng/aos-cd-jobs/pull/3935#issuecomment-1711429318), the lock identifier must be fed from Jenkins, because that's where the clean up procedure must be initiated. We can use a better env var name if we think `JOB_ID` is not appropriate, but it still needs to be set by Jenkins.

pyartcd/pyartcd/pipelines/ocp4.py
```diff
@@ -928,6 +930,13 @@ async def run(self):
...
+    except KeyError:
+        runtime.logger.error('Env var JOB_ID must be defined')
```
vfreex, line 937: Could you make this env var optional?

pyartcd/pyartcd/pipelines/ocp4.py
thegreyd: resolving this, as this is outdated now

pyartcd/pyartcd/jenkins.py
```diff
@@ -52,6 +53,30 @@ def init_jenkins():
...
+
+def get_build_id_from_url(build_url: str) -> int:
+    return int(list(filter(None, build_url.split('/')))[-1])
+
+
+def get_job_name():
+    return os.environ.get("JOB_NAME")
+
+
 def check_env_vars(func):
```
thegreyd, line 80: Maybe this is better as `set/init_current_job_vars`?

pyartcd/pyartcd/jenkins.py
```diff
@@ -118,6 +139,21 @@ def set_build_description(build: Build, description: str):
...
+    response = requests.get(f'{constants.JENKINS_SERVER_URL}/{build_path}/api/json')
+    if response.status_code != 200:
+        logger.error('Could not fetch data for build %s', build_path)
+        raise ValueError
+
+    build_data = response.json()
+    logger.info('Build %s %s in progress',
+                f'{constants.JENKINS_UI_URL}/{build_path}',
+                'is' if build_data['inProgress'] else 'is not')
+    return build_data['inProgress']
```
thegreyd, line 154: Is InProgress the only "job is still alive/running" state? I'm curious if there can be other weird but valid states that our jobs can fall in

pyartcd/pyartcd/locks.py
```diff
@@ -104,13 +104,33 @@ def from_lock(lock: Lock, use_ssl=True):
...
+        Otherwise, just return all locks currently acquired in Redis
+        """
+
+        if version:
+            pattern = f'*-lock-{version}'
+            self.logger.info('Retrieving active locks for OCP %s', version)
+        else:
+            pattern = '*-lock-*'
+            self.logger.info('Retrieving all active locks')
+
```
thegreyd, line 135: Would prefer logging the search pattern, something like ("Retrieving redis keys by pattern {}") makes it easier to debug & understand what's happening under the hood

pyartcd/pyartcd/pipelines/cleanup_locks.py
```diff
@@ -1,22 +1,38 @@
...
 
-    for lock_name in locks_to_release:
-        runtime.logger.warning('Deleting lock %s', lock_name)
-        res = await redis.delete_key(lock_name)
+    try:
+        for lock_name in active_locks:
+            # Lock ID is a build URL minus Jenkins server base URL
+            build_path = await lock_manager.get_lock_id(lock_name)
+            build_url = f'{constants.JENKINS_UI_URL}/{build_path}'
+            runtime.logger.info('Found build %s associated with lock %s', build_url, lock_name)
```
thegreyd, line 21: I think this can be wrong, when build_path is not a valid build path?

pyartcd/pyartcd/pipelines/cleanup_locks.py
```diff
@@ -1,22 +1,38 @@
...
-        if res:
-            runtime.logger.info('Lock %s deleted', lock_name)
-        else:
-            runtime.logger.warning('Lock %s could not be found', lock_name)
+            try:
+                if not jenkins.is_build_running(build_path):
+                    runtime.logger.warning('Deleting lock %s that was created by %s that\'s not currently running',
+                                           lock_name,
+                                           build_url.replace(constants.JENKINS_SERVER_URL, constants.JENKINS_UI_URL))
+                    lock = Lock(lock_manager=lock_manager, resource=lock_name, id=build_path)
```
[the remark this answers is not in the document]
thegreyd: I can see myself being confused thinking that a lock is being constructed here. I see there is a `get_lock` https://github.com/joanvila/aioredlock/blob/685d1b5b08113738083ac8a449cc56b5ce10cee2/aioredlock/algorithm.py#L260C15-L260C23 that we can use?

thegreyd:
Overall looks good, with some minor suggestions :sparkles:

pyartcd/pyartcd/jenkins.py
thegreyd: Also fine to keep it as is

pyartcd/pyartcd/jenkins.py
thegreyd: Ah missed that it's an attribute. Should be fine then

pyartcd/pyartcd/pipelines/cleanup_locks.py
locriandev: That will cause jenkinsapi to raise an exception, that's caught [here](https://github.com/openshift-eng/art-tools/pull/38/files#diff-30b25a19eea9451c6fffe6c527a6343545fb9e06c5949d1a5c4c8206a5f2bf89R35). The job will log the build that couldn't be fetched, and continue.

This shouldn't happen once old locks are gone, as new ones will always point to a valid Jenkins build

pyartcd/pyartcd/jenkins.py
```diff
@@ -52,6 +53,30 @@ def init_jenkins():
...
+def get_build_path():
+    url = get_build_url()
+    return '/'.join(url.split('/')[3:]) if url else None
+
+
+def get_build_id() -> str:
+    return os.environ.get("BUILD_ID")
+
+
+def get_build_id_from_url(build_url: str) -> int:
```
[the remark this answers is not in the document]
locriandev: You can refer to [unit tests](https://github.com/openshift-eng/art-tools/pull/38/files#diff-b91d4a3b92337394a7080437556248451c7b2f5c6fb0bc7e70f453537ff95f93R85) to see examples. Also pushed a docstring

pyartcd/pyartcd/jenkins.py
```diff
@@ -118,6 +139,21 @@ def set_build_description(build: Build, description: str):
     )
 
 
+def is_build_running(build_path: str) -> bool:
```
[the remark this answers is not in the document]
locriandev: Added a docstring

pyartcd/pyartcd/locks.py
locriandev: Done

pyartcd/pyartcd/pipelines/cleanup_locks.py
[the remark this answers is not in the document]
locriandev: Sounds reasonable, `get_lock()` is just a wrapper around a Lock object construction, but agree that it's going to look clearer. Going to leave this untested and hope for the best :laughing:

pyartcd/pyartcd/pipelines/cleanup_locks.py
thegreyd: Then I would move the log message to after we know it's a valid build
The message, like in hackjob, "Found build https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/7422cbdd-37fc-4d1e-8deb-c701ee6b8dcb associated with lock build-lock-4.14" would be confusing

thegreyd: Hm I think we should do some input validation on the value before constructing and calling it as part of a url, as a good security practice too - I don't know if some url code injection attack exists for jenkins - but yeah just to be safe and follow input validation practices

PR #41, opened 2023-09-07
[promote] Make sure image advisory exists for candidate-type assemblies

thegreyd:
Recently we had a case where rc promotion jumped the gun and went in parallel with prepare-rel, and 
4.14.0-rc.0 didn't get image advisory embedded in it's metadata, and a bunch
of unexpected things happened which confused us ([slack thread](https://redhat-internal.slack.com/archives/C04L1FPFHNH/p1694072265386539))

We have a check in place which complains if this happens, 
make it apply for candidate assemblies as well.


vfreex (commented):
/lgtm
/approval

PR #42, opened 2023-09-08
[ocp4] Handle rpm build failures

thegreyd:
Right now we are not resistant to rpm build failures, as we are to images.
If for some reason an rpm rebase/build fails, we are blocked and fail to build
images & produce nightlies, which can be very bad.

Test run https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Focp4/46517/console


vfreex (commented):
/lgtm
/approve

joepvd:
/hold cancel

Not sure why the hold was set in the first place. Unholding, as I think the (potential) next error will be better than the current :)

PR #44, opened 2023-09-08
distgit: Push to origin

joepvd:
In the current code, we get when building on a new branch:
```
fatal: The current branch rhaos-4.13-rhel-9 has no upstream branch.
To push the current branch and set the remote as upstream, use

    git push --set-upstream origin rhaos-4.13-rhel-9
```

We cannot configure `git` to do `push.autoSetupRemote`, as our git is too old. So move away from rhpkg to push, and use git directly.


locriandev (commented):
/lgtm
/approve

PR #20, opened 2023-08-23
Revert "mass-rebuilds: only for streams.<stream>.image attribute"

joepvd:
Reverts openshift-eng/art-tools#19

PR #48, opened 2023-09-13
mass-rebuilds: only for `streams.<stream>.image` attribute

Ximinhan:
Reverts openshift-eng/art-tools#20


joepvd:
/lgtm
/approve

Yet 
/hold
to wait for the weekend, as this will trigger a mass rebuild

PR #49, opened 2023-09-13
Handle EOL release phase

locriandev:
This PR will account for changes in `group.yaml`, where we can set the release phase for a specific group to be `eol`


Ximinhan:
I guess https://github.com/openshift-eng/art-tools/blob/main/pyartcd/pyartcd/pipelines/check_bugs.py#L22 This part is also affected

locriandev:
> I guess https://github.com/openshift-eng/art-tools/blob/main/pyartcd/pyartcd/pipelines/check_bugs.py#L22 This part is also affected

That can stay as is. `eol` means a release already GA'd in the past, so `pre-release` does not apply.

joepvd (approved):
/lgtm

PR #50, opened 2023-09-13
Skip check-bugs for releases not in 'release' phase

locriandev:


pyartcd/pyartcd/pipelines/check_bugs.py
```diff
@@ -116,16 +111,30 @@ async def _find_blockers(self, version: str):
         self.logger.info('Command returned: %s', out)
         return {version: out}
 
+    async def _is_build_permitted(self, version: str) -> bool:
+        """
+        Only include 'release' state group, exclude 'eol' and 'pre-release'
+        """
+
+        group_config = await util.load_group_config(group=f'openshift-{version}', assembly='stream')
+        phase = group_config['software_lifecycle']['phase']
```
vfreex, line 120: Should this tolerate missing keys?

locriandev: I don't think so. It already didn't, we're just changing the set of possible values. All releases have this key

vfreex (approved):
/lgtm
/approve

PR #51, opened 2023-09-13
[ART-7730] Trigger SAST scans for builds in candidate tags

ashwindasr:
Find all the builds tagged into our candidate tags, get the last brew event ID from the saved file `event.json` in workspace. Run scans for all of the builds with brew event after the last run. If there is no last run, get all the latest builds and run the scans.

Successful test [run](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/ashwin-aos-cd-jobs/job/build%252Fscan-osh-new/19/console).

### DIFF
--- a/doozer/doozerlib/cli/scan_osh.py
+++ b/doozer/doozerlib/cli/scan_osh.py
@@ -1,0 +1,180 @@
+import click
+import koji
+import asyncio
+from doozerlib.cli import cli, click_coroutine, pass_runtime
+from doozerlib.runtime import Runtime
+from doozerlib.exectools import fire_and_forget, cmd_gather_async, limit_concurrency
+from doozerlib.util import cprint
+from typing import Optional
+
+
+class ScanOshCli:
+    def __init__(self, runtime: Runtime, last_brew_event: int, dry_run: bool, nvrs: Optional[list],
+                 check_triggered: Optional[bool], all_builds: Optional[bool]):
+        self.runtime = runtime
+        self.last_brew_event = last_brew_event
+        self.dry_run = dry_run
+        self.brew_tags = []
+        self.specific_nvrs = nvrs
+        self.check_triggered = check_triggered
+        self.all_builds = all_builds
+        self.error_nvrs = []
+
+        # Initialize runtime and brewhub session
+        self.runtime.initialize(clone_distgits=False)
+        self.koji_session = koji.ClientSession(self.runtime.group_config.urls.brewhub)
+
+    @staticmethod
+    async def get_untriggered_nvrs(nvrs):
+        """
+        So we might have already triggered a scan for the same NVR. If this flag is enabled, we want to check
+        if it has and not re-trigger it again.
+        But please note that this will take a lot of time since it has to run for all 200+ packages.
+        """
+
+        @limit_concurrency(16)
+        async def run_get_untriggered_nvrs(nvr):
+            rc, _, _ = await cmd_gather_async(f"osh-cli find-tasks --nvr {nvr}")
+
+            return None if rc == 0 else nvr
+
+        tasks = []
+        for nvr in nvrs:
+            tasks.append(run_get_untriggered_nvrs(nvr))
+
+        nvrs = await asyncio.gather(*tasks)
+
+        untriggered_nvrs = [nvr for nvr in nvrs if nvr]
+
+        return untriggered_nvrs
+
+    def get_tagged_latest(self, tag):
+        """
+        Returns the latest RPMs and builds tagged in to the candidate tag received as input
+        """
+        latest_tagged = self.koji_session.listTagged(tag=tag, latest=True)
+        if latest_tagged:
+            return latest_tagged
+        else:
+            return []
+
+    def get_tagged_all(self, tag):
+        """
+        Returns all the RPMs and builds that are currently in the candidate tag received as input
+        """
+        latest_tagged = self.koji_session.listTagged(tag=tag, latest=False)
+        if latest_tagged:
+            return latest_tagged
+        else:
+            return []
+
+    def trigger_scans(self, nvrs: list):
+        cmd_template = "osh-cli mock-build --config={config} --brew-build {nvr} --nowait"
+        for nvr in nvrs:
+            # Skip rhcos for now
+            if nvr.startswith("rhcos"):
+                self.runtime.logger.warning(f"Skipping RHCOS builds. Scan is not triggered for {nvr}")
+                continue
+
+            if "container" in nvr:
+                cmd = cmd_template.format(config="cspodman", nvr=nvr)
+
+            else:
+                if "el7" in nvr:
+                    rhel_version = 7
+                elif "el8" in nvr:
+                    rhel_version = 8
+                elif "el9" in nvr or nvr.startswith("rhcos"):
+                    rhel_version = 9
+                else:
+                    self.runtime.logger.error("Invalid RHEL version")
+                    return
+
+                cmd = cmd_template.format(config=f"rhel-{rhel_version}-x86_64", nvr=nvr)
+
+            message = f"Ran command: {cmd}"
+
+            if not self.dry_run:
+                fire_and_forget(self.runtime.cwd, cmd)
+            else:
+                message = "[DRY RUN] " + message
+
+            self.runtime.logger.info(message)
+
+        self.runtime.logger.info(f"Total number of build scans kicked off: {len(nvrs)}")
+
+    async def run(self):
+        if not self.specific_nvrs:
+            tags = self.runtime.get_errata_config()["brew_tag_product_version_mapping"].keys()
+            for tag in tags:
+                major, minor = self.runtime.get_major_minor_fields()
+                self.brew_tags.append(tag.format(MAJOR=major, MINOR=minor))
+
+            builds = []
+
+            if self.last_brew_event or self.all_builds:
+                for tag in self.brew_tags:
+                    builds += self.get_tagged_all(tag=tag)
+
+                if self.last_brew_event:
+                    builds = [build for build in builds if build["create_event"] > self.last_brew_event]
+            else:
+                # If no --since field is specified, find all the builds that have been tagged into our candidate tags
+                for tag in self.brew_tags:
+                    builds += self.get_tagged_latest(tag=tag)
+
+            # Sort the builds based on the event ID by descending order so that latest is always on top
+            # Note: create_event is the event on which the build was tagged in the tag and not the build creation time
+            builds = sorted(builds, key=lambda x: x["create_event"], reverse=True)
+
+            nvrs = [build["nvr"] for build in builds]
+            nvr_brew_mapping = [(build["nvr"], build["create_event"]) for build in builds]
+
+            # To store the final list of NVRs that we will kick off scans for
+            nvrs_for_scans = []
+
+            if nvr_brew_mapping:
+                self.runtime.logger.info(f"NVRs to trigger scans for {nvr_brew_mapping}")
+
+            if builds:
+                latest_event_id = nvr_brew_mapping[0][1]
+
+                nvrs_for_scans = nvrs
+
+                # Return back the latest brew event ID
+                cprint(latest_event_id)
+            else:
+                self.runtime.logger.warning(
+                    f"No new NVRs have been found since last brew event: {self.last_brew_event}")
+                return None
+
+        else:
+            self.runtime.logger.info(f"Triggering scans for this particular list of NVRs: {self.specific_nvrs}")
+
+            nvrs_for_scans = self.specific_nvrs
+
+        if self.check_triggered:
+            nvrs_for_scans = await self.get_untriggered_nvrs(nvrs_for_scans)
+
+        self.trigger_scans(nvrs_for_scans)
+
+
+@cli.command("images:scan-osh", help="Trigger scans for builds with brew event IDs greater than the value specified")
+@click.option("--since", required=False, help="Builds after this brew event. If empty, latest builds will retrieved")
+@click.option("--dry-run", default=False, is_flag=True,
+              help="Do not trigger anything, but only print build operations.")
+@click.option("--nvrs", required=False,
+              help="Comma separated list to trigger scans specifically. Will not check candidate tags")
+@click.option("--check-triggered", required=False, is_flag=True, default=False,
+              help="Triggers scans for NVRs only after checking if they haven't already")
+@click.option("--all-builds", required=False, is_flag=True, default=False, help="Check all builds in candidate tags")
+@pass_runtime
+@click_coroutine
+async def scan_osh(runtime: Runtime, since: str, dry_run: bool, nvrs: str, check_triggered: bool, all_builds: bool):
+    cli_pipeline = ScanOshCli(runtime=runtime,
+                              last_brew_event=int(since) if since else None,
+                              dry_run=dry_run,
+                              nvrs=nvrs.split(",") if nvrs else None,
+                              check_triggered=check_triggered,
+                              all_builds=all_builds)
+    await cli_pipeline.run()
--- a/pyartcd/pyartcd/pipelines/brew_scan_osh.py
+++ b/pyartcd/pyartcd/pipelines/brew_scan_osh.py
@@ -1,0 +1,102 @@
+"""
+This job scans the candidate tags for a particular version, and triggers scans for builds that are tagged into it.
+"""
+import click
+from typing import Optional
+from pyartcd import exectools, redis
+from pyartcd.runtime import Runtime
+from pyartcd.cli import cli, pass_runtime, click_coroutine
+
+
+class OshScan:
+    def __init__(self, runtime: Runtime, version: str, check_triggered: Optional[bool], email: Optional[str],
+                 nvrs: Optional[list], all_builds: Optional[bool]):
+        self.runtime = runtime
+        self.email = email
+        self.version = version
+        self.major, self.minor = self.version.split(".")
+        self.check_triggered = check_triggered
+        self.specific_nvrs = nvrs
+        self.all_builds = all_builds
+
+        self.redis_key = f"appdata:sast-scan:{self.version}"
+
+    async def store_brew_event(self, brew_event_id: str):
+        await redis.set_value(self.redis_key, brew_event_id)
+
+    async def retrieve_brew_event(self):
+        last_brew_event = await redis.get_value(self.redis_key)
+
+        # Remove trailing \n
+        if last_brew_event:
+            last_brew_event = last_brew_event.strip()
+
+        return last_brew_event
+
+    async def run(self):
+        cmd = [
+            "doozer",
+            "--group",
+            f"openshift-{self.version}",
+            "images:scan-osh",
+        ]
+
+        last_brew_event = await self.retrieve_brew_event()
+
+        self.runtime.logger.info(f"Found last brew event from REDIS: {last_brew_event}")
+
+        if self.runtime.dry_run:
+            cmd.append("--dry-run")
+
+        if self.check_triggered:
+            cmd.append("--check-triggered")
+
+        if self.specific_nvrs or self.all_builds:
+            if self.specific_nvrs:
+                cmd += [
+                    "--nvrs",
+                    f"{','.join(self.specific_nvrs)}"
+                ]
+
+            if self.all_builds:
+                cmd.append("--all-builds")
+        else:
+            # So if we give the specific list of nvrs, we don't need to pass the last brew event
+            if last_brew_event:
+                cmd += [
+                    "--since",
+                    f"{last_brew_event}"
+                ]
+
+        _, brew_event, _ = await exectools.cmd_gather_async(cmd, stderr=True)
+        if not brew_event and not self.specific_nvrs:
+            self.runtime.logger.warning(f"No new builds found for candidate tags in {self.version}")
+            return
+
+        if not self.specific_nvrs and not self.all_builds:
+            # We do not need to write to Redis if we're manually triggering scans for specific builds
+            if not self.runtime.dry_run:
+                # Write the latest brew event back to the file
+                if last_brew_event != brew_event:
+                    # No need to set if it's the same value
+                    await self.store_brew_event(brew_event)
+            else:
+                self.runtime.logger.info("[DRY RUN] Would have written to Redis")
+
+
+@cli.command("scan-osh")
+@click.option("--version", required=True, help="openshift version eg: 4.15")
+@click.option("--email", required=False, help="Additional email to which the results of the scan should be sent out to")
+@click.option("--nvrs", required=False, help="Comma separated list to trigger scans specifically. Will not check candidate tags")
+@click.option("--check-triggered", required=False, is_flag=True, default=False, help="Triggers scans for NVRs only after checking if they haven't already")
+@click.option("--all-builds", required=False, is_flag=True, default=False, help="Check all builds in candidate tags")
+@pass_runtime
+@click_coroutine
+async def scan_osh(runtime: Runtime, version: str, email: str, nvrs: str, check_triggered: bool, all_builds: bool):
+    pipeline = OshScan(runtime=runtime,
+                       email=email,
+                       version=version,
+                       nvrs=nvrs.split(",") if nvrs else None,
+                       check_triggered=check_triggered,
+                       all_builds=all_builds)
+    await pipeline.run()
--- a/doozer/doozerlib/cli/__main__.py
+++ b/doozer/doozerlib/cli/__main__.py
@@ -42,6 +42,7 @@ from doozerlib.cli.config_plashet import config_plashet
 from doozerlib.cli.release_calc_upgrade_tests import release_calc_upgrade_tests
 from doozerlib.cli.inspect_stream import inspect_stream
 from doozerlib.cli.config_tag_rpms import config_tag_rpms
+from doozerlib.cli.scan_osh import scan_osh
 
 from doozerlib import coverity
 from doozerlib.exceptions import DoozerFatalError
--- a/doozer/doozerlib/runtime.py
+++ b/doozer/doozerlib/runtime.py
@@ -263,6 +263,9 @@ class Runtime(object):
 
         return assembly_group_config(self.get_releases_config(), self.assembly, tmp_config)
 
+    def get_errata_config(self, **kwargs):
+        return self.gitdata.load_data(key='erratatool', **kwargs).data
+
     def _get_replace_vars(self, group_config: Model):
         replace_vars = group_config.vars or Model()
         # If assembly mode is enabled, `runtime_assembly` will become the assembly name.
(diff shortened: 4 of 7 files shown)


ashwindasr:
Looks like `cmd_gather_async` is waiting for tasks to finish, will change to fire and forget. Working on fix.

thegreyd:
I see rhcos builds included as well `osh-cli mock-build --config=rhel-9-x86_64 --brew-build rhcos-s390x-415.92.202309221537-0 --nowait` is rhcos supported? I'm assuming these fail since rhcos isn't an rpm, maybe exclude these?

pyartcd/pyartcd/pipelines/brew_scan_osh.py
```diff
@@ -0,0 +1,111 @@
...
+from pyartcd import exectools
+from pyartcd.runtime import Runtime
+from pyartcd.cli import cli, pass_runtime, click_coroutine
+
+FILE_PATH = f"{os.getenv('WORKSPACE')}/event.json"
+
+
+class OshScan:
+    def __init__(self, runtime: Runtime, email: str, version: str):
+        self.runtime = runtime
```
thegreyd, line 20: we can add support from dry_run, using `self.runtime.dry_run`

ashwindasr:
@thegreyd rhcos runs on rhel9 right, so I added [this](https://github.com/openshift-eng/art-tools/blob/43af36a34e1cf6efc67602c1d7758948891a2756/doozer/doozerlib/cli/scan_osh.py#L56). Should I exclude them instead?

pyartcd/pyartcd/pipelines/brew_scan_osh.py
```diff
@@ -0,0 +1,111 @@
...
+
+        return yaml_data["brew_tag_product_version_mapping"]
+
+    async def _set_candidate_tags(self):
+        data = await self._get_ocp_candidate_tags()
+
+        self.tags = [tag.replace("{MAJOR}", self.major).replace("{MINOR}", self.minor) for tag in
+                     data]
+
+    def _get_last_event(self):
```
thegreyd, line 53: How about using redis instead to store and retrieve?

doozer/doozerlib/constants.py
```diff
@@ -8,6 +8,7 @@
 
 GITHUB_TOKEN = "GITHUB_TOKEN"
 BREWWEB_URL = "https://brewweb.engineering.redhat.com/brew"
+BREWHUB_URL = "https://brewhub.engineering.redhat.com/brewhub"
```
thegreyd, line 11: I don't think we want to hardcode this, instead use https://github.com/openshift-eng/ocp-build-data/blob/e7585b181712a17f548e6601ce434f1e0d08299e/group.yml#L90

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -0,0 +1,114 @@
...
+        """
+        Returns all the RPMs and builds that are currently in the candidate tag received as input
+        """
+        session = koji.ClientSession(BREWHUB_URL)
+        try:
+            latest_tagged = session.listTagged(tag=tag, latest=False)
+            if latest_tagged:
+                return latest_tagged
+        except Exception:
+            raise
```
thegreyd, line 43: Why catch the exception here?

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -0,0 +1,114 @@
...
+
+    async def trigger_scan(self, nvrs: list):
+        cmd_template = "osh-cli mock-build --config={} --brew-build {} --nowait"
```
thegreyd, line 46: we can name our template vars `--config={config} --brew-build {nvr}` and then later do `.format(config="cspodman", nvr=nvr)`

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -0,0 +1,114 @@
...
+from doozerlib.constants import BREWHUB_URL
+from doozerlib.util import cprint
+
+
+class ScanOshCli:
+    def __init__(self, runtime: Runtime, brew_tags: list, last_brew_event):
+        self.runtime = runtime
+        self.brew_tags = brew_tags
+        self.last_brew_event = last_brew_event
+
```
thegreyd, line 15: dry-run support would be good here

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -0,0 +1,114 @@
...
+        session = koji.ClientSession(BREWHUB_URL)
+        try:
+            latest_tagged = session.listTagged(tag=tag, latest=True)
+            if latest_tagged:
+                return latest_tagged
+        except Exception:
+            raise
+
+    @staticmethod
+    async def get_tagged_all(tag):
```
thegreyd, line 33: Why is this async?

thegreyd:
Can we trigger a scan for rhcos and see if it works / passes? If it doesn't don't see a value to keep triggering scan for it

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -0,0 +1,114 @@
...
+    async def get_tagged_latest(tag):
+        """
+        Returns the latest RPMs and builds tagged in to the candidate tag received as input
+        """
+        session = koji.ClientSession(BREWHUB_URL)
+        try:
+            latest_tagged = session.listTagged(tag=tag, latest=True)
+            if latest_tagged:
+                return latest_tagged
+        except Exception:
```
locriandev, line 29: This is a no-op, unless you do something in the exception handler. Otherwise, you might just as well remove the try-except

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -0,0 +1,114 @@
...
+    async def get_tagged_all(tag):
+        """
+        Returns all the RPMs and builds that are currently in the candidate tag received as input
+        """
+        session = koji.ClientSession(BREWHUB_URL)
+        try:
+            latest_tagged = session.listTagged(tag=tag, latest=False)
+            if latest_tagged:
+                return latest_tagged
+        except Exception:
```
locriandev, line 42: This is a no-op, unless you do something in the exception handler. Otherwise, you might just as well remove the try-except

pyartcd/pyartcd/pipelines/brew_scan_osh.py
```diff
@@ -0,0 +1,111 @@
...
+        self.version = version
+        self.major, self.minor = self.version.split(".")
+        self.last_event = {}
+        self.tags = None
+
+    async def _get_ocp_candidate_tags(self):
+        """
+        Get the candidate tags activated for a specific OCP version
+        """
+        # GITHUB_TOKEN env variable needs to be set in the Jenkinsfile to avoid rate limiting
```
locriandev, line 31: We could log a warning if GITHUB_TOKEN is not set

pyartcd/pyartcd/pipelines/brew_scan_osh.py
locriandev: +1. Let's think about a pattern to keep data organized. We currently have `count` and `failures` as main keys. We could add something like `appdata:{app_name}:{key}` where appname would be `sast_scan` in this case.

doozer/doozerlib/cli/scan_osh.py
ashwindasr: ahh since its an API call I thought async will be faster? unless I'm mistaken

pyartcd/pyartcd/pipelines/brew_scan_osh.py
ashwindasr: fixed in https://github.com/openshift-eng/art-tools/pull/51/commits/906d84751f15b00b02083b9cfe428b05c0888896

pyartcd/pyartcd/pipelines/brew_scan_osh.py
```diff
@@ -0,0 +1,111 @@
...
+class OshScan:
+    def __init__(self, runtime: Runtime, email: str, version: str):
+        self.runtime = runtime
+        self.email = email
+        self.version = version
+        self.major, self.minor = self.version.split(".")
+        self.last_event = {}
+        self.tags = None
+
+    async def _get_ocp_candidate_tags(self):
```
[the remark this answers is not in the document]
ashwindasr: fixed in https://github.com/openshift-eng/art-tools/pull/51/commits/906d84751f15b00b02083b9cfe428b05c0888896

doozer/doozerlib/constants.py
ashwindasr: fixed in https://github.com/openshift-eng/art-tools/pull/51/commits/906d84751f15b00b02083b9cfe428b05c0888896

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -0,0 +1,114 @@
...
+            raise
+
+    @staticmethod
+    async def get_tagged_all(tag):
+        """
+        Returns all the RPMs and builds that are currently in the candidate tag received as input
+        """
+        session = koji.ClientSession(BREWHUB_URL)
```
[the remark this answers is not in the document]
ashwindasr: fixed in https://github.com/openshift-eng/art-tools/pull/51/commits/906d84751f15b00b02083b9cfe428b05c0888896

doozer/doozerlib/cli/scan_osh.py
ashwindasr: I was testing before, so reused the code. Forgot to remove it apologies. Fixed in https://github.com/openshift-eng/art-tools/pull/51/commits/906d84751f15b00b02083b9cfe428b05c0888896

doozer/doozerlib/cli/scan_osh.py
ashwindasr: Ahh yes! I was looking for this but couldn't find it, thank you. Fixed in https://github.com/openshift-eng/art-tools/pull/51/commits/906d84751f15b00b02083b9cfe428b05c0888896

doozer/doozerlib/cli/scan_osh.py
ashwindasr: Done https://github.com/openshift-eng/art-tools/pull/51/commits/906d84751f15b00b02083b9cfe428b05c0888896

doozer/doozerlib/cli/scan_osh.py
ashwindasr: fixed in https://github.com/openshift-eng/art-tools/pull/51/commits/906d84751f15b00b02083b9cfe428b05c0888896

doozer/doozerlib/cli/scan_osh.py
ashwindasr: fixed in https://github.com/openshift-eng/art-tools/pull/51/commits/906d84751f15b00b02083b9cfe428b05c0888896

pyartcd/pyartcd/pipelines/brew_scan_osh.py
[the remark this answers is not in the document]
ashwindasr: looks like doozer didn't have it, added the function [from elliott](https://github.com/openshift-eng/art-tools/blob/906d84751f15b00b02083b9cfe428b05c0888896/doozer/doozerlib/runtime.py#L266) Runtime

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -3,50 +3,41 @@
...
-                return latest_tagged
-        except Exception:
-            raise
-
-    @staticmethod
-    async def get_tagged_all(tag):
+        latest_tagged = self.koji_session.listTagged(tag=tag, latest=True)
+        if latest_tagged:
+            return latest_tagged
+
```
thegreyd, line 27: We could return [] instead of None when there's no latest_tagged?

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -59,14 +50,23 @@ async def trigger_scan(self, nvrs: list):
...
 
-            # Comment out the below line for testing.
-            fire_and_forget(self.runtime.cwd, cmd)
+            self.runtime.logger.info(message)
 
     async def run(self):
+        tags = self.runtime.get_errata_config()["brew_tag_product_version_mapping"].keys()
+        for tag in tags:
+            major, minor = self.runtime.group.replace("openshift-", "").split(".")
+            self.brew_tags.append(tag.replace("{MAJOR}", major).replace("{MINOR}", minor))
```
thegreyd, line 68: `tag.format(MAJOR=major, MINOR=minor)` instead of .replace?

doozer/doozerlib/cli/scan_osh.py
thegreyd: Unless we're creating multiple coroutines then this doesn't make a difference

thegreyd: Also take a look at different koji client implementations, there's "build_retrying_koji_client" that we can use for auto retry. If you want to create multiple threads, then "shared_koji_client_session" is what we should use

doozer/doozerlib/cli/scan_osh.py
ashwindasr: done, fixed in https://github.com/openshift-eng/art-tools/pull/51/commits/c4c66665b3c5b9cc9d9b473fb0af3c5dbdf1253c

doozer/doozerlib/cli/scan_osh.py
ashwindasr: ahh yes, apologies. fixed in https://github.com/openshift-eng/art-tools/pull/51/commits/c4c66665b3c5b9cc9d9b473fb0af3c5dbdf1253c

doozer/doozerlib/cli/scan_osh.py
ashwindasr: Understood. Updated in https://github.com/openshift-eng/art-tools/pull/51/commits/35128eb17ea67bb154c2c093d91a370131e52f55

pyartcd/pyartcd/pipelines/brew_scan_osh.py
```diff
@@ -0,0 +1,75 @@
...
+        if self.runtime.dry_run:
+            cmd.append("--dry-run")
+
+        if self.check_triggered:
+            cmd.append("--check-triggered")
+
+        _, brew_event, _ = await exectools.cmd_gather_async(cmd, stderr=True)
+        if not brew_event:
+            self.runtime.logger.error(f"No new builds found for candidate tags in {self.version}")
+            return
```
thegreyd, line 56: I don't think this exits with an error code? sys.exit(1)?

pyartcd/pyartcd/pipelines/brew_scan_osh.py
```diff
@@ -0,0 +1,75 @@
...
+            "doozer",
+            "--group",
+            f"openshift-{self.version}",
+            "images:scan-osh",
+        ]
+
+        last_brew_event = await self.get_last_event_redis()
+
+        # Remove trailing /n
+        last_brew_event = last_brew_event.strip()
```
thegreyd, line 39: nit: this would be better in the get method

pyartcd/pyartcd/pipelines/brew_scan_osh.py
```diff
@@ -0,0 +1,75 @@
+"""
+[STORES STATE IN REDIS]
```
thegreyd, line 2: I would remove this, since it can get stale in the future - we are better off reading code to see what happens where

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -0,0 +1,157 @@
...
+                return None
+        else:
+            self.runtime.logger.info(f"Triggering scans for this particular list of NVRs: {self.specific_nvrs}")
+
+            self.nvrs = self.specific_nvrs
+
+        if self.check_triggered:
+            await self.untriggered_nvrs()
+
+        await self.trigger_scan()
```
thegreyd, line 138: Would recommend passing nvrs to the method and not storing as a class var, that way it's easy to see what data the method needs and operates on, and easier to test

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -0,0 +1,157 @@
...
+                self.runtime.logger.warning(
+                    f"No new NVRs have been found since last brew event: {self.last_brew_event}")
+                return None
+        else:
+            self.runtime.logger.info(f"Triggering scans for this particular list of NVRs: {self.specific_nvrs}")
+
+            self.nvrs = self.specific_nvrs
+
+        if self.check_triggered:
+            await self.untriggered_nvrs()
```
thegreyd, line 136: It's not apparent what is happening here. Something like 
`nvrs = self.get_untriggered_nvrs(nvrs)` would make it clear

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -0,0 +1,157 @@
...
+                for tag in self.brew_tags:
+                    builds += self.get_tagged_latest(tag=tag)
+                builds = sorted(builds, key=lambda x: x["create_event"], reverse=True)
+
+            nvrs = [build["nvr"] for build in builds]
+            nvr_brew_mapping = [(build["nvr"], build["create_event"]) for build in builds]
+
+            if nvr_brew_mapping:
+                self.runtime.logger.info(f"NVRs to trigger scans for {nvr_brew_mapping}")
+                pass
```
thegreyd, line 117: leftover

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -0,0 +1,157 @@
...
+
+            if builds:
+                latest_event_id = nvr_brew_mapping[0][1]
```
thegreyd, line 120: I would bring the sorting that's happening at line 104 closer to this, or in other words, sort just before you extract the latest_event_id

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -0,0 +1,157 @@
...
+            else:
+                message = "[DRY RUN] " + message
+
+            self.runtime.logger.info(message)
+
+    async def run(self):
+        if not self.specific_nvrs:
+            tags = self.runtime.get_errata_config()["brew_tag_product_version_mapping"].keys()
+            for tag in tags:
+                major, minor = self.runtime.group.replace("openshift-", "").split(".")
```
thegreyd, line 93: there's a method called `get_major_minor` that we can use here

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -0,0 +1,157 @@
...
+            else:
+                if "el7" in nvr:
+                    rhel_version = 7
+                elif "el8" in nvr:
+                    rhel_version = 8
+                elif "el9" in nvr or nvr.startswith("rhcos"):
+                    rhel_version = 9
+                else:
+                    self.runtime.logger.error("Invalid RHEL version")
+                    return
```
thegreyd, line 76: in a case like this, we might want to trigger all remaining scans for valid nvrs, and then exit with an error

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -0,0 +1,157 @@
...
+
+    async def untriggered_nvrs(self):
+        """
+        So we might have already triggered a scan for the same NVR. If this flag is enabled, we want to check
+        if it has and not re-trigger it again.
+        But please note that this will take a lot of time since it has to run for all 200+ packages.
+        """
+        untriggered_nvrs = []
+        for nvr in self.nvrs:
+            rc, _, _ = cmd_gather(f"osh-cli find-tasks --nvr {nvr}")
```
thegreyd, line 33: We can create couroutines here to make this faster. If not in this PR, let's remove the async from method def

locriandev: Also consider using `exectools.limit_concurrency` decorator if doing so, we might not spawn all tasks at the same time.

pyartcd/pyartcd/pipelines/brew_scan_osh.py
```diff
@@ -0,0 +1,75 @@
...
+    def __init__(self, runtime: Runtime, version: str, check_triggered: Optional[bool], email: Optional[str]):
+        self.runtime = runtime
+        self.email = email
+        self.version = version
+        self.major, self.minor = self.version.split(".")
+        self.check_triggered = check_triggered
+
+        self.redis_key = f"appdata:sast-scan:{self.version}"
+
+    async def set_last_event_redis(self, brew_event_id: str):
```
locriandev, line 22: I would rather name these `store_brew_event()` and `retrieve_brew_event()`. Redis is an implementation detail and might change over time

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -0,0 +1,157 @@
...
+        """
+        Returns all the RPMs and builds that are currently in the candidate tag received as input
+        """
+        latest_tagged = self.koji_session.listTagged(tag=tag, latest=False)
+        if latest_tagged:
+            return latest_tagged
+        else:
+            return []
+
+    async def trigger_scan(self):
```
locriandev, line 61: Does not look like a coroutine, it does not `await` anything. Can be a regular sync function

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -0,0 +1,157 @@
...
+        cmd_template = "osh-cli mock-build --config={config} --brew-build {nvr} --nowait"
+        for nvr in self.nvrs:
+            if "container" in nvr:
+                cmd = cmd_template.format(config="cspodman", nvr=nvr)
+
+            else:
+                if "el7" in nvr:
```
locriandev, line 68: Maybe we can add a function similar to `util.split_el_suffix_in_release()` to extract the rhel version from an NVR and return it as an int

pyartcd/pyartcd/pipelines/brew_scan_osh.py
ashwindasr: Well we don't want to error out since there might not be builds after an event ID right. Changed to logger.warning if that is better

pyartcd/pyartcd/pipelines/brew_scan_osh.py
ashwindasr: https://github.com/openshift-eng/art-tools/pull/51/commits/7c715b307aa02efa5d521ca0e931c9a2ae18766e

pyartcd/pyartcd/pipelines/brew_scan_osh.py
ashwindasr: https://github.com/openshift-eng/art-tools/pull/51/commits/7c715b307aa02efa5d521ca0e931c9a2ae18766e

doozer/doozerlib/cli/scan_osh.py
ashwindasr: https://github.com/openshift-eng/art-tools/pull/51/commits/7c715b307aa02efa5d521ca0e931c9a2ae18766e

doozer/doozerlib/cli/scan_osh.py
ashwindasr: https://github.com/openshift-eng/art-tools/pull/51/commits/7c715b307aa02efa5d521ca0e931c9a2ae18766e

doozer/doozerlib/cli/scan_osh.py
ashwindasr: https://github.com/openshift-eng/art-tools/pull/51/commits/7c715b307aa02efa5d521ca0e931c9a2ae18766e

doozer/doozerlib/cli/scan_osh.py
ashwindasr: ahh! Thank you! https://github.com/openshift-eng/art-tools/pull/51/commits/7c715b307aa02efa5d521ca0e931c9a2ae18766e

doozer/doozerlib/cli/scan_osh.py
ashwindasr: removing from PR for now https://github.com/openshift-eng/art-tools/pull/51/commits/7c715b307aa02efa5d521ca0e931c9a2ae18766e

pyartcd/pyartcd/pipelines/brew_scan_osh.py
ashwindasr: https://github.com/openshift-eng/art-tools/pull/51/commits/7c715b307aa02efa5d521ca0e931c9a2ae18766e

doozer/doozerlib/cli/scan_osh.py
ashwindasr: https://github.com/openshift-eng/art-tools/pull/51/commits/7c715b307aa02efa5d521ca0e931c9a2ae18766e

ashwindasr:
Hmm GitHub wiped out my replies to the comments. Retyped again for most. Sorry if you guys got notified multiple times.

ashwindasr:
@thegreyd regarding [this](https://github.com/openshift-eng/art-tools/pull/51#discussion_r1337795196) I feel like it will pollute the brew event in Redis. Since all the NVRs are sorted while we receive it, if we skip an NVR if there is an error, the next time it automatically runs, the brew event will be ahead and it will never be scanned

doozer/doozerlib/cli/scan_osh.py
ashwindasr: https://github.com/openshift-eng/art-tools/pull/51#issuecomment-1738059261

doozer/doozerlib/cli/scan_osh.py
ashwindasr: is it okay if we keep this as a refinement?

ashwindasr:
Will squash commits after review

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -0,0 +1,179 @@
...
+
+    @staticmethod
+    async def get_untriggered_nvrs(nvrs):
+        """
+        So we might have already triggered a scan for the same NVR. If this flag is enabled, we want to check
+        if it has and not re-trigger it again.
+        But please note that this will take a lot of time since it has to run for all 200+ packages.
+        """
+
+        @limit_concurrency(50)
```
thegreyd, line 35: Let's change this to 16, I see in other places 16/32 were the numbers chosen for multiple http calls, so that should be a safe number

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -0,0 +1,179 @@
...
+                    builds += self.get_tagged_all(tag=tag)
+
+                if self.last_brew_event:
+                    builds = [build for build in builds if build["create_event"] > self.last_brew_event]
+            else:
+                # If no --since field is specified, find all the builds that have been tagged into our candidate tags
+                for tag in self.brew_tags:
+                    builds += self.get_tagged_latest(tag=tag)
+
+            # Sort the builds based on the event ID by descending order so that latest is always on top
```
thegreyd, line 126: also mention that create_Event is the event on which the build was tagged in the tag and not the build creation time

doozer/doozerlib/cli/scan_osh.py
ashwindasr: Buildsync is using 500. So this should be safe to keep right

thegreyd:
1 comment to change the concurrency limit & 1 more minor comment 
otherwise lgtm, 
feel free apply lgtm lable/merge it when those 2 comments are addressed

ashwindasr:
Tests passed locally.

PR #53, opened 2023-09-14
Don't skip advisory prep if builds are pinned

thegreyd:
Follow up to https://github.com/openshift-eng/art-tools/pull/31
This was somehow missed in #31 - fixing that
Test run: https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fbuild-microshift/592/console

### DIFF
--- a/pyartcd/pyartcd/pipelines/build_microshift.py
+++ b/pyartcd/pyartcd/pipelines/build_microshift.py
@@ -116,7 +116,8 @@ class BuildMicroShiftPipeline:
                 await self.slack_say(f":construction: Microshift prep for assembly {self.assembly} :construction:")
 
             # For named assemblies, check if builds are pinned or already exist
-            nvrs, pinned_nvrs = [], []
+            nvrs = []
+            pinned_nvrs = dict()
             pr = None
 
             if assembly_type is not AssemblyTypes.STREAM and not self.force:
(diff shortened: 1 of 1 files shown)


pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -126,8 +127,9 @@ async def run(self):
                                "rebuild.")
                     self._logger.info(message)
                     await self.slack_say(message)
-                    return
```
thegreyd, line 129: We don't want to return here

pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -126,8 +127,9 @@ async def run(self):
...
-                nvrs = await self._find_builds()
+                    nvrs = pinned_nvrs.values()
+                else:
+                    nvrs = await self._find_builds()
```
ashwindasr, line 132: A small nit, but nvrs will have two types depending on the if-else, it can be `dict_values` or `list`. Ofc it won't make a difference since I don't see anything that will break, but if in future we try to access it as a list, it will fail since it could also be a `dict_values`.

thegreyd: That's a good point! I'll change it

PR #55, opened 2023-09-15
[promote] Move/verify advisories only for standard assemblies

thegreyd:
## Summary
When promoting candidate assemblies:
- Don't move advisories to QE
- Don't error out if advisories are not on QE

In case of candidate assemblies advisories are being prepared every week and are usually missing cdn repos or have other things going on with them which can prevent them from being on QE.

These operations would still be performed for these set of advisories, when they are 
promoted as part of the GA standard assembly.

## Test run
https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fpromote-assembly/843/parameters/

### DIFF
--- a/pyartcd/pyartcd/pipelines/promote.py
+++ b/pyartcd/pyartcd/pipelines/promote.py
@@ -179,20 +179,21 @@ class PromotePipeline:
                     justifications.append(justification)
                 logger.info("No blocker bugs found.")
 
-            # Attempt to move all advisories to QE
-            tasks = []
-            for impetus, advisory in impetus_advisories.items():
-                if not advisory:
-                    continue
-                logger.info("Moving advisory %s to QE...", advisory)
-                if not self.runtime.dry_run:
-                    tasks.append(self.change_advisory_state(advisory, "QE"))
-                else:
-                    logger.warning("[DRY RUN] Would have moved advisory %s to QE", advisory)
-            try:
-                await asyncio.gather(*tasks)
-            except ChildProcessError as err:
-                logger.warn("Error moving advisory %s to QE: %s", advisory, err)
+            if assembly_type == assembly.AssemblyTypes.STANDARD:
+                # Attempt to move all advisories to QE
+                tasks = []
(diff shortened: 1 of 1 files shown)


pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -207,9 +208,10 @@ async def run(self):
                     logger.info("Verifying associated image advisory %s...", image_advisory)
                     image_advisory_info = await self.get_advisory_info(image_advisory)
                     try:
-                        self.verify_image_advisory(image_advisory_info)
-                        live_id = self.get_live_id(image_advisory_info)
-                        assert live_id
+                        self.verify_advisory_live_id(image_advisory_info)
+                        if assembly_type != assembly.AssemblyTypes.CANDIDATE:
+                            self.verify_advisory_status(image_advisory_info)
+                        live_id = self.get_live_id(image_advisory_info) or 'live-id'
```
locriandev, line 214: IIUC, when `live_id` is None `verify_advisory_live_id()` will throw an exception. So what situation do you expect to cover with this check?

Also, `verify_advisory_live_id()` is just running `get_live_id()` which you are doing again at line 214, then possibly throw an exception. Wouldn't it make sense to re-write this as something like:

```
live_id = self.get_live_id(image_advisory_info)
if not live_id:
    raise VerificationError(f"Advisory {advisory_info['id']} doesn't have a live ID.")
if assembly_type != assembly.AssemblyTypes.CANDIDATE:
    self.verify_advisory_status(image_advisory_info)
```

and get rid of the redundant `verify_advisory_live_id()` function?

thegreyd: Right, I was thinking about `errata_url` still having something other than an empty string when live_id was None, but that doesn't happen since the exception gets raised. I think it's fine for errata_url to stay an empty string instead of a dummy value "access.redhat.com/errata/live-id" - cincinnat/ota team might be relying on it to be that way. Updated with your suggestion

PR #57, opened 2023-09-18
[ART-7781] --gather-dependencies for elliott verify-attached-operators

sosiouxme:
Automate pre-release gathering of bundle dependencies to a single advisory.

https://issues.redhat.com/browse/ART-7781

### DIFF
--- a/elliott/elliottlib/cli/common.py
+++ b/elliott/elliottlib/cli/common.py
@@ -4,7 +4,7 @@ from functools import update_wrapper
 
 import click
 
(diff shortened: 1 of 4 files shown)


sosiouxme:
/hold
i found an edge case that needs to be handled better

sosiouxme:
/hold cancel
I needed a small refactor to enable re-verifying after moving things around and potentially adding builds that weren't attached and couldn't be verified at the start (or anything else that might have gone wrong).

PR #60, opened 2023-09-20
ART-7628 Store Jenkins build path as lock value

locriandev:
Provide all lock-using jobs with the capability of tracking Jenkins builds:
- ocp4_scan
- rebuild
- olm_bundle
- promote

### DIFF
--- a/pyartcd/pyartcd/jenkins.py
+++ b/pyartcd/pyartcd/jenkins.py
@@ -61,6 +61,14 @@ def get_build_url():
 
 
 def get_build_path():
+    """
+    Examples:
+    - https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Focp4/46870/ =>
+        job/aos-cd-builds/job/build%252Focp4/46870
+    - https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Focp4/46870 =>
+        job/aos-cd-builds/job/build%252Focp4/46870
+    """
+
     url = get_build_url()
     return '/'.join(url.split('/')[3:]) if url else None
 
--- a/pyartcd/pyartcd/pipelines/ocp4_scan.py
+++ b/pyartcd/pyartcd/pipelines/ocp4_scan.py
@@ -27,7 +27,7 @@ class Ocp4ScanPipeline:
 
     async def run(self):
         # Check if automation is frozen for current group
-        if not await util.is_build_permitted(self.version, doozer_working=self._doozer_working):
+        if not await util.is_build_permitted(self.version, doozer_working=str(self._doozer_working)):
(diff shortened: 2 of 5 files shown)


pyartcd/pyartcd/jenkins.py
```diff
@@ -61,6 +61,14 @@ def get_build_url():
 
 
 def get_build_path():
+    """
+    Examples:
+    - https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Focp4/46870/ =>
+        job/aos-cd-builds/job/build%252Focp4/46870
+    - https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Focp4/46870 =>
```
thegreyd, line 68: repeat?

pyartcd/pyartcd/pipelines/ocp4_scan.py
```diff
@@ -136,14 +138,19 @@ async def ocp4_scan(runtime: Runtime, version: str):
...
+    if not lock_identifier:
+        runtime.logger.warning('Env var BUILD_URL has not been defined: a random identifier will be used for the locks')
 
     try:
         # Skip the build if already locked
         if await lock_manager.is_locked(lock_name):
             runtime.logger.info('Looks like there is another build ongoing for %s -- skipping for this run', version)
         else:
-            async with await lock_manager.lock(lock_name):
+            async with await lock_manager.lock(resource=lock_name, lock_identifier=lock_identifier):
```
thegreyd, line 153: I'm thinking about a pattern where to request a lock, we enforce providing a job path identifier. And maybe we can roll the lock manager creation in there as well

So something like

```
async def request_lock(lock_type, lock_name, lock_identifier, **kwargs):
    lock_manager = locks.LockManager.from_lock(lock_type)
    if not lock_identifier:
        raise ValueError('...')
    return lock_manager.lock(resource=lock_name, lock_identifier=lock_identifier)
```

pyartcd/pyartcd/jenkins.py
locriandev: It shows that the trailing '/' is removed

pyartcd/pyartcd/pipelines/ocp4_scan.py
locriandev: Sounds good to me, but I'd rather do that in a separate PR

locriandev: Looking at it more closely: we need a `LockManager` instance available outside this function to clear redis connection (which happens by calling `lock_manager.destroy()`. What we could do is enforce a `lock_identifier` to be specified as keyword arg [here](https://github.com/openshift-eng/art-tools/blob/main/pyartcd/pyartcd/locks.py#L107). Not sure though if we want to give up the chance of getting an auto-generated random identifier, in cases where we might not be interested in the job that created it

PR #61, opened 2023-09-20
Use : separator for lock names

locriandev:
Using `:` as a separator in Redis keys allows for a better visualization in the UI.

image

Tested with:

```
import asyncio

from pyartcd import redis
from pyartcd.locks import LockManager


async def main():
    lock_manager = LockManager([redis.redis_url()])

    locks = await lock_manager.get_locks()
    for lock in locks:
        print(f'{lock} - {await lock_manager.get_lock_id(lock)}')

    print('-----------------------------------------')

    locks = await lock_manager.get_locks('4.5')
    for lock in locks:
        print(f'{lock} - {await lock_manager.get_lock_id(lock)}')


if __name__ == '__main__':
    asyncio.new_event_loop().run_until_complete(main())
```

### DIFF
--- a/pyartcd/pyartcd/locks.py
+++ b/pyartcd/pyartcd/locks.py
@@ -8,12 +8,12 @@ from pyartcd import redis
 
 # Defines the pipeline locks managed by Redis
 class Lock(enum.Enum):
-    OLM_BUNDLE = 'olm-bundle-lock-{version}'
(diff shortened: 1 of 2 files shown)


thegreyd:
Should be merged at a time when there aren't any/many locks acquired and builds running since it can cause parallel runs

locriandev:
/unhold

I've created new lock by hand

PR #65, opened 2023-09-21
[build-sync] Mark viable=false as partial success

thegreyd:
Having an exit code different than 1 will enable us to mark job as Unstable
in cases where assembly is not viable, but pipeline did not crash with unexpected failure.

### DIFF
--- a/pyartcd/pyartcd/pipelines/build_sync.py
+++ b/pyartcd/pyartcd/pipelines/build_sync.py
@@ -16,6 +16,8 @@ from doozerlib.util import go_suffix_for_arch
 
 GEN_PAYLOAD_ARTIFACTS_OUT_DIR = 'gen-payload-artifacts'
 
+class ViableFalseError(Exception):
+    pass
 
 class BuildSyncPipeline:
     def __init__(self, runtime: Runtime, version: str, assembly: str, publish: bool, data_path: str,
@@ -256,7 +258,15 @@ class BuildSyncPipeline:
             cmd.extend([f'--exclude-arch {arch}' for arch in self.exclude_arches])
         if self.runtime.dry_run:
             cmd.extend(['--skip-gc-tagging', '--moist-run'])
-        await exectools.cmd_assert_async(cmd)
+
+        rc, stdout, stderr = await exectools.cmd_gather_async(cmd)
+        viable_false_message = "DO NOT PROCEED WITH THIS ASSEMBLY PAYLOAD"
+        if rc != 0:
+            if viable_false_message in stderr:
+                raise ViableFalseError(stderr)
+            else:
+                msg = f'Command {cmd} exited with code {rc}'
+                raise ChildProcessError(msg)
 
         # Populate CI imagestreams
         await self._populate_ci_imagestreams()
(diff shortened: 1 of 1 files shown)


pyartcd/pyartcd/pipelines/build_sync.py
```diff
@@ -255,7 +259,15 @@ async def _update_nightly_imagestreams(self):
             cmd.extend([f'--exclude-arch {arch}' for arch in self.exclude_arches])
         if self.runtime.dry_run:
             cmd.extend(['--skip-gc-tagging', '--moist-run'])
-        await exectools.cmd_assert_async(cmd)
+
+        rc, stdout, stderr = await exectools.cmd_gather_async(cmd)
```
locriandev, line 263: Running `cmd_gather_async` like this will capture stdout and stderr, and we won't be able to see them in the logs.
We currently need to capture stderr to check for viability, so we could change this into something like:
```
rc, _, stderr = await exectools.cmd_gather_async(cmd, stderr=None)
self.runtime.logger.info('doozer release:gen-payload output:\n%s, stderr)
```

However, we won't see any log until the process is completed, then get a full dump which doesn't quite help to understand what's going on. What we can do instead, is change [this line](https://github.com/openshift-eng/art-tools/blob/main/doozer/doozerlib/cli/release_gen_payload.py#L300) in doozer and have that message sent to stdout. Then, pyartcd would do something like:
```
rc, stdout, _ = await exectools.cmd_gather_async(cmd, stdout=None)
self.runtime.logger.info('doozer release:gen-payload output:\n%s, stdout)

File: pyartcd/pyartcd/pipelines/build_sync.py [hunk 2]
```diff
@@ -395,13 +407,18 @@ async def build_sync(runtime: Runtime, version: str, assembly: str, publish: boo
...
         # Only for 'stream' assembly, track failure to enable future notifications
         if assembly == 'stream':
             await pipeline.handle_failure()
 
-        # Re-reise the exception to make the job as failed
-        raise
+        # check if error is viable false, if it is set exit code to 2 to indicate partial success
+        if isinstance(e, ViableFalseError):
+            runtime.logger.error(f'Exiting with code 2 (partial success): {e}')
+            sys.exit(2)
```
locriandev, line 418: Optional: maybe we can use some standard Unix exit code. Took a quick look [here](https://tldp.org/LDP/abs/html/exitcodes.html) but haven't found a good match for our use case at first glance.

locriandev (changes requested):

pyartcd/pyartcd/pipelines/build_sync.py
```diff
@@ -255,7 +259,15 @@ async def _update_nightly_imagestreams(self):
             cmd.extend([f'--exclude-arch {arch}' for arch in self.exclude_arches])
         if self.runtime.dry_run:
             cmd.extend(['--skip-gc-tagging', '--moist-run'])
-        await exectools.cmd_assert_async(cmd)
+
+        rc, stdout, stderr = await exectools.cmd_gather_async(cmd)
+        viable_false_message = "DO NOT PROCEED WITH THIS ASSEMBLY PAYLOAD"
+        if rc != 0:
```
locriandev, line 265: If `rc != 0` `cmd_gather_async` will raise a `ChildProcessError` since `check` is True by default. We don't need to check it

PR #66, opened 2023-09-22
fix sha256sum not signed in retrigger job

Ximinhan:
if promote job failed with signing step due to UMB server broken but the sync to mirror step complete then if we retrigger the job with skip mirroring artifacts then the sha256sum.txt is not signed, add a check to get it not skiped
ref: https://redhat-internal.slack.com/archives/C02CJ4FNYSV/p1695370596029039?thread_ts=1693545269.704049&cid=C02CJ4FNYSV

### DIFF
--- a/pyartcd/pyartcd/pipelines/promote.py
+++ b/pyartcd/pyartcd/pipelines/promote.py
@@ -346,6 +346,14 @@ class PromotePipeline:
                 if not self.skip_mirror_binaries:
                     message_digests = await self.extract_and_publish_clients(client_type, release_infos)
                 if not self.skip_signing:
+                    if not message_digests:
+                        # if job retriggered due to UMB server issue, then sha256sum.txt may not get signed
+                        for arch, release_info in release_infos.items():
+                            if arch == "multi":
+                                message_digest = f"multi/clients/{client_type}/{release_name}/sha256sum.txt"
+                            else:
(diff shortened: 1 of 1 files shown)


pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -346,6 +346,14 @@ async def run(self):
                 if not self.skip_mirror_binaries:
                     message_digests = await self.extract_and_publish_clients(client_type, release_infos)
                 if not self.skip_signing:
+                    if not message_digests:
```
joepvd, line 349: I think this means we never sign on the first promote that gets this far, and always sign on the second promote that gets this far.

joepvd (changes requested):

pyartcd/pyartcd/pipelines/promote.py
Ximinhan: if mirror step not skipped the message digest list exist then will signing them.

vfreex:
Copied from Slack:
>The PR is not going to work. If message_digests is empty, it means syncing client binaries is skipped and there is no sha256sum to sign. Manually adding the paths will lead to a file not found error.
> I think client binaries need to go with gpg files. They need to be all published together or nothing. Currently the client binaries sync step is not reentrant. I am not sure if rerunning client binaries sync will invalidate the gpg files or not.

/hold

PR #67, opened 2023-09-22
Store job failures in redis

locriandev:
Jobs affected by this change are:
- ocp4
- ocp4-scan
- olm_bundle

Ideas in this PR:
- track in Redis failures for the jobs we're mostly interested in, as we already do for `build-sync`. Keeping them organized will let us implement scheduled checks and notifications as needed.

- as an example, `olm-bundle` [spams slack](https://redhat-internal.slack.com/archives/C04L1FPFHNH/p1694077550422039) as soon as a run fails; that could be improved, setting a threshold for each group, and starting to spam when that threshold is exceeded; we could also spam only for `stream`, and start pinging @release-artists without being too noisy.

- this would simplify manual checks/messages like [this one](https://redhat-internal.slack.com/archives/GDBRP5YJH/p1695302687889259)

- changing the fail count format to include a `failure` field will let us "count" anything else in the future, while keeping data organized in a hierarchical fashion

### DIFF
--- a/pyartcd/pyartcd/pipelines/build_sync.py
+++ b/pyartcd/pyartcd/pipelines/build_sync.py
@@ -35,7 +35,7 @@ class BuildSyncPipeline:
         self.skip_multiarch_payload = skip_multiarch_payload
         self.logger = runtime.logger
         self.working_dir = self.runtime.working_dir
-        self.fail_count_name = f'count:build-sync-failure:{assembly}:{version}'
+        self.fail_count_name = f'count:failure:build-sync:{assembly}:{version}'
 
     async def run(self):
         # Make sure we're logged into the OC registry
--- a/pyartcd/pyartcd/pipelines/build_sync.py
+++ b/pyartcd/pyartcd/pipelines/build_sync.py
@@ -400,7 +400,7 @@ async def build_sync(runtime: Runtime, version: str, assembly: str, publish: boo
         if assembly == 'stream':
             await pipeline.handle_failure()
 
-        # Re-reise the exception to make the job as failed
+        # Re-raise the exception to make the job as failed
         raise
 
     except RedisError as e:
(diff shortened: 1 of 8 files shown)


pyartcd/pyartcd/pipelines/build_sync.py
```diff
@@ -297,18 +297,11 @@ async def _publish(self, filename):
...
-            current_count = 0
-        fail_count = int(current_count) + 1
-        self.runtime.logger.info('Failure count for %s: %s', self.version, fail_count)
-
         # Update fail counter on Redis
-        await redis.set_value(self.fail_count_name, fail_count)
+        fail_count = await redis.increment_counter(self.fail_count_name)
 
-        # Less than 2 failures, assembly != stream: just break the build
-        if fail_count < 2 or self.assembly != 'stream':
```
thegreyd, line 311: Why did we change this?

pyartcd/pyartcd/pipelines/ocp4.py
```diff
@@ -852,7 +841,7 @@ async def _sweep(self):
                 content=f'Check Jenkins console for details: {jenkins.current_build_url}/console'
             )
 
-    def _report_success(self):
+    async def handle_success(self):
```
thegreyd, line 844: why is this async now?

pyartcd/pyartcd/pipelines/build_sync.py
locriandev: Line 393 already has
```
        if assembly == 'stream':
            await pipeline.handle_failure()
```
so no need to check that again

pyartcd/pyartcd/pipelines/ocp4.py
locriandev: Right, fixed

vfreex (commented):
/lgtm

PR #69, opened 2023-09-25
gen-payload: Require explicit permits for non-stream assemblies

joepvd:


locriandev (commented):
/lgtm

vfreex:
This PR itself LGTM. However I have a concern that we might want to keep this option for named assemblies in case we have emergencies. Also we can’t permit IMPERMISSIBLE issues:
https://github.com/openshift/art-tools/blob/bdc69378bc79899a9cf8570b92dd20bcbb8ac4b1/doozer/doozerlib/assembly.py#L339

/lgtm

vfreex:
Let’s update Doozer to not generate IMPERMISSIBLE for missing kernel tags first:
https://github.com/openshift-eng/art-tools/pull/70/files

PR #70, opened 2023-09-25
gen-assembly: Generate MISSING_SHIP_OK_TAG for kernels without ship-ok

vfreex:
tags

`IMPERMISSIBLE` is not generated anymore.


joepvd:
Maybe we should have this config under `stream.permits` in releases.yml. Not sure we want to have this in code. Think it is conceivable that the requirements are different for 4.12 and 4.15, and would be nice to be able to express that.

joepvd (approved):
/lgtm

PR #71, opened 2023-09-25
attach-cve-flaws: Permit explicitly included bugs

thegreyd:
This allows us to permit bugs that may have component-build mismatch due to some
reason, but are still valid to ship. This mismatch can be due to rpm components being
mentioned in image bugs or base image bugs whose builds we don't ship but still we
would want to ship the bugs

I also discovered that we can fail with a not obvious error when a subtask does not have any description 
set which is the reason for [327b8d3](https://github.com/openshift-eng/art-tools/pull/71/commits/327b8d3911a04fb48001694cec7fb958f091f650)

See https://github.com/openshift-eng/ocp-build-data/pull/3549
Test run: https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fprepare-release/858/console


vfreex (commented):
/lgtm

PR #78, opened 2023-09-27
Skip fail counter update for build-sync dry runs

locriandev:


pyartcd/pyartcd/pipelines/build_sync.py
```diff
@@ -55,7 +55,7 @@ async def run(self):
         await self._update_nightly_imagestreams()
 
         #  All good: delete fail counter
-        if self.assembly == 'stream':
+        if self.assembly == 'stream' and not self.runtime.dry_run:
```
thegreyd, line 58: It would be nice to handle success at the same place where we are handling failure, in the calling func below

locriandev: Done. Feel free to approve/merge when ready

PR #86, opened 2023-10-02
[ART-4932] refactor rhcos into artcommon

sosiouxme:
Since we need to remove the last vestiges of `machine-os-content` assumptions from RHCOS, and it's relatively simple code, it's the first thing to migrate to a new `artcommon` module.


art-common/artcommonlib/model.py
```diff
@@ -0,0 +1,159 @@
+from future.utils import as_native_str
```
locriandev, line 1: We should remove `doozerlib/model.py` and `elliottlib/model.py`

doozer/doozerlib/rhcos.py
```diff
@@ -13,85 +13,9 @@
 from doozerlib.repodata import OutdatedRPMFinder, Repodata
 from doozerlib.runtime import Runtime
 from doozerlib.util import brew_suffix_for_arch, isolate_el_version_in_release
+from artcommonlib import rhcos
 
 RHCOS_BASE_URL = "https://releases-rhcos-art.apps.ocp-virt.prod.psi.redhat.com/storage/releases"
```
locriandev, line 18: This constant is also defined in `elliottlib/constants.py` and could be moved in a common place (`artcommonlib/constants.py`?)

locriandev (changes requested):

art-common/artcommonlib/model.py
joepvd: We are in the future. Let's not import it.

art-common/README.md
```diff
@@ -0,0 +1 @@
+# commonlib - ART common library
```
joepvd, line 1: What is the purpose of this directory? Think the Makefile can or should be at the project root, and then there is no reason for this directory?

art-common/pyproject.toml
```diff
@@ -0,0 +1,60 @@
+[build-system]
```
joepvd, line 1: Would love to have this project definition one level up. Or what is the reason to declare this an independent package?

art-common/pyproject.toml
```diff
@@ -0,0 +1,60 @@
...
+]
+
+[project.urls]
+homepage = "https://github.com/openshift-eng/art-tools"
+repository = "https://github.com/openshift-eng/art-tools.git"
+
+[project.optional-dependencies]
+tests = [
+    # test reqs
+    "flexmock",
```
joepvd, line 54: If we can, let's get rid of flexmock,  or try not to reintroduce it.

art-common/pyproject.toml
```diff
@@ -0,0 +1,60 @@
...
+    "Operating System :: POSIX :: Linux",
+    "Operating System :: MacOS",
+    "Topic :: Software Development :: Libraries",
+    "Topic :: Software Development :: Bug Tracking",
+    "Topic :: Software Development :: Build Tools",
+    "Topic :: Utilities",
+    "Natural Language :: English",
+]
+dependencies = [
+    "future",
```
joepvd, line 43: Think we're in the future since we dropped support of python2.

doozer/doozerlib/rhcos.py
sosiouxme: sure... a few others as well

art-common/pyproject.toml
sosiouxme: right...

art-common/artcommonlib/runtime.py
```diff
@@ -0,0 +1,13 @@
+from abc import ABC, abstractmethod
+
+# an abstract class intended to be used by anything looking for config specific to a single
+# ocp-build-data group.
+
+
+class GroupRuntime(ABC):
```
vfreex, line 7: If possible, I'd like to not have a common Runtime (otherwise I could imagine that it would grow dramatically). Maybe we can have single-purpose classes and functions for loading group data over parent classes to inherit.

art-common/pyproject.toml
vfreex: I think having an independent package is better. For things like art-bot and maybe ocp-build-data-validator, they probably don't need the code and dependencies in elliott or doozer. If we keep everything in a single package, that means we have to install all python code and dependencies together.

Also there is another good reason for me: Elliott works on macOS but Doozer doesn't. I have elliott installed locally on my Mac but if we make everything as a single package, I won't be able to elliott natively :(

sosiouxme: working on it at https://github.com/openshift-eng/art-tools/pull/96 (includes this)

art-common/README.md
sosiouxme: added a Makefile in the top level, but for now keeping this as a separate project.

art-common/pyproject.toml
sosiouxme: closed #96; refactoring into a single project may be worthwhile but yuxiang makes good points and i think it's too big a change for this PR.

art-common/artcommonlib/runtime.py
sosiouxme: i would like to have a common Runtime for things that are actually... common. but not a single consolidated Runtime for everything. even elliott and doozer are different enough to warrant different runtimes, and pyartcd is even more so because it cannot assume group and assembly.

my thought was that `pyartcd.runtime.GroupRuntime` (parallel to the normal runtime), `doozer.runtime`, and `elliott.runtime` would all implement this abstract class, so that when you need to pass in "a runtime" to something related to group and assembly (like the `rhcos` module takes), you know that runtime at least has this class's properties to work with. i didn't yet refactor GroupRuntime into elliott/doozer runtimes but it wouldn't take long...

sosiouxme: @vfreex there, i put it in 349fed6afd11d6a6e39f9d7f711129fe3b62262e

now, if you really want it to be a separate concept from the other Runtimes, I think it would take a bit of work to untangle it from them, but I can easily just remove that commit and leave it as an adjacent-but-separate concept that elliott and doozer can implement later. what do you think?

vfreex: Your explanation persuaded me. Thanks

vfreex (commented):
/lgtm

vfreex:
@sosiouxme I updated github actions to accommodate setuptools_scm and it caused a conflict. Could you rebase this?

PR #88, opened 2023-10-03
Move cincinnati PR creation to promote.py

thegreyd:
ec.0 promote job [failed](https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fpromote-assembly/857/) on readJSON jenkins pipeline step flaking for perfectly valid json, rerunning it did the job - but [missing](https://redhat-internal.slack.com/archives/CJARLA942/p1696352245035539) cincinnati PRs can cause unecessary delays. The remaining steps outside of pyartcd promote.py are sending release message on umb, create cincinnati PRs, validating rhsas - I would like to move the cincinnati step inside pyartcd, so all important steps are done before the possibility of flaking on json

Should go with https://github.com/openshift-eng/aos-cd-jobs/pull/3983
Requesting /lgtm

### DIFF
--- a/pyartcd/pyartcd/jenkins.py
+++ b/pyartcd/pyartcd/jenkins.py
@@ -30,6 +30,7 @@ class Jobs(Enum):
     SYNC_FOR_CI = 'scheduled-builds/sync-for-ci'
     MICROSHIFT_SYNC = 'aos-cd-builds/build%2Fmicroshift_sync'
     SCAN_OSH = 'aos-cd-builds/build%2Fscan-osh'
+    CINCINNATI_PRS = 'aos-cd-builds/build%2Fcincinnati-prs'
 
 
 def init_jenkins():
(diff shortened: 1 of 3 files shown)


pyartcd/pyartcd/jenkins.py
```diff
@@ -279,6 +280,20 @@ def start_build_sync(build_version: str, assembly: str, doozer_data_path: Option
...
+def start_cincinnati_prs(from_releases: list, release_name: str, advisory_id: int,
+                         candidate_pr_note: str, skip_ota_notification, **kwargs) -> Optional[str]:
+    return start_build(
+        job=Jobs.CINCINNATI_PRS,
+        params={
+            'FROM_RELEASE_TAG': ','.join(from_releases),
+            'RELEASE_NAME': release_name,
+            'ADVISORY_NUM': advisory_id,
+            'CANDIDATE_PR_NOTE': candidate_pr_note,
+            'SKIP_OTA_SLACK_NOTIFICATION': skip_ota_notification
```
ashwindasr, line 292: Just wanted to note that the job that's being called has a `GITHUB_ORG` param with a default value of `openshift`. Since joep mentioned that the repos will soon be migrated, it might be a good idea to explicitly pass that so that we can keep track of this? Just a nit though

thegreyd: Added :+1:

PR #90, opened 2023-10-05
Add support for Redis lists in pyartcd

ashwindasr:
Could potentially need this to store pending task IDs of SAST scans. Using Redis would be much easier to setup than a database.

Usecase:
SAST scans take a lot of time to finish. Hence after we kick them off we can store the value in Redis. In the beginning of each `scan-osh` job run, we can check to see if the pending tasks has finished. If it has we can take the next steps, which maybe sending out emails (this hasn't been finalized yet) 

_Requesting /lgtm and /approve_

### DIFF
--- a/pyartcd/pyartcd/redis.py
+++ b/pyartcd/pyartcd/redis.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import typing
 from functools import wraps
 from string import Template
 
@@ -39,6 +40,7 @@ def handle_connection(func):
         res = await func(conn, *args, **kwargs)
         conn.close()
         return res
+
     return wrapper
 
 
@@ -85,3 +87,45 @@ async def delete_key(conn: aioredis.commands.Redis, key: str) -> int:
     res = await conn.delete(key)
     logger.debug('Key %s %s', key, 'deleted' if res else 'not found')
     return res
+
+
+@handle_connection
+async def list_push(conn: aioredis.commands.Redis, key: str, value: str) -> None:
+    """
+    Push to a list in Redis, like a queue (FIFO). Docs: https://redis.io/docs/data-types/lists/
+    """
+
+    logger.debug('Pushing value %s to list %s', value, key)
+    await conn.lpush(key, value)
+
+
+@handle_connection
+async def list_pop(conn: aioredis.commands.Redis, key: str) -> typing.Optional[str]:
+    """
+    Pop a value from a list, like a queue (FIFO). Docs: https://redis.io/docs/data-types/lists/
+    :return: A string if a value is present, else None
+    """
+
+    value = await conn.rpop(key)
+    logger.debug('List %s has value %s', key, value)
+    return value
+
+
+@handle_connection
+async def list_see_all(conn: aioredis.commands.Redis, key: str) -> typing.Optional[list]:
+    """
+    Get all values from the Redis list. Does not pop values, list will still exist in Redis
+
+    :return: List of values from the key, or None if the list is empty or does not exist.
+    """
+    values = await conn.lrange(key, 0, -1)
+
+    return values
+
+
+async def list_push_all(key: str, values: list) -> None:
+    """
+    Push all values from list to the Redis key
+    """
+    for value in values:
+        await list_push(key, value)


pyartcd/pyartcd/redis.py
```diff
@@ -85,3 +86,24 @@ async def delete_key(conn: aioredis.commands.Redis, key: str) -> int:
...
+    await conn.lpush(key, value)
+
+
+@handle_connection
+async def list_pop(conn: aioredis.commands.Redis, key: str) -> str:
+    """
+    Pop a value from a list, like a queue (FIFO). Docs: https://redis.io/docs/data-types/lists/
+    """
+
+    value = conn.rpop(key)
```
locriandev, line 107: Needs an `await`

pyartcd/pyartcd/redis.py
```diff
@@ -85,3 +86,24 @@ async def delete_key(conn: aioredis.commands.Redis, key: str) -> int:
...
+    """
+    Push to a list in Redis, like a queue (FIFO). Docs: https://redis.io/docs/data-types/lists/
+    """
+
+    logger.debug('Pushing value %s to list %s', value, key)
+    await conn.lpush(key, value)
+
+
+@handle_connection
+async def list_pop(conn: aioredis.commands.Redis, key: str) -> str:
```
locriandev, line 102: It will return `None` if the list is empty or does not exist. I'd state that in the docstring, and change the return type to `typing.,Optional[str]` (could be done for other functions too)

locriandev (changes requested):

pyartcd/pyartcd/redis.py
```diff
@@ -85,3 +86,51 @@ async def delete_key(conn: aioredis.commands.Redis, key: str) -> int:
...
+    """
+    Push to a list in Redis, like a queue (FIFO). Docs: https://redis.io/docs/data-types/lists/
+    """
+
+    logger.debug('Pushing value %s to list %s', value, key)
+    await conn.lpush(key, value)
+
+
+@handle_connection
+async def list_pop(conn: aioredis.commands.Redis, key: str) -> (str, None):
```
locriandev, line 102: This means a return type of `tuple(string, None)`. We do need to use `typing.Optional`, there's places where it is used you can take inspiration from

pyartcd/pyartcd/redis.py
```diff
@@ -85,3 +86,51 @@ async def delete_key(conn: aioredis.commands.Redis, key: str) -> int:
...
+
+@handle_connection
+async def list_pop_all(conn: aioredis.commands.Redis, key: str) -> list:
+    """
+    Get all values from the Redis list
+
+    :return: List of values from the key
+    """
+    values = []
+    while True:
```
locriandev, line 121: I think there should be some function in redis to specifically read/write lists. This looks quite inefficient

locriandev (changes requested):

pyartcd/pyartcd/redis.py
ashwindasr: Found the read one! thank you! https://github.com/openshift-eng/art-tools/pull/90/commits/a6b9789f3280c3d5f2a8a507226a921d1f223a3f

Couldn't find the bulk write one though

pyartcd/pyartcd/redis.py
ashwindasr: like this right https://github.com/openshift-eng/art-tools/pull/90/commits/5548fd127628fab3ad7bb6b4685091107ebdb2dd

ashwindasr:
Tested. Values show up in redis commander.

pyartcd/pyartcd/redis.py
```diff
@@ -85,3 +87,45 @@ async def delete_key(conn: aioredis.commands.Redis, key: str) -> int:
...
+
+    return values
+
+
+async def list_push_all(key: str, values: list) -> None:
+    """
+    Push all values from list to the Redis key
+    """
+    for value in values:
+        await list_push(key, value)
```
locriandev, line 131: Don't we have something similar to `lrange` that pushes values instead?

pyartcd/pyartcd/redis.py
```diff
@@ -85,3 +87,45 @@ async def delete_key(conn: aioredis.commands.Redis, key: str) -> int:
...
+    :return: A string if a value is present, else None
+    """
+
+    value = await conn.rpop(key)
+    logger.debug('List %s has value %s', key, value)
+    return value
+
+
+@handle_connection
+async def list_see_all(conn: aioredis.commands.Redis, key: str) -> list:
```
locriandev, line 115: Will still return `None` if the list is empty or does not exist. We should state this in the docs and update the return type

locriandev (changes requested):

pyartcd/pyartcd/redis.py
ashwindasr: I tried, but sadly I couldn't find the command to push an entire list at once

locriandev (commented):
/lgtm

ashwindasr:
Squashed commits

locriandev:
> Squashed commits, readding lgtm

LGTM, but how did you get that label re-added?

ashwindasr:
manually haha

PR #94, opened 2023-10-10
Announce mass rebuild in slack

joepvd:


pyartcd/pyartcd/pipelines/ocp4.py
```diff
@@ -866,6 +866,7 @@ async def run(self):
         # Rebase and build images
         if self.build_plan.build_images:
             if self.mass_rebuild:
+                await self._slack_client.say(f'Mass rebuild for {self.version}')
```
Ximinhan, line 869: the message `:construction: Mass rebuild for {self.version} start :construction:` seem more clear
and also announce the build complete

joepvd: Updated PR. Voting against a notification with "rebuild done", as that does not seem actionable. Think we can improve here, but I suppose that would be better suited with structured logging and queries and alerts from that space.

Ximinhan:
would you mind rebase your commit?

pyartcd/pyartcd/pipelines/ocp4.py
joepvd: I want to be careful with the amount of notifications. I think "There is a mass rebuild for versions x, y, z" is a good improvement over what we have now. Then we know what maybe should be actively managed. 

Extra details on exact state would indeed be nice, but not sure it is worth extra messages. 

IMHO, let's merge this and see if and how we are improving our handling and awareness. We can tweak later, also with extra tech like structured logging.

PR #95, opened 2023-10-10
Exclude certain keys from config digest calculation

vfreex:
So that making a change to `scan_sources` or `content.source.ci_alignment` will not trigger a rebuild caused by config change.

### DIFF
--- a/doozer/doozerlib/image.py
+++ b/doozer/doozerlib/image.py
@@ -5,6 +5,7 @@ import json
 import logging
 from typing import (Any, Awaitable, Dict, List, Optional, Set, Tuple, Union,
                     cast)
+from copy import deepcopy
 
 import doozerlib
 from doozerlib import brew, coverity, exectools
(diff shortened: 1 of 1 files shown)


joepvd:
Setting 
/hold
...as deploy would mean mass rebuild, so needs some care.

joepvd:
Might want to take out `ci_alignment` from `content.source`. Suspect that we'll see movement there kind of soon-ish. But also fine if that is not in the current scope. 

/approve

doozer/doozerlib/image.py
```diff
@@ -459,11 +460,25 @@ def covscan(self, cc: coverity.CoverityContext) -> bool:
                 return False
 
     def calculate_config_digest(self, group_config, streams):
-        image_config: Dict[str, Any] = self.config.primitive()  # primitive() should create a shallow clone for the underlying dict
+        ignore_keys = ["owners", "scan_sources", "content.source.ci_alignment"]  # list of keys that shouldn't be involved in config digest calculation
```
locriandev, line 463: Do you think we can also ignore `content.source.git` fields? If there's a change in repo or Dockerfile path, scan-sources should notice in advance that SHAs do not match latest builds, and mark the component as changed.

 This could avoid mass rebuilds with PRs like [4.15 ocp-build-data moved to openshift-eng](https://github.com/openshift-eng/ocp-build-data/pull/3643)

locriandev (changes requested):

doozer/doozerlib/image.py
locriandev, line 463:
```suggestion
        ignore_keys = ["owners", "scan_sources", "content.source.ci_alignment", "content.source.git"]  # list of keys that shouldn't be involved in config digest calculation
```

PR #96, opened 2023-10-10
[wip] refactor art tools

sosiouxme:
Includes https://github.com/openshift-eng/art-tools/pull/86

Brings all art-tools packages under a single project.


sosiouxme:
this worked except for 6 test cases i needed to fix.

it might still be worthwhile to consolidate like this (removing all the cruft that i kept here) but i think that big a change will block the success of my short term goal for ART-4932

PR #98, opened 2023-10-11
use Release Controller API get payload info

Ximinhan:
The most usage of the parse_release_payloads function is to get payload digest from release, we currently trigger oc cli to parse result. This PR is to get the digest from the Release Controller which is simpler and more direct and don't need to consider authentication issues.

test job: https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fbuild-microshift/639/

### DIFF
--- a/pyartcd/pyartcd/constants.py
+++ b/pyartcd/pyartcd/constants.py
@@ -17,6 +17,12 @@ NIGHTLY_PAYLOAD_REPOS = {
     "ppc64le": "registry.ci.openshift.org/ocp-ppc64le/release-ppc64le",
     "aarch64": "registry.ci.openshift.org/ocp-arm64/release-arm64",
 }
+RELEASE_CONTROLLER_ENDPOINT = {
+    "x86_64": "https://amd64.ocp.releases.ci.openshift.org/api/v1/releasestream",
+    "aarch64": "https://arm64.ocp.releases.ci.openshift.org/api/v1/releasestream",
+    "s390x": "https://s390x.ocp.releases.ci.openshift.org/api/v1/releasestream",
+    "ppc64le": "https://ppc64le.ocp.releases.ci.openshift.org/api/v1/releasestream",
+}
 
 # Maps the name of a release component tag to the filename element to include
 # when creating artifacts on mirror.openshift.com.
--- a/pyartcd/pyartcd/pipelines/build_microshift.py
+++ b/pyartcd/pyartcd/pipelines/build_microshift.py
@@ -3,6 +3,7 @@ import json
 import logging
 import os
 import re
+import requests
 import traceback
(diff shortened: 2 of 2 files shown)


pyartcd/pyartcd/constants.py
```diff
@@ -17,6 +17,12 @@
     "ppc64le": "registry.ci.openshift.org/ocp-ppc64le/release-ppc64le",
     "aarch64": "registry.ci.openshift.org/ocp-arm64/release-arm64",
 }
+RELEASE_CONTROLLER_ENDPOINT = {
```
locriandev, line 20: This can be a string Template, we would then replace the arch where we need

locriandev (commented):
LGTM with one nit

pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -267,28 +268,20 @@ async def _change_advisory_status(self):
...
-        pullspecs = []
         for payload in payloads:
-            if "/" not in payload:
-                # Convert nightly name to pullspec
-                # 4.12.0-0.nightly-2022-10-25-210451 ==> registry.ci.openshift.org/ocp/release:4.12.0-0.nightly-2022-10-25-210451
-                _, brew_cpu_arch, _ = isolate_nightly_name_components(payload)
-                pullspecs.append(constants.NIGHTLY_PAYLOAD_REPOS[brew_cpu_arch] + ":" + payload)
+            if "arm64" in payload:
+                brew_arch = "aarch64"
+                res = requests.get(f"{constants.RELEASE_CONTROLLER_ENDPOINT[brew_arch]}/{payload.split('-')[0]}-0.nightly-arm64/release/{payload}")
```
vfreex, line 276: This assumes a certain pullspec format like `4.12.0-0.nightly-2022-10-25-210451`. This is probably not a big issue for automated stream build but we will certainly lose the flexibility to rebase onto any other kinds of payloads (like named releases or sha256sum).

pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -267,28 +268,20 @@ async def _change_advisory_status(self):
         await exectools.cmd_assert_async(cmd, env=self._elliott_env_vars)
 
     @staticmethod
-    async def parse_release_payloads(payloads: Iterable[str]):
+    def parse_release_payloads(payloads: Iterable[str]):
```
vfreex, line 271: Use async. Calling a sync function from sync will block the event loop.

vfreex:
Per discussion at https://redhat-internal.slack.com/archives/C048BG69N8H/p1696984071357359, this is no longer needed.

PR #101, opened 2023-10-13
Notify about mass rebuilds when they actually start/complete

locriandev:
Notifying about a mass rebuild before acquiring the lock can be misleading. This PR proposes to be a little more verbose (acceptable in case of mass rebuilds, that do not happen every day) and make it clear:
- when a version wants to mass rebuild (acquiring the lock)
- when the mass rebuild has actually started (lock acquired)
- when the mass rebuild has terminated (lock released)

Follows https://github.com/openshift-eng/art-tools/pull/94

### DIFF
--- a/pyartcd/pyartcd/pipelines/ocp4.py
+++ b/pyartcd/pyartcd/pipelines/ocp4.py
@@ -5,7 +5,6 @@ import traceback
 
 import click
 import yaml
-from aioredlock import LockError
 
 from pyartcd import locks, util, plashets, exectools, constants, \
     jenkins, record as record_util, oc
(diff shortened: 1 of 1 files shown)


pyartcd/pyartcd/pipelines/ocp4.py
```diff
@@ -866,7 +871,8 @@ async def run(self):
         # Rebase and build images
         if self.build_plan.build_images:
             if self.mass_rebuild:
-                await self._slack_client.say(f':construction: Mass rebuild for {self.version} :construction:')
+                await self._slack_client.say(
+                    f':loading-correct: Enqueing mass rebuild for {self.version} :loading-correct:')
```
joepvd, line 875: Enqueueing

joepvd: Isn't this triggering at virtually the same moment as the message from line 714?

locriandev: fixed

joepvd (approved):
/lgtm
Nice!

PR #102, opened 2023-10-13
key source is not required

Ximinhan:
fix modification scripts error

### DIFF
--- a/doozer/doozerlib/source_modifications.py
+++ b/doozer/doozerlib/source_modifications.py
@@ -82,7 +82,7 @@ class AddModifier(object):
         :param doozer_source: Relative source path within doozer
 
         """
-        self.source = kwargs["source"]
+        self.source = kwargs.get("source", None)
         self.path = kwargs["path"]
         self.overwriting = kwargs.get("overwriting", False)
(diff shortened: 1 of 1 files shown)


Ximinhan:
ut result is not accurate, the test job is https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fgolang-builder/197/

doozer/doozerlib/source_modifications.py
```diff
@@ -125,7 +125,7 @@ def act(self, *args, **kwargs):
                 else:
                     raise ValueError("Unknown 'validate' value: {self.validate}")
         else:  # use doozer local source
-            content = Path(Path(__file__).parent, self.doozer_source).read_text()
+            content = Path(Path(__file__).parent, self.doozer_source).read_text().encode()
```
vfreex, line 128: Instead of calling `read_text` then `encode`, we need to use `read_bytes` here in case it is a binary file. Otherwise LGTM

Ximinhan: updated

vfreex (commented):
/lgtm

PR #103, opened 2023-10-13
ART-3506 Rebase openshift-priv if needed

locriandev:
When scanning sources, and before checking for any changes, we'll do a rebase from `openshift` into `openshift-priv` if the public upstream has newer contents. This will let us build the latest updates and reduce delays between PRs merge time and their actual availability in a nightly.

This is the proposed workflow:
- check if a component has a public upstream; if not, no need to rebase. Otherwise,
- check if commit SHAs on public and private upstreams match; if they do, no need to rebase (quick API call, no need to clone anything. Otherwise, clone repo and
- check if public upstream is an ancestor of the private one. If so, private has newer content, thus no need to rebase. Otherwise
- attempt a fast-forward; if successful, push to priv and mark the repo as changed (will trigger a rebuild). Otherwise,
- attempt to create a merge commit from public into priv; if successful, push to priv and mark the repo as changed. Otherwise,
- if no rebase was possible, priv needs manual reconciliation. Send a notification on Slack and continue with the scan.

If anything else goes wrong (e.g. push to priv fails) send a notification on Slack and continue with the scan.

### DIFF
--- a/doozer/doozerlib/cli/scan_sources.py
+++ b/doozer/doozerlib/cli/scan_sources.py
@@ -1,28 +1,37 @@
 import asyncio
+import os
 from datetime import datetime, timezone
 
 import click
 import yaml
 from typing import List, Tuple, Optional
 
-from doozerlib import brew, exectools, rhcos, util
-from doozerlib.cli import cli, pass_runtime, click_coroutine
+from github import Github
+
+from doozerlib import brew, exectools, rhcos, util, constants
+from doozerlib.cli import cli, pass_runtime
 from doozerlib.cli import release_gen_payload as rgp
 from doozerlib.image import ImageMetadata
 from doozerlib.metadata import RebuildHint, RebuildHintCode, Metadata
+from doozerlib.model import Missing
+from doozerlib.pushd import Dir
 from doozerlib.runtime import Runtime
 
 
 class ConfigScanSources:
-    def __init__(self, runtime: Runtime, ci_kubeconfig: str, as_yaml: bool):
+    def __init__(self, runtime: Runtime, ci_kubeconfig: str, as_yaml: bool,
+                 rebase_priv: bool = False, dry_run: bool = False):
         self.runtime = runtime
         self.ci_kubeconfig = ci_kubeconfig
         self.as_yaml = as_yaml
+        self.rebase_priv = rebase_priv
+        self.dry_run = dry_run
 
         self.changing_rpm_metas = set()
         self.changing_image_metas = set()
         self.changing_rpm_packages = set()
         self.assessment_reason = dict()  # maps metadata qualified_key => message describing change
+        self.issues = list()  # tracks issues that arose during the scan, which did not interrupt the job
 
         self.all_rpm_metas = set(runtime.rpm_metas())
         self.all_image_metas = set(runtime.image_metas())
@@ -31,13 +40,19 @@ class ConfigScanSources:
         self.oldest_image_event_ts = None
         self.newest_image_event_ts = 0
 
+        self.github_client = Github(login_or_token=os.getenv(constants.GITHUB_TOKEN))
+
     def run(self):
         with self.runtime.shared_koji_client_session() as koji_api:
             self.runtime.logger.info(f'scan-sources coordinate: brew_event: '
                                      f'{koji_api.getLastEvent(brew.KojiWrapperOpts(brew_event_aware=True))}')
             self.runtime.logger.info(f'scan-sources coordinate: emulated_brew_event: {self.runtime.brew_event}')
 
-            # First, scan for any upstream source code changes. If found, these are guaranteed rebuilds.
+            # First, try to rebase into openshift-priv to reduce upstream merge -> downstream build time
+            if self.rebase_priv:
+                self.rebase_into_priv()
+
+            # Then, scan for any upstream source code changes. If found, these are guaranteed rebuilds.
             self.scan_for_upstream_changes(koji_api)
 
             # Check for other reasons why images should be rebuilt (e.g. config changes, dependencies)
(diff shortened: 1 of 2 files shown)


doozer/doozerlib/cli/scan_sources.py
```diff
@@ -54,6 +68,142 @@ def run(self):
...
+        pub_commit = self.github_client.get_repo(f'{public_org}/{repo_name}').get_branch(branch_name).commit
+
+        # Check private commit ID
+        priv_commit = self.github_client.get_repo(f'{priv_org}/{repo_name}').get_branch(branch_name).commit
+
+        if pub_commit.sha == priv_commit.sha:
+            self.runtime.logger.debug('Commit SHAs match on public and priv upstreams for %s', repo_name)
+            return True
+
+        self.runtime.logger.info('Commit SHAs do not match on public and priv upstreams for %s: '
```
thegreyd, line 146: Latest Commit SHAs? or maybe just "Latest commits do not match.."

doozer/doozerlib/cli/scan_sources.py
```diff
@@ -54,6 +68,142 @@ def run(self):
...
+            message = f'{metadata.name} reconciled with public openshift/{repo_name}'
+            rebuild_hint = RebuildHint(RebuildHintCode.REBASE_INTO_PRIV, message)
+            self.runtime.logger.info(message)
+
+            if metadata.meta_type == 'rpm':
+                self.add_assessment_reason(metadata, rebuild_hint)
+                self.changing_rpm_metas.add(metadata)
+
+            elif metadata.meta_type == 'image':
+                if rebuild_hint.rebuild:
```
thegreyd, line 118: Why not have this before we check the meta_type?

doozer/doozerlib/cli/scan_sources.py
```diff
@@ -54,6 +68,142 @@ def run(self):
...
+        # Attempt a fast-forward merge
+        rc, _, _ = exectools.cmd_gather(cmd=['git', 'pull', '--ff-only', 'public_upstream', branch_name])
+        if not rc:
+            # fast-forward succeeded, will push to openshift-priv
+            self.runtime.logger.info('Fast-forwarded %s from public_upstream/%s', metadata.name, branch_name)
+            reconciled = True
+
+        else:
+            # fast-forward failed, trying a merge commit
+            rc, _, _ = exectools.cmd_gather(cmd=['git', 'merge', f'public_upstream/{branch_name}',
```
thegreyd, line 83: Would this be logging stdout/stderr? Would like that to help us investigate issue

doozer/doozerlib/cli/scan_sources.py
locriandev: Updated

doozer/doozerlib/cli/scan_sources.py
locriandev: Updated

thegreyd:
Changes lgtm overall
Can we test it out in hack job?

doozer/doozerlib/cli/scan_sources.py
```diff
@@ -54,6 +68,143 @@ def run(self):
...
+            else:
+                raise IOError(f'Unsupported meta type: {metadata.meta_type}')
+
+        except ChildProcessError:
+            # Failed pushing to openshift-priv
+            self.runtime.logger.warning('failed pushing to openshift-priv for %s', metadata.name)
+            self.issues.append({'name': metadata.distgit_key,
+                                'issue': 'Failed pushing to openshift-priv'})
+
+    def _do_shas_match(self, public_org, priv_org, repo_name, branch_name) -> bool:
```
vfreex, line 131: We can't assume the branch names are the same in public and private repo. Let's separate them:
```
    def _do_shas_match(self, public_org, public_repo_name, public_branch_name, priv_org, priv_repo_name, priv_branch_name) -> bool:
```

doozer/doozerlib/cli/scan_sources.py
```diff
@@ -54,6 +68,143 @@ def run(self):
...
+
+        for metadata, public_upstream in upstream_mappings:
+            if not public_upstream:
+                # Upstream repo does not have a public counterpart: no need to rebase
+                self.runtime.logger.warning('%s %s does not have a public upstream: skipping openshift-priv rebase',
+                                            metadata.meta_type, metadata.name)
+                continue
+
+            public_url, branch_name = public_upstream
+            branch_name = branch_name or metadata.config.content.source.git.branch.target
```
vfreex, line 186: Sometimes `branch.target` is a sha. I think we need to skip in that case (unless we have a good reason to handle that well).

doozer/doozerlib/cli/scan_sources.py
```diff
@@ -54,6 +68,143 @@ def run(self):
         # We have our information. Now build and print the output report
         self.generate_report()
 
+    def _try_reconciliation(self, metadata: Metadata, repo_name: str, branch_name: str):
+        reconciled = False
+
+        # Attempt a fast-forward merge
+        rc, _, _ = exectools.cmd_gather(cmd=['git', 'pull', '--ff-only', 'public_upstream', branch_name])
+        if not rc:
```
vfreex, line 76: `rc` is a number. let's not use it like a boolean.

doozer/doozerlib/cli/scan_sources.py
```diff
@@ -54,6 +68,143 @@ def run(self):
...
+            # fast-forward succeeded, will push to openshift-priv
+            self.runtime.logger.info('Fast-forwarded %s from public_upstream/%s', metadata.name, branch_name)
+            reconciled = True
+
+        else:
+            # fast-forward failed, trying a merge commit
+            rc, _, _ = exectools.cmd_gather(cmd=['git', 'merge', f'public_upstream/{branch_name}',
+                                                 '-m', f'Reconciliation with public openshift/{repo_name}'],
```
vfreex, line 84: Well, the public org name may not be openshift. Let's be more accurate.

doozer/doozerlib/cli/scan_sources.py
```diff
@@ -307,7 +458,9 @@ def generate_report(self):
         if self.as_yaml:
             click.echo('---')
             click.echo(yaml.safe_dump(results, indent=4))
+            click.echo(yaml.safe_dump({'issues': self.issues}, indent=4))
```
vfreex, line 461: This will dump multiple yaml documents into stdout. Not sure if it will create a subtle issue. This may need more testing, or we can combine `issues` to the previous yaml document.

doozer/doozerlib/cli/scan_sources.py
```diff
@@ -412,8 +571,8 @@ def config_scan_source_changes(runtime: Runtime, ci_kubeconfig, as_yaml):
     It will report RHCOS updates available per imagestream.
     """
 
-    runtime.initialize(mode='both', clone_distgits=False, clone_source=False, prevent_cloning=True)
-    ConfigScanSources(runtime, ci_kubeconfig, as_yaml).run()
+    runtime.initialize(mode='both', clone_distgits=False, clone_source=False, prevent_cloning=False)
```
vfreex, line 574: Previously `prevent_cloning` was used to ensure we don't clone upstream or distgit repos in scan-sources. With this PR, this job will clone upstream repos when there is a need to rebase into the private repo.
I think we need to be cautious when changing this to False. More testing is needed to ensure we don't clone upstream repos unless there is a need, and ensure that we never clone distgit repos.

doozer/doozerlib/cli/scan_sources.py
```diff
@@ -1,28 +1,36 @@
...
 
 
 class ConfigScanSources:
-    def __init__(self, runtime: Runtime, ci_kubeconfig: str, as_yaml: bool):
+    def __init__(self, runtime: Runtime, ci_kubeconfig: str, as_yaml: bool,
+                 rebase: bool = False, dry_run: bool = False):
         self.runtime = runtime
         self.ci_kubeconfig = ci_kubeconfig
         self.as_yaml = as_yaml
+        self.rebase = rebase
```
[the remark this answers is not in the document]
locriandev: Done with 0ae5ac867c67c46671e797b2bc67a9216e525fd5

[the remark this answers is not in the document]
locriandev: Done with 0ae5ac867c67c46671e797b2bc67a9216e525fd5

doozer/doozerlib/cli/scan_sources.py
locriandev: In Python it's broadly accepted (and even suggested) to run such checks against non-boolean types such as strings, lists, and objects in general. I'd like to leave this as-is. Besides, we use the same pattern everywhere across our code.

doozer/doozerlib/cli/scan_sources.py
locriandev: Done with e2b238c9c01bf1a95d40be9c5fec5090e61af359

doozer/doozerlib/cli/scan_sources.py
locriandev: `prevent_cloning` needs to be set to False, as I verified we'd fall in this codepath:
```
    def clone(self, distgits_root_dir, distgit_branch):
        if self.metadata.prevent_cloning:
            raise IOError('...')
```
I also verified that upstream repos are only cloned when needed during rebase to priv, and that distgits are never cloned.

doozer/doozerlib/cli/scan_sources.py
locriandev: Done with aa34afcf3f39807ee8296454145cfc50ca846907

I chose to apply a minimal change, and let the user find the public upstream if interested. In most cases (if not all), this will be `openshift` so I think it's an acceptable compromise to minimize changes.

doozer/doozerlib/cli/scan_sources.py
```diff
@@ -54,6 +68,143 @@ def run(self):
...
+                return False
+
+        self.runtime.logger.info('Private upstream is ahead of public for %s: no need to rebase', repo_name)
+        return True
+
+    def rebase_into_priv(self):
+        self.runtime.logger.info('Rebasing public upstream contents into openshift-priv')
+
+        upstream_mappings = exectools.parallel_exec(
+            lambda meta, _: (meta, self.runtime.get_public_upstream(meta.config.content.source.git.url)),
```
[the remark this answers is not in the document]
locriandev: Global aliases have been recently removed (see for example [4.12 Remove global aliases](https://github.com/openshift-eng/ocp-build-data/pull/3620), and all its siblings).

Handled distgit-only components in 29f3d31f69658df888b7801be12ca1ab02c679ac

doozer/doozerlib/cli/scan_sources.py
```diff
@@ -54,6 +68,143 @@ def run(self):
...
+            self.all_metas,
+            n_threads=20,
+        ).get()
+
+        for metadata, public_upstream in upstream_mappings:
+            if not public_upstream:
```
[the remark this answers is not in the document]
locriandev: Done with 5afb92d653e7c08ccee12bfd2a98fd107b5e2025

I checked that, like you said, in case `public_upstreams` is not  defined that function will just return the same upstreams as defined in `content.source.git.url`. In this case, we can skip the rebase if the public upstream (as returned by `get_public_upstream()`) match the private one

doozer/doozerlib/cli/scan_sources.py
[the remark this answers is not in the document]
vfreex: There are 2 kinds of aliases. Besides the global alias you have mentioned, distgit names are also valid aliases. Not sure if we still use that too.

doozer/doozerlib/cli/scan_sources.py
```diff
@@ -54,6 +69,162 @@ def run(self):
...
+            message = f'{metadata.name} reconciled with public openshift/{repo_name}'
+            rebuild_hint = RebuildHint(RebuildHintCode.REBASE_INTO_PRIV, message)
+            self.runtime.logger.info(message)
+
+            if metadata.meta_type == 'rpm':
+                self.add_assessment_reason(metadata, rebuild_hint)
+                self.changing_rpm_metas.add(metadata)
+
+            elif metadata.meta_type == 'image':
+                self.add_image_meta_change(metadata, rebuild_hint)
```
thegreyd, line 121: After today's excursion with https://github.com/openshift-eng/art-tools/pull/112 , I'm not sure if we should add this hint here. Because `scan_for_upstream_changes` which gets run after this will detect that there is no build for the commit, it also checks if image/rpm isn't `mode:disabled ` (which I think we can do in several ways). There's also `needs_rebuild` and `RebuildHintCode.DELAYING_NEXT_ATTEMPT` which seem like special cases where we don't want to rebuild.

thegreyd: This reminds me maybe we shouldn't sync disabled images/rpms..

locriandev: Turns out doozer will exclude disabled images by default (they do not appear in the image map)

locriandev: skipping rebase for disabled rpms/images: 0fb153af57bf22d1c6e1fdd4de6ec9a3172a0ccd

locriandev: Regarding the rebuild hint, good catch. Removed in 6c659302ff725c6744b6a6cb1fa97fb44d3f43b6

doozer/doozerlib/metadata.py
```diff
@@ -69,6 +69,7 @@ class RebuildHintCode(Enum):
     PACKAGE_CHANGE = (True, 13)
     ARCHES_CHANGE = (True, 14)
     DEPENDENCY_NEWER = (True, 15)
+    REBASE_INTO_PRIV = (True, 16)
```
thegreyd, line 72: Leftover

thegreyd:
Overall lgtm, before merging I would like to see it in action, can we define our own public upstream, say a fork of tests repo in our fork of https://github.com/openshift-eng/ocp-build-data/blob/openshift-4.15/images/openshift-enterprise-tests.yml , which is behind and then use it to see it in action?

PR #106, opened 2023-10-16
Update microshift sync error message

ashwindasr:
Don't error out the job, if ocp-artifacts can't connect to buildvm. Report back in slack


locriandev (commented):
/lgtm

PR #111, opened 2023-10-18
Revert "Revert "Exclude certain keys from config digest calculation""

locriandev:
Reverts openshift-eng/art-tools#105


thegreyd:
Noting that after https://github.com/openshift-eng/art-tools/pull/117 and say 4.15 pr which modifies base images https://github.com/openshift-eng/ocp-build-data/pull/3737 - if we merge this, base images will not be built until a next commit lands for them - but all other images which don't have a noop commit touching them, would get rebuilt

PR #110, opened 2023-10-17
[ART-7954] Ignore image config change on noop commit message

thegreyd:
https://issues.redhat.com/browse/ART-7954

Tested with https://github.com/openshift-eng/ocp-build-data/pull/3643

```
./doozer --assembly=stream --group 'openshift-4.15'@4.15-update-build-data-org --data-path https://github.com/locriandev/ocp-build-data --images openshift-base-nodejs config:scan-sources

2023-10-18 10:02:54,087 INFO openshift-base-nodejs config_digest sha256:5fa2beca671ca46d957547e08ad3cfba1b0b512c57994757c8d82896a4956963 is differing from sha256:734bb25abfe89ce7f032a202d62f21725288e8b69ed809f3f643d62a9b7a2878
2023-10-18 10:02:54,090 INFO $22: ['git', 'log', '-1', '--format=%s'] - [cwd=/tmp/oit-or_of314.tmp/ocp-build-data]: Executing:cmd_gather
2023-10-18 10:02:54,092 INFO Time elapsed (hh:mm:ss.ms) 0:00:00.002366 in exectools.py:167 from scan_sources.py:203:rc, commit_message, _ = exectools.cmd_gather('git log -1 --format=%s', strip=True) : $22: ['git', 'log', '-1', '--format=%s'] - [cwd=/tmp/oit-or_of314.tmp/ocp-build-data]: Executed:cmd_gather
2023-10-18 10:02:54,092 INFO Ignoring digest change since commit message indicates noop
```

PR #112, opened 2023-10-18
ART-7954 Clone and save config digest on noop change

thegreyd:
Follow up to https://github.com/openshift-eng/art-tools/pull/110


thegreyd:
digests are fetched from existing build commit and not from head, so this approach will not work.

PR #115, opened 2023-10-18
Make go compliance wrapper quiet by default

jupierce:


jupierce:
replaced with https://github.com/openshift-eng/ocp-build-data/pull/3730

cfergeau:
> Closing in preference of doing this only in CI builds. This will allow things to roll out more quickly and allow brew logs to include compliance data for ART post processing.

This is an issue in more than CI, for example in OpenShift Local we use these images in various (non OpenShift CI) places as we need build images close to what OpenShift uses, and with go 1.20+.
We'll have to add `GO_COMPLIANCE_INFO=0` every time we use these images.
When running `go build` without the wrapper, I only get messages when there's a problem (either a warning or an error), if everything went fine, `go build` is 100% silent.
When the `go` binary is replaced by a wrapper, it would be nice if it had the same behaviour, silent unless there's a problem. At the moment, the wrapper is quite verbose even when only in _INFO mode.

PR #126, opened 2023-10-20
Delete locks if associated build cannot be found

locriandev:
In case a build has been aborted (thus leaving an orphan lock in Redis), and then manually removed by a user, the clean up procedure will complain that the build associated to a lock cannot be found in Jenkins. Currently, it will display a warning and skip that lock.

This leads to locks that are never clean up. This PR proposes to delete those locks for which associated builds cannot be found. The only side effect I am seeing is that if we run `pyartcd` locally, we'll be creating locks with random IDs. These will be garbage collected by the cleanup procedure. It seems reasonable to me, as we shouldn't be creating those locks outside of a Jenkins context, and we might forgot to clean them up.

### DIFF
--- a/pyartcd/pyartcd/jenkins.py
+++ b/pyartcd/pyartcd/jenkins.py
@@ -408,3 +408,12 @@ def start_scan_osh(nvrs: list, version: str, email: Optional[str] = "", **kwargs
         params=params,
         **kwargs
     )
+
+
+def is_api_reachable() -> bool:
+    """
+    Returns True if Jenkins API is reachable, False otherwise
+    """
+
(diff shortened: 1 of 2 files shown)


joepvd:
At the risk of over-engineering things:

What if there is a hickup and the jenkins api is not reachable or so? Maybe the unlock action should take place when the job cannot be reached, but another end point can?

locriandev:
> At the risk of over-engineering things:
> 
> What if there is a hickup and the jenkins api is not reachable or so? Maybe the unlock action should take place when the job cannot be reached, but another end point can?

I added a health check; locks will be deleted only if API are responding on their main endpoint

PR #131, opened 2023-10-24
Do not crash if config_digest is not found, just skip the check

locriandev:
https://redhat-internal.slack.com/archives/GDBRP5YJH/p1698137843148919?thread_ts=1698133398.861759&cid=GDBRP5YJH


vfreex (commented):
/lgtm
/approve

PR #133, opened 2023-10-24
[find-builds] Log advisories that filtered builds are attached to

thegreyd:
It is helpful to know which advisories, the filtered builds are attached to

`./elliott -g openshift-4.14 --assembly rc.7 find-builds -k image --non-payload`


locriandev (commented):
/lgtm

PR #134, opened 2023-10-24
[build-microshift] Skip advisory prep if there is no advisory

thegreyd:
So that we don't run these steps for ECs


joepvd (approved):
/lgtm

PR #141, opened 2023-10-26
[ART-7926] Create OCPBUGS tickets to track vulnerabilities

ashwindasr:
Ticket: https://issues.redhat.com/browse/ART-7926
Policy document [here](https://docs.google.com/document/d/1GsT3tvw5qf3sOSAkTSVWnMIO6vzIuirhsVhhsbK_qf0/edit#heading=h.uilnzpc6uuc7)

Proposed improvement requests to osh-cli
 -  https://issues.redhat.com/browse/OSH-376
 - https://issues.redhat.com/browse/OSH-390
 
This will significantly reduce the time to search for scan results. But moving forward with a less efficient approach due to urgency from ProdSec.

Changes in this PR:
- New feature `--create-jira-tickets`:
   Finds the last successfully completed scan for a brew package for a particular version and creates a an OCPBUGS ticket if found. ProdSec only wants us to look at `scan-results-imp` file ([example](https://cov01.lab.eng.brq2.redhat.com/osh/task/361037/log/sriov-network-operator-container-v4.14.0-202311020548.p0.g514df18.assembly.stream/scan-results-imp.js)) which has all the important vulnerabilities that need to be looked at by developers. A test ticket that got created by this feature: https://issues.redhat.com/browse/ART-8177
- Skipping `rhcos` builds while collecting NVRs to scan
- Refactor code for new `generate_commands` function to log command before `fire_and_forget`
- Raise tickets only if diff is satisfied (can override with `--skip-diff-check` ([code](https://github.com/openshift-eng/art-tools/commit/b7efbdde56badb6ce9bd82aab0ba3ace9daa0e35#diff-990549c3908d335c0917a3ed7d6fda9b6069fd0e0551ede2e5838cd87dc6b92cR124))
- Enable by image component level ([code](https://github.com/openshift-eng/art-tools/commit/b7efbdde56badb6ce9bd82aab0ba3ace9daa0e35#diff-990549c3908d335c0917a3ed7d6fda9b6069fd0e0551ede2e5838cd87dc6b92cR160)) and OCP version level ([code](https://github.com/openshift-eng/art-tools/commit/b7efbdde56badb6ce9bd82aab0ba3ace9daa0e35#diff-990549c3908d335c0917a3ed7d6fda9b6069fd0e0551ede2e5838cd87dc6b92cR436)) (disabled by default)
- Set ticket security to `Red Hat Employee` by default ([code](https://github.com/openshift-eng/art-tools/commit/b7efbdde56badb6ce9bd82aab0ba3ace9daa0e35#diff-990549c3908d335c0917a3ed7d6fda9b6069fd0e0551ede2e5838cd87dc6b92cR195))
- Consider any state > `ASSIGNED` in OCPBUGS as closed ([code](https://github.com/openshift-eng/art-tools/pull/141/commits/d4445f4b2ac790fb4bc9c3b7c4d320c9968c9488))
  

_Requesting `/lgtm` and `/approve`_

### DIFF
--- a/doozer/doozerlib/cli/scan_osh.py
+++ b/doozer/doozerlib/cli/scan_osh.py
@@ -1,16 +1,41 @@
 import click
 import koji
 import asyncio
+
+import requests
+import yaml
 from doozerlib.cli import cli, click_coroutine, pass_runtime
 from doozerlib.runtime import Runtime
-from doozerlib.exectools import fire_and_forget, cmd_gather_async, limit_concurrency
+from doozerlib.exectools import fire_and_forget, cmd_gather_async, limit_concurrency, cmd_gather
 from doozerlib.util import cprint
+from doozerlib.image import ImageMetadata
 from typing import Optional
+from jira import JIRA, Issue
+from tenacity import retry, stop_after_attempt, wait_fixed
+from doozerlib.rpm_utils import parse_nvr
+
+SCAN_RESULTS_URL_TEMPLATE = "https://cov01.lab.eng.brq2.redhat.com/osh/task/{task_id}/log/{nvr}/scan-results-imp.js" \
+                            "?format=raw"
+
+MESSAGE = """
+One or more software security scan issues have been detected for the package *{package_name}*.
+
+*Build*: [{nvr}|https://brewweb.engineering.redhat.com/brew/buildinfo?buildID={brew_build_id}]
+*Scan result*: https://cov01.lab.eng.brq2.redhat.com/osh/task/{scan_id}
+*View issues*: https://cov01.lab.eng.brq2.redhat.com/osh/task/{scan_id}/log/{nvr}/scan-results-imp.js (Check 'defects')
+
+Once these issues are addressed, the Bug can be closed. But a new one will be opened if new issues are found in the next build.
+
+For information on triaging these issues and dispositioning this ticket, please review the [policy document for SAST Scanning Jira Tickets|https://docs.google.com/document/d/1GsT3tvw5qf3sOSAkTSVWnMIO6vzIuirhsVhhsbK_qf0].
+
+For other questions please reach out to @release-artists in #forum-ocp-art.
+"""
 
 
 class ScanOshCli:
     def __init__(self, runtime: Runtime, last_brew_event: int, dry_run: bool, nvrs: Optional[list],
-                 check_triggered: Optional[bool], all_builds: Optional[bool]):
+                 check_triggered: Optional[bool], all_builds: Optional[bool], create_jira_tickets: Optional[bool],
+                 skip_diff_check: Optional[bool]):
         self.runtime = runtime
         self.last_brew_event = last_brew_event
         self.dry_run = dry_run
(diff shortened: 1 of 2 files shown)


doozer/doozerlib/cli/scan_osh.py
```diff
@@ -1,29 +1,186 @@
...
+            open_issues = search_issues(query)
+            self.runtime.logger.info(f"Issues found with query '{query}': {open_issues}")
+
+            if len(open_issues) == 0:
+                # Create a new JIRA ticket
+                fields = {
+                    "project": {"key": f"{self.jira_project}"},
+                    "issuetype": {"name": "Bug"},
+                    "versions": [{"name": f"{self.version}.z"}],  # Affects Version/s
+                    "components": [{"name": "Release"}],
```
jupierce, line 103: package to component mapping information is available here: https://github.com/openshift-eng/ocp-build-data/blob/main/product.yml . You can see similar Jira issue creation / reconciliation here: https://github.com/openshift-eng/art-tools/blob/a3cbdbfcf1654d2c78a07992f6baea43b4230e97/doozer/doozerlib/cli/images_streams.py#L683 .

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -1,29 +1,186 @@
...
+        """
+        for brew_package_name in packages:
+            data = packages[brew_package_name]
+
+            # Check if ticket exits
+            summary = f"{self.version} SAST scan issues for {brew_package_name}"
+            query = f"project={self.jira_project} AND ( summary ~ '{summary}' ) AND " \
+                    "statusCategory in ('To Do', 'In Progress')"
+            description = "Vulnerabilities have been detected for this package. Check here for the last scan results: " \
+                          f"{SCAN_RESULTS_URL_TEMPLATE.format(task_id=data['latest_coverity_scan'], nvr=data['nvr'])}"
```
jupierce, line 88: Is this an HTML view we can link to as well?

doozer/doozerlib/cli/scan_osh.py
ashwindasr: Thanks for pointer! I've added that in https://github.com/openshift-eng/art-tools/pull/141/commits/117f1efb6b41a9e4bcd3027f152bda54d83def23

doozer/doozerlib/cli/scan_osh.py
ashwindasr: Yes, if we remove the `raw` suffix, we get a web view. Added in https://github.com/openshift-eng/art-tools/pull/141/commits/117f1efb6b41a9e4bcd3027f152bda54d83def23

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -1,29 +1,240 @@
...
+    def create_update_ocpbugs_ticket(self, packages: dict):
+        """
+        Check if an OCPBUGS ticket already exists for a particular package. If it doesn't create one and update
+        the contents to point to the latest scan results
+        """
+        for brew_package_name in packages:
+            data = packages[brew_package_name]
+            distgit_name = self.get_distgit_name_from_brew_nvr(data["nvr"])
+
+            image_meta: ImageMetadata = self.runtime.image_map[distgit_name]
```
jupierce, line 119: Suggesting we allow jira ticket creation to be enabled/disabled at the metadata level. We can default it to disabled at first and onboard a single image/rpm and to how things work. e.g. in the metadata, require
`scanning.jira_integration.enabled: true` to be set explicitly before opening issues. 
We will also probably want something similar in group.yml to enable Jira integration slowly across z-streams so as to not overwhelm folks.

ashwindasr:
New updates in this commit https://github.com/openshift-eng/art-tools/pull/141/commits/b7efbdde56badb6ce9bd82aab0ba3ace9daa0e35

- Raise tickets only if diff is satisfied (can override with `--skip-diff-check`
- Enable by image component level and OCP version level (disabled by default)
- Set ticket security to `Red Hat Employee` by default

doozer/doozerlib/cli/scan_osh.py
ashwindasr: Added in https://github.com/openshift-eng/art-tools/pull/141/commits/b7efbdde56badb6ce9bd82aab0ba3ace9daa0e35

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -1,29 +1,289 @@
...
+
+    def get_distgit_name_from_brew_nvr(self, nvr: str) -> str:
+        """
+            Returns the distgit name from the brew nvr
+
+            :param nvr: Full NVR eg: openshift-enterprise-pod-container-v4.14.0-202310031045.p0.g9ef6de6.assembly.stream
+            """
+        brew_package_name = self.koji_session.getBuild(buildInfo=nvr, strict=True)["package_name"]
+
+        _, output, _ = cmd_gather(f"doozer --disable-gssapi -g {self.runtime.group} "
```
thegreyd, line 110: if this function gets called for each nvr, it's better if we cache this command output

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -158,6 +433,16 @@ async def run(self):
 
         self.trigger_scans(nvrs_for_scans)
 
+        if not self.runtime.group_config.scanning.jira_integration.enabled:
+            self.runtime.logger.info(f"Skipping OCPBUGS creation workflow since not enabled in group.yaml for "
```
thegreyd, line 437:
```suggestion
            self.runtime.logger.info(f"Skipping OCPBUGS creation workflow since not enabled in group.yml for "
```

doozer/doozerlib/cli/scan_osh.py
ashwindasr: Do we already have a cache header anywhere or shall I create one?

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -158,6 +433,16 @@ async def run(self):
 
         self.trigger_scans(nvrs_for_scans)
 
+        if not self.runtime.group_config.scanning.jira_integration.enabled:
```
thegreyd, line 436: Maybe an explicit check is better here `if enabled in [True, "true", "yes"]`

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -68,16 +328,16 @@ def get_tagged_all(self, tag):
         else:
             return []
 
-    def trigger_scans(self, nvrs: list):
+    def generate_commands(self, nvrs) -> (list, None):
```
thegreyd, line 331:
```suggestion
    def generate_commands(self, nvrs) -> list:
```

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -90,8 +350,14 @@ def trigger_scans(self, nvrs: list):
                     self.runtime.logger.error("Invalid RHEL version")
                     return
```
thegreyd, line 351: let's raise the error here?

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -1,29 +1,289 @@
...
+    async def ocp_bugs_workflow_run(self, nvrs: list):
+        # List of brew packages to check
+        brew_package_names = []
+
+        for nvr in nvrs:
+            brew_info = self.koji_session.getBuild(nvr)
+
+            # Get ose-network-tools-container-v4.15.0 from ose-network-tools-container-v4.15.0-202310170244.p0.gdf85e45.assembly.stream
+            brew_package_names.append({
+                "package_name_with_version": brew_info["nvr"].split(f"v{self.version}")[0] + f"v{self.version}.0",
```
thegreyd, line 267: Another alternative is using `parse_nvr` from rpm_utils.py
```python
nvr = parse_nvr(brew_info["nvr"])
pkg_name_w_version = f'nvr["name"]-v{self.version}.0'
```

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -1,29 +1,289 @@
...
+                "nvr": brew_info["nvr"],
+                "brew_build_id": brew_info["id"],
+                "package_name": brew_info["package_name"]
+            })
+
+        # Get latest scan results
+        brew_coverity_mapping = {}
+
+        for package in brew_package_names:
+            package["latest_coverity_scan"], package["previous_scan_id"], package["previous_scan_nvr"] = \
```
thegreyd, line 277: I think we can make this simpler by passing package to get_scan_info and then it fills the fields in

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -1,29 +1,289 @@
...
+    def is_latest_scan_different(latest_scan_id, latest_scan_nvr, previous_scan_id, previous_scan_nvr):
+        latest_scan_response = requests.get(
+            SCAN_RESULTS_URL_TEMPLATE.format(task_id=latest_scan_id, nvr=latest_scan_nvr))
+        previous_scan_response = requests.get(
+            SCAN_RESULTS_URL_TEMPLATE.format(task_id=previous_scan_id, nvr=previous_scan_nvr))
+
+        latest_scan_defects = latest_scan_response.json()["defects"]
+        previous_scan_defects = previous_scan_response.json()["defects"]
+
+        for defect in latest_scan_defects:
```
thegreyd, line 133: Is defect a string here?

doozer/doozerlib/cli/scan_osh.py
thegreyd: I'm not sure what you mean by cache header .. I was thinking of a var that stores this mapping the first time this method is called and then on subsequent calls we return from it

ashwindasr: I meant something like [this](https://github.com/openshift-eng/art-bot/blob/9dd835fa6b1c76a1f9f2f7d4d0bfdccdefd0b3f4/artbotlib/util.py#L77)

doozer/doozerlib/cli/scan_osh.py
ashwindasr: Would `Optional[list]` work?

doozer/doozerlib/cli/scan_osh.py
ashwindasr: Nope, its an `dict` :slightly_smiling_face: 

This is the [sample](https://cov01.lab.eng.brq2.redhat.com/osh/task/361037/log/sriov-network-operator-container-v4.14.0-202311020548.p0.g514df18.assembly.stream/scan-results-imp.js). If you check there, you'll see that its a dict. The aim is to raise a ticket only if the diff changes.

doozer/doozerlib/cli/scan_osh.py
ashwindasr: fixed in https://github.com/openshift-eng/art-tools/pull/141/commits/e4031289cbb0f229b3524b08cc04909ef204bc17

doozer/doozerlib/cli/scan_osh.py
ashwindasr: fixed in https://github.com/openshift-eng/art-tools/pull/141/commits/e4031289cbb0f229b3524b08cc04909ef204bc17

doozer/doozerlib/cli/scan_osh.py
ashwindasr: fixed in https://github.com/openshift-eng/art-tools/pull/141/commits/e4031289cbb0f229b3524b08cc04909ef204bc17

doozer/doozerlib/cli/scan_osh.py
ashwindasr: thank you! fixed in https://github.com/openshift-eng/art-tools/pull/141/commits/e4031289cbb0f229b3524b08cc04909ef204bc17

doozer/doozerlib/cli/scan_osh.py
ashwindasr: Hmm, I was hoping to keep it like this for now. Want to refine the entire workflow anyways. Shall I keep this as a refinement?

doozer/doozerlib/cli/scan_osh.py
ashwindasr: added in https://github.com/openshift-eng/art-tools/pull/141/commits/e4031289cbb0f229b3524b08cc04909ef204bc17

doozer/doozerlib/cli/scan_osh.py
ashwindasr: But what you mentioned is the best way I feel, implemented in https://github.com/openshift-eng/art-tools/pull/141/commits/ce0bb7167b9640df3b86cbe1e497a419585d656a

doozer/doozerlib/cli/scan_osh.py
ashwindasr: @jupierce I hope this way of checking diff is acceptable. Please let me know if I missed something?

doozer/doozerlib/cli/scan_osh.py
thegreyd: Instead of calling it in the constructor always, I would suggest we move it to the func itself, so
```python
# in constructor
self.brew_distgit_mapping = None

def get_brew_distgit_mapping(self):
    if self.brew_distgit_mapping:
        return self.brew_distgit_mapping[brew_package_name]
    # else call doozer and store
```

doozer/doozerlib/cli/scan_osh.py
thegreyd: I don't see it, did we miss it :)

doozer/doozerlib/cli/scan_osh.py
thegreyd: Don't see it..

doozer/doozerlib/cli/scan_osh.py
thegreyd: It's more of a comment about code style.. I don't know if creating a jira story is worth it

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -1,29 +1,302 @@
...
+                "security": {"id": "11697"},  # Restrict to Red Hat Employee
+                "summary": summary,
+                "description": description
+            }
+
+            if len(open_issues) == 0:
+                # No tickets exist in ('New', 'Assigned') so we could consider opening a new one
+                # Do not reraise a ticket if the diff has not changed, but we would still need to update the ticket
+                # with the new build and scan
+                if not self.skip_diff_check and not self.is_latest_scan_different(
```
thegreyd, line 201: Making sure I understand this - we will not open a jira ticket until there is a new defect since this scan starts running? Why not open ticket for existing ones? Are we factoring in that the dev team may have closed the previous ticket?

doozer/doozerlib/cli/scan_osh.py
thegreyd: Thanks, lgtm

doozer/doozerlib/cli/scan_osh.py
ashwindasr: I ran pytest locally and it wiped out my doozer directory. Dont know what happened. It was just before I was about to push to GitHub. So had to make the changes from memory and I missed it. Sorry about that

doozer/doozerlib/cli/scan_osh.py
ashwindasr: Fixed and pushed https://github.com/openshift-eng/art-tools/pull/141/files#diff-990549c3908d335c0917a3ed7d6fda9b6069fd0e0551ede2e5838cd87dc6b92cR455

doozer/doozerlib/cli/scan_osh.py
ashwindasr: added https://github.com/openshift-eng/art-tools/pull/141/commits/6f3bc633c8c3bcb85742eb30ef2e4a68c0a5d861

doozer/doozerlib/cli/scan_osh.py
ashwindasr: Yeah I'm thinking about keep it like this for now. There could be some changes coming in, like openscanhub is being migrated to DefectDojo soon, so will have to change stuff anyways.

doozer/doozerlib/cli/scan_osh.py
ashwindasr: So the policy that Justin suggested is that if the developers close the ticket, we will consider that as acknowledgement that they fixed the issue. So we need to raise only if we find new issues.

ashwindasr: I've added the [policy document](https://docs.google.com/document/d/1GsT3tvw5qf3sOSAkTSVWnMIO6vzIuirhsVhhsbK_qf0/edit#heading=h.uilnzpc6uuc7) to the PR description.

ashwindasr: This is currently under discussion with other Staff Engineers and managers, since the problem appears to be that we need to make sure that we don't spam devs with tickets.

doozer/doozerlib/cli/scan_osh.py
thegreyd: Makes sense

doozer/doozerlib/cli/scan_osh.py
thegreyd: Makes sense, I'm thinking of the case of opening tickets for the first time - if an image that does not receive frequent updates, will not have a ticket created until some change which includes a new defect gets pushed, which could be months in the worst case? Maybe we can explicitly search for closed issues to get around this case?

thegreyd: Or we can run it with the `--skip-diff-check` flag for the first time for all images?

ashwindasr: Ahh the first scan case, another thing I wrote the code for but got wiped.

ashwindasr: So we just need to say that its different right? Like this https://github.com/openshift-eng/art-tools/pull/141/commits/3442602132283a1ad22a78455e4217182553ba6d

ashwindasr: Ahh wait, I see what you're saying. So the scenario is that the first scan had vulnerabilities, we raised the ticket, devs closed it. But they didn't fix it. The second commit to that repo is after 2 months. No new vulnerabilties are found, we will not reraise the ticket as per our current policy.

ashwindasr: Well I guess right now, the policy is lenient so that we can ease in the devs to this new process. Once everyone gets adjusted to the new normal we can make it more strict?

> Maybe we can explicitly search for closed issues to get around this case?
 There is the `--skip-diff-check` flag as well, so we could use that for this case?

ashwindasr:
/wip fix edge case when we are enabling this feature for the first time

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -173,19 +177,29 @@ def create_update_ocpbugs_ticket(self, packages: dict):
...
-            open_issues = search_issues(query)
+            # Check if this is the first time that we are raising the ticket for this component
+            query = f"project={self.jira_project} AND ( summary ~ '{summary}' ) AND " \
+                    "statusCategory not in ('New', 'Assigned')"
+            closed_issues = self.search_issues(query)
+
+            if not closed_issues:
+                # There are no previous tickets for this component, raise the ticket this time if
+                # vulnerabilities are found
+                self.skip_diff_check = True
```
ashwindasr, line 194: Thanks for pointing this out @thegreyd , hope this would resolve it?

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -173,19 +177,29 @@ def create_update_ocpbugs_ticket(self, packages: dict):
...
+
+            # Find if there is already a "OPEN" ticket for this component.
+            # A ticket is said to be "OPEN" if statusCategory <= 'Assigned'.
+            # If we detect a net-new security issue, we should open a new ticket instead of updating the one
+            # that is in the process of being fixed.
```
ashwindasr, line 199: @thegreyd This is what I wanted to mention today. So if an engineer has started to work on the ticket, any new vulnerabilities should be raised as a new ticket. Right now we may not be listing out the new vulnerabilities found, but we could add that as a refinement.

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -1,29 +1,336 @@
...
+One or more vulnerabilities have been detected for the package *{package_name}*
+
+*Build*: [{nvr}|https://brewweb.engineering.redhat.com/brew/buildinfo?buildID={brew_build_id}]
+*Scan result*: https://cov01.lab.eng.brq2.redhat.com/osh/task/{scan_id}
+*View vulnerabilities*: https://cov01.lab.eng.brq2.redhat.com/osh/task/{scan_id}/log/{nvr}/scan-results-imp.js (Check 'defects')
+
+Once these issues are addressed, the Bug can be closed. But a new one will be opened if new vulnerabilities are found in the next build.
+
++POCs+
+*ProdSec*: If you have any questions regarding the vulnerability listed, email secalert@redhat.com
```
jupierce, line 30: I think we can safely replace this whole section now with a link to the policy document. Something like:

For information on triaging these issues and dispositioning this ticket, please review the [policy document for SAST Scanning Jira Tickets](https://docs.google.com/document/d/1GsT3tvw5qf3sOSAkTSVWnMIO6vzIuirhsVhhsbK_qf0/edit).

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -1,29 +1,336 @@
...
+MESSAGE = """
+One or more vulnerabilities have been detected for the package *{package_name}*
+
+*Build*: [{nvr}|https://brewweb.engineering.redhat.com/brew/buildinfo?buildID={brew_build_id}]
+*Scan result*: https://cov01.lab.eng.brq2.redhat.com/osh/task/{scan_id}
+*View vulnerabilities*: https://cov01.lab.eng.brq2.redhat.com/osh/task/{scan_id}/log/{nvr}/scan-results-imp.js (Check 'defects')
+
+Once these issues are addressed, the Bug can be closed. But a new one will be opened if new vulnerabilities are found in the next build.
+
++POCs+
```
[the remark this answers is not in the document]
ashwindasr: Noted and Learned! Added it to my "Quotes by Justin" list!

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -1,29 +1,336 @@
...
+
+            previous_ticket_id = None
+            if not closed_issues and not open_issues:
+                # There are no tickets for this component, raise a ticket
+                self.skip_diff_check = True
+            else:
+                self.runtime.logger.info(f"Closed Issues: {closed_issues}")
+                previous_ticket_id = closed_issues.pop()  # Returned in LIFO (last-in, first-out) order.
+
+            description = "{*}{color:#FF0000}WARNING{color}{*}\n" + MESSAGE.format(package_name=brew_package_name,
```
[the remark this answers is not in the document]
ashwindasr: So OpenScanHub is working on adding new results to the `imp` list. Regardless, we need Devs to at least look at the results right. They may need to triage and figure out if its a vulnerability or not right. So keeping this would at least motivate them to look at them right. :thinking:

ashwindasr:
/hold

giving time for people to give feedback on the policy doc

ashwindasr:
/unhold

Need to create aos-cd-jobs PR to activate the feature

joepvd (approved):
/lgtm

PR #142, opened 2023-10-27
ART-8041 Use git ls-remote instead of GitHub API

locriandev:
Changes in this PR:
-  `_do_shas_match()` updated to use `git ls-remote` to compare commit SHAs on private/public upstrams without calling GitHub API
- if `ocp4-scan` is built with `DRY_RUN` set to True, pass the `--dry-run` option to `doozer config:scan-sources`; this will not perform the rebase, only print what it would have done
- re-enable the currently disabled rebase procedure, by adding the `--rebase-priv` flag to doozer

Test run: https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Focp4_scan/553872/console

### DIFF
--- a/doozer/doozerlib/cli/scan_sources.py
+++ b/doozer/doozerlib/cli/scan_sources.py
@@ -1,14 +1,13 @@
 import asyncio
-import os
 from datetime import datetime, timezone
 
 import click
 import yaml
 from typing import List, Tuple, Optional
 
-from github import Github, GithubException
+from github import GithubException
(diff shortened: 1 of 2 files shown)


joepvd (approved):
/hold Not sure if special care needs to be take for deploy.

locriandev:
No care is required, can merge right away (with proper supervision which I'll provide)

PR #143, opened 2023-10-27
[build-microshift] ART-8097 Check if pinned builds match current assembly

thegreyd:


pyartcd/pyartcd/util.py
```diff
@@ -157,8 +157,10 @@ def get_release_name_for_assembly(group_name: str, releases_config: Dict, assemb
     return doozerutil.get_release_name_for_assembly(group_name, model.Model(releases_config), assembly_name)
 
 
-def get_rpm_if_pinned(releases_config: Dict, assembly_name: str, rpm_name: str) -> dict:
-    pinned_rpms = assembly._assembly_config_struct(model.Model(releases_config), assembly_name, 'members', {'rpms': []})['rpms']
```
thegreyd, line 161: Curiously this gets the oldest assemblies attributes, so for 4.14.1 it got rc.7 rpm .. 
That's why I removed it and instead directly queried the current assembly

thiagoalessio (approved):
/lgtm

PR #144, opened 2023-10-27
Revert "[ART-7056] deprecate placeholder bug"

thegreyd:
Reverts openshift-eng/art-tools#25

now that advisories have started requiring bugs again


Ximinhan:
not true, we should consider if the errata rule changed

PR #146, opened 2023-10-30
ART-7990: Improve wording about blocker bug messages

mbiarnes:
related JIRA: [ART-7990](https://issues.redhat.com/browse/ART-7990)

### DIFF
--- a/elliott/elliottlib/cli/find_bugs_sweep_cli.py
+++ b/elliott/elliottlib/cli/find_bugs_sweep_cli.py
@@ -387,9 +387,13 @@ def extras_bugs(bugs: type_bug_set) -> type_bug_set:
 
 
 def print_report(bugs: type_bug_list, output: str = 'text') -> None:
+    approved_url = 'https://source.redhat.com/groups/public/openshift/openshift_wiki/openshift_bugzilla_process'
     if output == 'slack':
         for bug in bugs:
-            click.echo("<{}|{}> - {:<25s} ".format(bug.weburl, bug.id, bug.component))
+            if bug.release_blocker:
+                click.echo("<{}|_Release blocker: Approved_> bug for <{}|{}> - {:<25s} ".format(approved_url, bug.weburl, bug.id, bug.component))
+            else:
+                click.echo("<{}|{}> - {:<25s} ".format(bug.weburl, bug.id, bug.component))
 
     elif output == 'json':
         print(json.dumps(
--- a/pyartcd/pyartcd/pipelines/check_bugs.py
+++ b/pyartcd/pyartcd/pipelines/check_bugs.py
@@ -178,15 +178,14 @@ class CheckBugsPipeline:
                 report[k].extend(v)
 
         # Format output message
-        message = ':red-siren: *There are some issues to look into:*'
         for k in report.keys():
-            message += f'\n:warning:*{k}*'
+            message = f':red-siren:  `Bug(s) requiring attention for:` \n:warning: *{k}*'
             for i in report[k]:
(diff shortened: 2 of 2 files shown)


pyartcd/pyartcd/pipelines/check_bugs.py
```diff
@@ -184,10 +184,11 @@ async def _slack_report(self):
...
                 message += f'\n{i}'
 
-        self.logger.info('Sending notification to Slack')
-        self.logger.debug(message)
-        await self.slack_client.say(message)
 
+            self.logger.info('Sending notification to Slack')
+            self.logger.debug(message)
+            await self.slack_client.say(message)
+            message = ''
```
ashwindasr, line 191: Curious as to why manually setting it to `null` at the end of a function?

ashwindasr (dismissed):

pyartcd/pyartcd/pipelines/check_bugs.py
mbiarnes: For sending slack messages individually for each version and not as a collection over all versions

pyartcd/pyartcd/pipelines/check_bugs.py
```diff
@@ -184,9 +184,10 @@ async def _slack_report(self):
             for i in report[k]:
                 message += f'\n{i}'
 
-        self.logger.info('Sending notification to Slack')
-        self.logger.debug(message)
-        await self.slack_client.say(message)
+            self.logger.info('Sending notification to Slack')
```
thegreyd, line 187: This will send one slack message per version, I don't think we want that

elliott/elliottlib/cli/find_bugs_sweep_cli.py
```diff
@@ -387,9 +387,13 @@ def extras_bugs(bugs: type_bug_set) -> type_bug_set:
 
 
 def print_report(bugs: type_bug_list, output: str = 'text') -> None:
+    approved_url = 'https://source.redhat.com/groups/public/openshift/openshift_wiki/openshift_bugzilla_process'
     if output == 'slack':
         for bug in bugs:
-            click.echo("<{}|{}> - {:<25s} ".format(bug.weburl, bug.id, bug.component))
+            if bug.release_blocker:
+                click.echo("<{}|_Release blocker: Approved_> bug for <{}|{}> - {:<25s} ".format(approved_url, bug.weburl, bug.id, bug.component))
```
thegreyd, line 394: I don't know if this is the best place to implement this formatting, I would rather keep this in check_bugs.py

thegreyd: Hmm but I see that we are relying on already this report to come pre-formatted .. what we can do is use `--output json` instead of `--output slack` for the elliott commands that we call in check_bugs.py and then proceed to do formatting as we want there.

joepvd: For better or worse, we are currently creating the markup in elliott, and not in pyartcd. Don't think there is a reason to change.

pyartcd/pyartcd/pipelines/check_bugs.py
joepvd: I think it is great to have a dedicated thread for each issue. I have been confusing myself and others by having conversations about different topics in the same thread, and I think a separate thread for each version gets us closer to a one-topic-per-thread situation. 

Multiple issues per version is rather rare, so this lgtm :)

joepvd (approved):
/lgtm

PR #147, opened 2023-10-30
Treat go-install just as go-build

jupierce:
In one case, a component was using 'go install ...' to perform the build instead of directly running 'go build'. The FIPS mitigations were not fully applied because the script was only looking for build invocations.


vfreex (commented):
/lgtm
/approve

PR #148, opened 2023-11-02
ART-7731 use actual advisroy type for release

Ximinhan:
**boilerplate_name** in the create advisory function need to be one of the type ['rpm', 'microshift', 'metadata', 'image', 'extras'] with match the key in errata template
**impetus** is used for advisory metadata info need to be one of [standard, cve, ga, test]
**kind** is used for advisory class that need to be one of [rpm, image]

### DIFF
--- a/elliott/elliottlib/cli/create_cli.py
+++ b/elliott/elliottlib/cli/create_cli.py
@@ -24,6 +24,9 @@ LOGGER = elliottlib.logutil.getLogger(__name__)
               type=click.Choice(elliottlib.constants.errata_valid_impetus),
               default='standard',
               help="Impetus for the advisory creation. 'standard' by default")
+@click.option("--art-advisory-key",
+              type=click.Choice(['rpm', 'microshift', 'metadata', 'image', 'extras']),
+              help="Class of the advisory. Should be one of [rpm, microshift, metadata, image, extras].")
 @click.option("--date", required=True,
               callback=validate_release_date,
               help="Release date for the advisory. Format: YYYY-Mon-DD.")
@@ -49,7 +52,7 @@ LOGGER = elliottlib.logutil.getLogger(__name__)
               help="Bug IDs for attaching to the advisory on creation. Required for creating a security advisory.")
 @click.pass_obj
 @click.pass_context
-def create_cli(ctx, runtime, errata_type, kind, impetus, date, assigned_to, manager, package_owner, with_liveid, yes, bugs):
+def create_cli(ctx, runtime, errata_type, kind, impetus, art_advisory_key, date, assigned_to, manager, package_owner, with_liveid, yes, bugs):
     """Create a new advisory. The kind of advisory must be specified with
 '--kind'. Valid choices are 'rpm' and 'image'.
 
@@ -112,7 +115,7 @@ advisory.
             et_data,
             errata_type=errata_type,
             kind=kind,
-            boilerplate_name=(impetus if impetus != "standard" else kind),
+            boilerplate_name=(art_advisory_key if art_advisory_key else kind),
             release_date=release_date.strftime(YMD),
             assigned_to=assigned_to,
             manager=manager,
--- a/pyartcd/pyartcd/pipelines/prepare_release.py
+++ b/pyartcd/pyartcd/pipelines/prepare_release.py
@@ -149,9 +149,9 @@ class PrepareReleasePipeline:
             for ad in advisories:
                 if advisories[ad] < 0:
                     if ad == "microshift":
-                        advisories[ad] = self.create_advisory("RHEA" if is_ga else "RHBA", "rpm", ad)
+                        advisories[ad] = self.create_advisory("RHEA" if is_ga else "RHBA", "rpm", ad, ad)
                     else:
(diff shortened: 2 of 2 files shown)


elliott/elliottlib/cli/create_cli.py
```diff
@@ -112,7 +115,7 @@ def create_cli(ctx, runtime, errata_type, kind, impetus, date, assigned_to, mana
             et_data,
             errata_type=errata_type,
             kind=kind,
-            boilerplate_name=(impetus if impetus != "standard" else kind),
```
Ximinhan, line 115: we pass standard for non microshift here is incorrect, it will make advisory creation check failed https://github.com/openshift-eng/art-tools/blob/main/elliott/elliottlib/errata.py#L238 and create from default value not the boilerplate config

thegreyd: I looked at impetus values, why not remove all values that are not in boilerplate?
```
boilerplate_name = (impetus if impetus not in ['standard', 'cve', 'ga', 'test'] else kind)
```

thegreyd: And then whichever place we are calling create for microshift, we make sure to pass it as impetus?

Ximinhan: you can see the invoking from preparing releaae job. the issue here is impetus is different usage here. we did't pass sort like extras or metadata type here.

thegreyd: Okay I traced how we use impetus, impetus is basically the boilerplate name. I think we should refactor and fix our use of impetus instead of adding "sort". Let's start by analyzing `elliottlib.constants.errata_valid_impetus` unused values, why do we have ga/test defined if we don't have boilerplate value for them? I would prefer removing "standard" as well, if impetus is not defined we already fall back on kind, so we don't really need an explicit value for it like "standard". Once we have removed ga and standard, we can simply pass the advisory type as impetus.

Ximinhan: It is misleading with those types, I would like to get this fix pr in so docs can have correct content and refactor the implementation later in a separate card

Ximinhan: It is misleading with those types, I would like to get this fix pr in so docs can have the correct content and refactor the implementation later in a separate card

elliott/elliottlib/cli/create_cli.py
```diff
@@ -24,6 +24,9 @@
               type=click.Choice(elliottlib.constants.errata_valid_impetus),
               default='standard',
               help="Impetus for the advisory creation. 'standard' by default")
+@click.option("--sort",
```
joepvd, line 27: `--sort` lets me think more of ordering a list. 
Maybe `--advisory-key` is better? In any case, think all values in https://github.com/openshift-eng/ocp-build-data-validator/blob/master/validator/json_schemas/assembly_group_config.schema.json#L30 should be possible.

joepvd (commented):
I am having similar reservations as thegreyd. This PR does not make it obvious to me what all the type descriptions of different advisories actually are, and introduces a new concept. 

If we introduce a new category in our tooling to classify advisories, I would like to fully understand why this is necessary. 

I disagree with "it is a bug fix, let's merge it," because the introduction of a new concept should be more thought out, imho.

elliott/elliottlib/cli/create_cli.py
Ximinhan: if you are suggesting the naming, I can change it to `--advisory-property`, I don't remember I saw a type with `silentops` or `advance` before and they are not apply the errata template

Ximinhan:
> I am having similar reservations as thegreyd. This PR does not make it obvious to me what all the type descriptions of different advisories actually are, and introduces a new concept.
> 
> If we introduce a new category in our tooling to classify advisories, I would like to fully understand why this is necessary.
> 
> I disagree with "it is a bug fix, let's merge it," because the introduction of a new concept should be more thought out, imho.

We previously mistakenly used the three concepts introduced in the description, which resulted in us not creating the document correctly. This PR distinguishes the three concepts in the description when used by adding new parameters.

elliott/elliottlib/cli/create_cli.py
Ximinhan: as it described in https://github.com/openshift-eng/art-tools/blob/main/elliott/elliottlib/cli/add_metadata_cli.py#L33 the impetus is used in an advisory's comment like https://errata.devel.redhat.com/advisory/123315#c1, not sure if it's still useful for someone, maybe we can try remove the step to see if someone complain
and for the kind it said https://github.com/openshift-eng/art-tools/blob/main/elliott/elliottlib/errata.py#L210 I think it's used for distinguished the key in boilerplates so maybe it can also removed

Ximinhan: @thegreyd I created [ART-8191](https://issues.redhat.com/browse/ART-8191)

joepvd:
/lgtm
/approve
Yet:
/hold
...to give possibility to improve the name `--property` before it merges :)

PR #150, opened 2023-11-03
Fix liveID verification

thegreyd:
Move liveID verification out of `invalid_errata_status` permit


thiagoalessio (approved):
/lgtm

let's run it and see if it works :smile:

PR #156, opened 2023-11-07
ART-8187 check blocking advisory when drop advisory

Ximinhan:
Remove blocking advisory when drop advisory so it's child won't be blocked when push

### DIFF
--- a/elliott/elliottlib/cli/advisory_drop_cli.py
+++ b/elliott/elliottlib/cli/advisory_drop_cli.py
@@ -8,7 +8,7 @@ from requests_gssapi import HTTPSPNEGOAuth
 from elliottlib.cli.common import cli
 from elliottlib.constants import errata_drop_url
 from elliottlib.exceptions import ElliottFatalError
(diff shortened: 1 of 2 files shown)


vfreex (dismissed):
Did you test this?
I doubt this is going to work. It doesn't seem legitimate to remove blocking advisories with dependent advisories.

Ximinhan:
> Did you test this? I doubt this is going to work. It doesn't seem legitimate to remove blocking advisories with dependent advisories.

I updated the pr, the case can be
rpm advisory has the config
```
depends_on: extra advisory (remove item need to call /remove_blocking_errata)
blocks: microshift advisory (remove item need to call /remove_dependent_errata)
```
and we will call /remove_dependent_errata clear the blocks from child side while in microshift advisory
```
depends_on: rpm advisory
blocks:
```
it still has depends_on to the dropped rpm advisory and it make push failed, so we need to remove rpm from microshift advisory's depends_on

joepvd (approved):
/lgtm

PR #159, opened 2023-11-07
ART-8188: Deal with missing containerfile

joepvd:


joepvd:
[Test build](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Focp4/49395/console)
Resulting in [this tree](https://pkgs.devel.redhat.com/cgit/containers/openstack-cluster-api-controllers/tree/?h=rhaos-4.15-rhel-8&id=309a5f1e55cddf420ba1c392eb0b1b4fdc360cd9)
Resulting in [task](https://brewweb.engineering.redhat.com/brew/taskinfo?taskID=56797713)

ashwindasr:
As per https://github.com/containers/buildah/discussions/3170, this could warrant further investigation to confirm if we are passing the correct dockerfile to OSBS

ashwindasr:
As joep pointed out, OSBS modifies our Dockerfile as well, so this is the best solution for now

PR #160, opened 2023-11-08
ART-5914 Trigger-build-sync-automatically after gen-assembly

locriandev:
[gen-assembly](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/dpaolell/job/gen-assembly/18/console) started [build-sync](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fbuild-sync/39352/). Since this feature needs a custom art-tools (and aos-cd-jobs)  that run has been aborted in favor of a [manual one](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/dpaolell/job/build-sync/26/console)

build-sync had the bot [comment at triggering time](https://github.com/openshift-eng/ocp-build-data/pull/3860#issuecomment-1801856800), and again once build-sync completed ([here](https://github.com/openshift-eng/ocp-build-data/pull/3860#issuecomment-1801876782)).

Moreover, the bot sent a [message](https://redhat-internal.slack.com/archives/C02TN2SJR0F/p1699450632210449) on Slack to communicate that build-sync completed, by stating its result.

This PR adds some refactoring bits, such as:
- create `artcommonlib.util` as a common place for utility functions
- add `artcommon.tests`
-  update Makefile to execute those tests along with doozer/elliott/pyartcd ones


pyartcd/pyartcd/pipelines/build_sync.py
```diff
@@ -52,8 +55,70 @@ def __init__(self, runtime: Runtime, version: str, assembly: str, publish: bool,
         self.logger = runtime.logger
         self.working_dir = self.runtime.working_dir
         self.fail_count_name = f'count:build-sync-failure:{assembly}:{version}'
+        self.job_run = get_build_url()
+
+        self.slack_client = self.runtime.new_slack_client()
+        self.slack_client.bind_channel(f'openshift-{self.version}')
+
+    async def comment_on_assembly_pr(self, text_body):
```
Ximinhan, line 63: I think we can move the comment action in gen_assembly since pr object already have the context there and reduce the checking work. we only need much plugin in build_sync only need the result, and handle github operation in the same place.

pyartcd/pyartcd/pipelines/gen_assembly.py
```diff
@@ -164,50 +156,75 @@ async def _gen_assembly_from_releases(self, candidate_nightlies: Iterable[str])
...
+        else:
+            pull_number = existing_prs.items[0].number
+            result = api.pulls.update(pull_number=pull_number, title=title, body=body)
+
+        # Trigger build-sync
+        if self.auto_trigger_build_sync:
+            self._logger.info("Triggering build-sync")
+            build_version = self.group.split("-")[1]  # eg: 4.14 from openshift-4.14
+            # we're not passing doozer_data_path to build-sync because we always create branch on the base repo
+            start_build_sync(build_version=build_version,
```
Ximinhan, line 223: `result` has the pr number, we can check the build_sync job result, add the comment here also notify in the same thread.

doozer/doozerlib/assembly_inspector.py
```diff
@@ -315,8 +317,9 @@ def check_group_image_consistency(self, build_inspector: BrewBuildImageInspector
         content_git_url = image_meta.config.content.source.git.url
         if content_git_url:
             # Make sure things are in https form so we can compare
-            content_git_url, _ = self.runtime.get_public_upstream(util.convert_remote_git_to_https(content_git_url))
-            build_git_url = util.convert_remote_git_to_https(build_inspector.get_source_git_url())
+            content_git_url, _ = self.runtime.get_public_upstream(
+                artcommonlib.util.convert_remote_git_to_https(content_git_url))
```
Ximinhan, line 321: we can use `import import artcommonlib.util as artutil` and use shorter syntax here `artutil.convert_remote_git_to_https(content_git_url)`

pyartcd/pyartcd/pipelines/gen_assembly.py
```diff
@@ -164,50 +156,75 @@ async def _gen_assembly_from_releases(self, candidate_nightlies: Iterable[str])
...
+                             assembly=self.assembly,
+                             doozer_data_path=self.data_path,
+                             doozer_data_gitref=branch)
+
```
Ximinhan, line 227: if build_sync passed I think we can auto merge the pr after here.

pyartcd/pyartcd/pipelines/build_sync.py
locriandev: How would gen-assembly know about build-sync outcome? `comment_on_assembly_pr()` is called when build-sync completed, indicating either its success or failure. I don't see how this could be moved to gen-assembly

pyartcd/pyartcd/pipelines/gen_assembly.py
locriandev: `build-sync` is started in non-blocking mode. I think this is fine, as we don't want to have these two jobs tightly coupled. This means gen-assembly won't care about `build-sync` result, it will just fire it and exit

pyartcd/pyartcd/pipelines/gen_assembly.py
locriandev: I would still leave the merging duty to @release-artists. Once build-sync completes, they get a Slack notification, can inspect and decide.

doozer/doozerlib/assembly_inspector.py
locriandev: Done in https://github.com/openshift-eng/art-tools/pull/160/commits/472105997483d39d8c8068bd30b200b79b719fdc

vfreex (commented):
/lgtm

pyartcd/pyartcd/pipelines/build_sync.py
Ximinhan: it can wait and get the job run result https://github.com/openshift-eng/art-tools/blob/main/pyartcd/pyartcd/jenkins.py#L220 , if gen_assembly start with trigger build-sync parameter then it need to wait build-sync complete and do error alert or success comment there

pyartcd/pyartcd/pipelines/gen_assembly.py
Ximinhan: I think it should care about the result, like if success just comments with `api.issues.create_comment(issue_number=pr_number, body=text_body)` after

Ximinhan: slack messages can also move to here, so it can append in the thread

pyartcd/pyartcd/pipelines/build_sync.py
locriandev: yeah that can be done, but it's a different approach. I still prefer to keep the two jobs loosely coupled, so I'd rather not have gen-assembly block and wait. Besides, unless I missed some important facts, it doesn't see relevant to have a blocking approach. The implemented code is already handling this by having build-sync infer PR info and comment there

locriandev: Besides, I think having build-sync comment about its own results is more appropriate than having gen-assembly do so. I don't want to establish a dependency relationship between the two

PR #162, opened 2023-11-09
[ART-6149] Fix comment on PR feature

ashwindasr:
The comment on PR feature aims to comment on the PR that kicked off an ocp4 build for images. The feature was disabled because multiple messages were being posted on the same PR ([example](https://github.com/openshift/network-tools/pull/75)). But it was because I was checking whether the entire message was there. For new messages, the NVR will change. This issue is now fixed by checking if there is any comment with `[ART PR BUILD NOTIFIER]`

### DIFF
--- a/doozer/doozerlib/comment_on_pr.py
+++ b/doozer/doozerlib/comment_on_pr.py
@@ -1,12 +1,17 @@
+import os
 from ghapi.all import GhApi
 from doozerlib.pushd import Dir
 from dockerfile_parse import DockerfileParser
+from doozerlib.constants import BREWWEB_URL, GITHUB_TOKEN
 
 
 class CommentOnPr:
-    def __init__(self, distgit_dir: str, token: str):
+    def __init__(self, distgit_dir: str, nvr: str, build_id: str, distgit_name: str):
         self.distgit_dir = distgit_dir
-        self.token = token
+        self.nvr = nvr
+        self.build_id = build_id
+        self.distgit_name = distgit_name
+        self.token = os.getenv(GITHUB_TOKEN)
         self.owner = None
         self.repo = None
         self.commit = None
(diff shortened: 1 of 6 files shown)


vfreex (commented):
/lgtm

locriandev:
Changes LGTM. Looking at the code, I think the interface for the `CommentOnPr` class should be simplified. All we need to know from outside packages is how to construct that object, and how to call the `post_comment()` function. Methods like
- `set_repo_details()`
- `set_github_client()`
- `set_pr_from_commit()`

can be called internally, thus made protected. This will simplify usages for that class, that could be needed somewhere else in the future.

Of course, this can be in the scope of a new PR, unless it's a simple change so I wouldn't miss the chance to kick that in ;)

locriandev:
/hold
to give @ashwindasr a chance to read the comment before deciding to merge

doozer/doozerlib/comment_on_pr.py
```diff
@@ -1,11 +1,15 @@
 from ghapi.all import GhApi
 from doozerlib.pushd import Dir
 from dockerfile_parse import DockerfileParser
+from doozerlib.constants import BREWWEB_URL
 
 
 class CommentOnPr:
-    def __init__(self, distgit_dir: str, token: str):
+    def __init__(self, distgit_dir: str, token: str, nvr: str, build_id: str, distgit_name: str):
```
locriandev, line 8: `token` is just `os.getenv(constants.GITHUB_TOKEN)`, so it can be inferred locally and not passed in as a constructor parameter

ashwindasr:
Issues resolved, squashed commits, re-adding lgtm

ashwindasr:
Hoping to get feedback from upstream and then improving on this feature.

PR #170, opened 2023-11-14
Comment on source PR only for stream

ashwindasr:
We don't want to [comment on upstream PRs](https://github.com/openshift/cluster-olm-operator/pull/32#issuecomment-1809702693) for test builds


locriandev (commented):
/lgtm

PR #171, opened 2023-11-14
[ART-8232] Setup FIPS scan on ART cluster

ashwindasr:
Ref https://issues.redhat.com/browse/ART-7953

We now require to run the FIPS check using [check-payload](https://github.com/openshift/check-payload) tool. This command needs to be run with `root`. Setting up this pipeline to automatically run after new images are built (after ocp4), by passing the NVRs as parameters.


ashwindasr:
/cc @vfreex 

I made an [update](https://github.com/openshift-eng/art-tools/commit/9ca856eea8ebfe2ac8446ec15bd3f4e8288c7d34) to the slack_client as well. Can you take a look when you get a chance?

pyartcd/pyartcd/pipelines/scan_fips.py
```diff
@@ -0,0 +1,64 @@
...
+    def __init__(self, runtime: Runtime, version: str, nvrs: Optional[list]):
+        self.runtime = runtime
+        self.version = version
+        self.nvrs = nvrs
+
+        # Setup slack client
+        self.slack_client = self.runtime.new_slack_client()
+        self.slack_client.bind_channel(f"openshift-{self.version}")
+
+    async def run(self):
```
Ximinhan, line 23: can we import doozer lib to use the function directly?

ashwindasr: Hmm @Ximinhan I'm not sure if I follow you? Which part of the code can we get from doozerlib? :thinking:

locriandev: I think Ximin might be suggesting to do something like:

```
    async def run(self):
        fips_pipeline = doozerlib.cli.scan_fips.ScanFipsCli(runtime=runtime,
                                                                                                 nvrs=nvrs.split(",") if self.nvrs else None)
        fips_pipeline.run()
```

Instead of running Doozer in a separate process, we could import doozerlib and run it directly. Only thing is, `doozerlib.cli.scan_fips.ScanFipsCli.run()` does not return anything, but prints to stdout. So we'd need to change that. Although I am not sure this is worth the pain, we generally call Doozer in its own process from pyartcd.

doozer/doozerlib/cli/scan_fips.py
```diff
@@ -0,0 +1,55 @@
...
+
+        problem_images = {}
+        for nvr, pull_spec in image_pullspec_mapping:
+            # Run fips scan
+            rc, _, _ = cmd_gather(f"check-payload scan image --spec {pull_spec}")
+
+            if rc != 0:
+                problem_images[nvr] = pull_spec
+
+        print(json.dumps(problem_images))
```
locriandev, line 45: I think we generally use `click.echo` for this

doozer/doozerlib/cli/scan_fips.py
```diff
@@ -0,0 +1,55 @@
...
+        # Initialize runtime and brewhub session
+        self.runtime.initialize(clone_distgits=False)
+        self.koji_session = koji.ClientSession(self.runtime.group_config.urls.brewhub)
+
+    def run(self):
+        # Get the list of NVRs to scan for
+        # (nvr, pull-spec) list of tuples
+        image_pullspec_mapping = []
+
+        for nvr in self.nvrs:
```
locriandev, line 28: We could use `exectools.parallel_exec()` to get some parallelism and save some time. If you think this is worth it, you can see an example [here](https://github.com/openshift-eng/art-tools/blob/main/doozer/doozerlib/cli/scan_sources.py#L167)

doozer/doozerlib/cli/scan_fips.py
```diff
@@ -0,0 +1,55 @@
...
+            # Find the registry pull spec
+            build_info = self.koji_session.getBuild(nvr)
+
+            # Eg registry-proxy.engineering.redhat.com/rh-osbs/openshift-ose-sriov-network-operator@sha256:da95750d31cb1b9539f664d2d6255727fa8d648e93150ae92ed84a9e993753be
+            # from https://brewweb.engineering.redhat.com/brew/buildinfo?buildID=2777601
+            pull_spec = build_info["extra"]["image"]["index"]["pull"][0]
+            image_pullspec_mapping.append((nvr, pull_spec))
+
+        problem_images = {}
+        for nvr, pull_spec in image_pullspec_mapping:
```
locriandev, line 38: It might be worth (as this is an easy change) to use `cmd_gather_async` instead, and run all these commands in a pool of async tasks. Something like:

```

class ScanFipsCli:

   ...

    async def check_payload(self, pullspec, problems):
        rc, _, _ = await cmd_gather_async(f"check-payload scan image --spec {pull_spec}")
        if rc != 0:
            problems[nvr] = pull_spec

    async def run(self):
        ...

        problem_images = {}
        await asyncio.gather(*[self.check_payload(pullspec, problem_images) 
                               for nvr, pull_spec in image_pullspec_mapping])
```

ashwindasr: Thank you! Added in https://github.com/openshift-eng/art-tools/pull/171/commits/29c790f396924732039cd3e027ca8a355e59d936

doozer/doozerlib/cli/scan_fips.py
ashwindasr: Thank you! Added in https://github.com/openshift-eng/art-tools/pull/171/commits/29c790f396924732039cd3e027ca8a355e59d936

doozer/doozerlib/cli/scan_fips.py
ashwindasr: getBuild is usually pretty quick. So I think we can skip for now. Once the feature is out, I'll test this out against a mass rebuild and see how it goess

pyartcd/pyartcd/pipelines/scan_fips.py
ashwindasr: Hmm I haven't tried that out before, but if you guys think its worth it, I can work on the change

Ximinhan: yes, I mean maybe we can import ScanFipsCli here, ScanFipsCli(runtime=runtime,nvr=self.nvrs)
the pr is going to run "check-payload scan image --spec" this cli it doesn't seems to be need tightly bonded with doozer, can we directly call it here?

art-cluster/pipelines/data/project/art-cd/image/files/krb5-redhat.conf
```diff
@@ -0,0 +1,32 @@
+[libdefaults]
```
locriandev, line 1: We have several copies of this same file at:
- ./doozer/.devcontainer/krb5-redhat.conf
- ./elliott/.devcontainer/krb5-redhat.conf
- ./.devcontainer/krb5-redhat.conf

and there's only a couple of differences from this one. Maybe worth unifying?

pyartcd/pyartcd/pipelines/scan_fips.py
ashwindasr: > this cli it doesn't seems to be need tightly bonded with doozer, can we directly call it here
You're right. But since the current pattern in our code base is to create a doozer command then calling the art-cd command, didn't want to break that pattern.

ashwindasr: > directly use pull spec can save some effort to get the build object again in the job

Is this scenario when humans run the command? Cause in that case the fastest would be to directly use the `check-payload scan image --spec <pull-spec>` command right.

art-cluster/pipelines/data/project/art-cd/image/files/krb5-redhat.conf
ashwindasr: Changed in https://github.com/openshift-eng/art-tools/pull/171/commits/d8ab9eaf9a928c2e83679881ec2fb75f4336b323

ashwindasr: Created a separate PR https://github.com/openshift-eng/art-tools/pull/490

ashwindasr:
/hold

until https://github.com/openshift-eng/art-tools/pull/490 is merged an rebased

ashwindasr:
/hold

Needs rebase https://github.com/openshift-eng/art-tools/pull/500

PR #173, opened 2023-11-15
Fix rhcos version in promote

locriandev:
For 4.15, due to machine-os-content having disappeared from the payload, promote job is failing on 
```
rhcos_version = rhcos["annotations"]["io.openshift.build.versions"].split("=")[1]  # machine-os=48.84.202112162302-0 => 48.84.202112162302-0
KeyError: 'io.openshift.build.versions'
```

The `version` label is always found and can be used as an alternative, as it appears in all versions since 4.6:
```
versions=(
  "4.14.2"
  "4.13.22"
  "4.12.43"
  "4.11.53"
  "4.10.67"
  "4.9.59"
  "4.8.57"
  "4.7.60"
  "4.6.62"
)

for version in ${versions[@]}; do
  echo $version;
  oc image info -o json $(oc adm release info --image-for machine-os-content ${version}) | jq -r '.config.config.Labels.version';
  echo "------------------------";
done
```

### DIFF
--- a/pyartcd/pyartcd/pipelines/promote.py
+++ b/pyartcd/pyartcd/pipelines/promote.py
@@ -419,7 +419,7 @@ class PromotePipeline:
             rhcos_name = get_primary_container_name(self.group_runtime)
             rhcos = next((t for t in release_info.get("references", {}).get("spec", {}).get("tags", []) if t["name"] == rhcos_name), None)
             if rhcos:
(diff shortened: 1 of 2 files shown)


joepvd:
/approve
This looks great. Yet:
/hold
as I think probably rhcos should +1 this implementation

PR #176, opened 2023-11-16
Skip when no nvrs to scan

ashwindasr:
Fixes issue in [job](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fscan-osh/13773/console), where `nvrs_for_scans` is an empty list


joepvd (approved):
/lgtm

PR #181, opened 2023-11-17
[scan-osh] Run for all variants of job

ashwindasr:
Create JIRA tickets when ART is building the images as well


ashwindasr:
Merged without lgtm because the --create-tickets logic wasn't being executed

PR #184, opened 2023-11-17
[scan-osh] Update title with version, dry_run

ashwindasr:
For better title description


ashwindasr:
Wont work since job is in ocp-artifacts

/close

PR #194, opened 2023-11-21
[scan-osh] Update OCPBUGS workflow run

ashwindasr:
Changes in this PR related to OCPBUGS workflow:
- Run only for the scheduled variant of the job
- Check OSH only once per brew component. _(Sometimes more than one build of a component can be found after a run. So running just once, for the latest build NVR)_

_Ready to merge, requesting /lgtm and /approve_


joepvd (approved):
/lgtm

PR #198, opened 2023-11-22
[scan-osh] Skip component if can't file imp file

ashwindasr:
Due to an upstream bug, sometimes we don't have the scan-imp file. Lets skip that component now.

### DIFF
--- a/doozer/doozerlib/cli/scan_osh.py
+++ b/doozer/doozerlib/cli/scan_osh.py
@@ -148,8 +148,11 @@ class ScanOshCli:
         self.runtime.logger.info(f"Checking OSH Scan for issues: {url}")
         response = requests.get(url)
 
-        if len(response.json()["defects"]) > 0:
-            return True
+        try:
+            if len(response.json()["defects"]) > 0:
+                return True
+        except Exception:
+            self.runtime.logger.warning(f"Don't see scan-results-imp.js for {nvr}")
 
         return False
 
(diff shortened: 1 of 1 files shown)


doozer/doozerlib/cli/scan_osh.py
```diff
@@ -363,9 +363,13 @@ def get_packages_with_scan_issues(self, mapping: dict):
                 self.runtime.logger.info(f"No successful scan found for package {brew_package_name}")
                 continue
 
-            if self.check_if_scan_issues_exist(task_id=data["latest_coverity_scan"],
-                                               nvr=data["latest_coverity_scan_nvr"]):
-                nvrs_with_scan_issues[brew_package_name] = data
+            try:
+                if self.check_if_scan_issues_exist(task_id=data["latest_coverity_scan"],
```
thegreyd, line 367: maybe better to have try/except block in check_if_scan_issues_exist, log and then return False ?

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -363,9 +363,13 @@ def get_packages_with_scan_issues(self, mapping: dict):
...
+                                                   nvr=data["latest_coverity_scan_nvr"]):
+                    nvrs_with_scan_issues[brew_package_name] = data
+            except Exception:
+                self.runtime.logger.warning(f"Don't see scan-results-imp.js. Skipping {brew_package_name}")
```
thegreyd, line 371: think this is an anti-pattern - including internal implementation details of a func outside of it

doozer/doozerlib/cli/scan_osh.py
ashwindasr: Thanks, updated in https://github.com/openshift-eng/art-tools/pull/198/commits/292176b0bfab66bb36c0f6d4252c48d2a2470a25

doozer/doozerlib/cli/scan_osh.py
ashwindasr: Ahh! Noted! Thank you!

Hope https://github.com/openshift-eng/art-tools/pull/198/commits/292176b0bfab66bb36c0f6d4252c48d2a2470a25 addresses that

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -148,8 +148,11 @@ def check_if_scan_issues_exist(self, task_id: str, nvr: str) -> bool:
         self.runtime.logger.info(f"Checking OSH Scan for issues: {url}")
         response = requests.get(url)
 
-        if len(response.json()["defects"]) > 0:
-            return True
+        try:
+            if len(response.json()["defects"]) > 0:
```
thegreyd, line 152: I think we want to include requests.get line inside try as well?

ashwindasr: Hmm I thought that line will never fail, like it will have a failed status code but it would break the code right.

But updated nevertheless https://github.com/openshift-eng/art-tools/pull/198/commits/6af2210ba3e775168f1f2aae7689612c1aa474bd

PR #199, opened 2023-11-23
WIP: Do not exempt when strictfipsruntime is in GOEXPERIMENT

joepvd:


joepvd:
/close
Think [ths](https://github.com/openshift/check-payload/pull/143) solves the concrete issue at hand in the better way.

PR #201, opened 2023-11-23
Use new target version field

thegreyd:
Jira admins changed the current field to "Target Version - old"
this is the new "Target Version"
https://issues.redhat.com/rest/api/2/field


joepvd (approved):
/lgtm

PR #205, opened 2023-11-24
update operator-sdk job update subtask function

Ximinhan:
replace the way using subtask index locates subtask to using keyword searching locate subtask
and auto complete  pushes release content to customer portal task by checking if the release pushed to fast channel

### DIFF
--- a/pyartcd/pyartcd/jira.py
+++ b/pyartcd/pyartcd/jira.py
@@ -29,16 +29,14 @@ class JIRAClient:
     def start_task(self, key):
         self._client.transition_issue(key, 'In Progress')
 
-    # For release jira ticket
-    # subtask_id = 6 relate to subtask [Fri/Mon] QE moves advisories to REL_PREP
-    # subtask_id = 7 relate to subtask [Once REL_PREP] ART provides non-golang container-first sources and operator-sdk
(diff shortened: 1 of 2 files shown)


pyartcd/pyartcd/jira.py
```diff
@@ -29,16 +29,14 @@ def close_task(self, key):
...
-    # subtask_id = 8 relate to subtask [Mon-Wed] EXD CLOUDDST pushes advisory content to production CDN
-    def complete_subtask(self, parent_jira_id, subtask_id, comment):
+    def complete_subtask(self, parent_jira_id, subtask_keyword, comment):
         parent_jira = self.get_issue(parent_jira_id)
-        subtask = self.get_issue(parent_jira.fields.subtasks[subtask_id].key)
-        self.add_comment(subtask, comment)
-        self.assign_to_me(subtask)
-        self.close_task(subtask)
+        for subtask in parent_jira.fields.subtasks:
+            if subtask_keyword in subtask.fields.summary:
```
vfreex, line 35: An exception might be needed if no subtasks is found.

Ximinhan: if a ticket has no subtask it will return [], won't raise exception

thegreyd:
/lgtm
This is something that was on my list the last time I was fiddling with subtasks in the codebase, but could never get to it.. thanks for tackling it

PR #206, opened 2023-11-24
bug sweep: look more precise at cutoff event

joepvd:
OCPBUGS-21035 got ON_QA in October, got back to Assigned, got fixed again. The logic for establishing if this was fixed at the cutoff time is short circuited, as the question "was it ON_QA or higher before cutoff time?" is Yes.

This changes the query to ask for the status at the cutoff time.

The `WAS` operator is explained here: https://support.atlassian.com/jira-software-cloud/docs/jql-operators/


locriandev (commented):
/lgtm

PR #210, opened 2023-11-28
Refactor doozer rpms: commands to rpms.py

ashwindasr:
Refactor doozer commands that has `rpms:` as prefix to `rpms.py`

Tested commands with `doozer rpms:* -h` which correctly printed out the output


locriandev (commented):
/lgtm

PR #211, opened 2023-11-28
Include disabled rpms flag for rpms:print

ashwindasr:
Follows https://github.com/openshift-eng/art-tools/pull/209

Since microshift is disabled, it won't show up in this command, without specifying the disabled flag. We need this option so that our SAST pipeline can see which all RPMs are managed by us.

_Requesting /lgtm and /approve_


ashwindasr:
Yeah I think we can skip WIP for now

PR #213, opened 2023-11-29
Update ocp4 and build-sync job to honor doozer_data_path

Ximinhan:


pyartcd/pyartcd/pipelines/build_sync.py
```diff
@@ -33,19 +33,19 @@ async def create(cls, *args, **kwargs):
         self = cls(*args, **kwargs)
         self.group_runtime = await GroupRuntime.create(
             self.runtime.config, self.working_dir,
-            self.group, self.assembly
+            self.group, self.assembly, self.doozer_data_path, self.doozer_data_gitref
         )
         return self
 
-    def __init__(self, runtime: Runtime, version: str, assembly: str, publish: bool, data_path: str,
+    def __init__(self, runtime: Runtime, version: str, assembly: str, publish: bool, doozer_data_path: str,
```
locriandev, line 40: This looks unnecessary, and pollutes the actual change with lots of changed lines with no reason. Isn't `data_path` clear enough?

pyartcd/pyartcd/pipelines/ocp4.py
```diff
@@ -661,9 +661,15 @@ def _handle_image_build_failures(self):
     async def _build_images(self):
         self.runtime.logger.info('Building images')
 
+        group_config = await util.load_group_config(
+            group=f'openshift-{self.version.stream}',
+            assembly=self.assembly,
+            doozer_data_path=self.data_path
+        )
         signing_mode = await util.get_signing_mode(
```
locriandev, line 669: I think the best way to fix this is provide `get_signing_mode` function with `assembly` and `data_path` params, and let it load the group config if needed

locriandev (changes requested):

pyartcd/pyartcd/pipelines/ocp4.py
Ximinhan: the get_signing_mode function will use passed group_config as a key parameter, so if group_config not passed, it will use group and assembly to load again, let's make the explicit loading outside the function

pyartcd/pyartcd/pipelines/ocp4_scan.py
```diff
@@ -170,16 +170,18 @@ async def _rhcos_inconsistent(self):
 
 @cli.command('ocp4-scan')
 @click.option('--version', required=True, help='OCP version to scan')
+@click.option('--data-path', required=False, default=constants.OCP_BUILD_DATA_URL,
```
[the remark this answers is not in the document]
Ximinhan: when do branch cut I used this job to check all images have new branch setup, so I think it's different with released usecase

pyartcd/pyartcd/pipelines/build_sync.py
Ximinhan: the issue is GroupRuntime.create function will reload the self args, thus in other functions they all use doozer_data_path and if we pass self.data_path it will raise an error with self.doozer_data_path not found

pyartcd/pyartcd/pipelines/ocp4_scan.py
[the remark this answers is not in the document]
locriandev: We probably have other ways to check for the existence of the desired target branches, this doesn't sound like a good reason to let scan-sources run against forks (that would trigger ocp4, rhcos, build-sync, etc), and it would be useful only at branch time, dangerous otherwise

pyartcd/pyartcd/pipelines/ocp4.py
locriandev: This is exactly what I mean: `get_signing_mode` should be updated to accept assembly and data path, so it can work effectively against forks/assemblies. It could be useful in other places, and we wouldn't need to load group config outside it every time

Ximinhan: the get_signing_mode is just got software_lifecycle from config then do a translate, maybe we can define a config in group.yml like
```
software_lifecycle:
  phase: pre-release
  signing: unsigned
```
and then get rid of the function

pyartcd/pyartcd/pipelines/ocp4_scan.py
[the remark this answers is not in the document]
Ximinhan: ok, I will hold this change

pyartcd/pyartcd/pipelines/ocp4.py
locriandev: In this case, I think `software_lifecycle` might just go away. AFAIK, we're only using the pre-release/release/eol fields to decide about signing mode.

locriandev: Although, I would rather do this in a separate task, because we'd need to change lots of places in the code, and group config for all versions.

Ximinhan: ok I will create a separate task for signing mode [ART-8480](https://issues.redhat.com/browse/ART-8480)

PR #214, opened 2023-11-29
[scan-osh] Comment on bugs ticket with scan urls

ashwindasr:
Recently we had issues in which a new OCPBUGS ticket was opened even though no new issues were found. Seems like this happens when there are two scans for the same NVR. When ART triggers a scan, we check to see if there is none triggered for that NVR. So if we comment on the ticket, we will be able to see which two scans the automation looked at, at that point in time, since the status changes after a while.

### DIFF
--- a/doozer/doozerlib/cli/scan_osh.py
+++ b/doozer/doozerlib/cli/scan_osh.py
@@ -314,6 +314,14 @@ class ScanOshCli:
 
                     self.runtime.logger.info(f"Linked {previous_ticket_id.key} to {issue.key}")
 
+                # Report the latest and previous scan on the ticket as a comment
+                comment = f"This ticket was raised because ART automation found new issues in the " \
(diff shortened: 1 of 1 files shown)


doozer/doozerlib/cli/scan_osh.py
```diff
@@ -314,6 +314,14 @@ def create_update_ocpbugs_ticket(self, packages: dict):
 
                     self.runtime.logger.info(f"Linked {previous_ticket_id.key} to {issue.key}")
 
+                # Report the latest and previous scan on the ticket as a comment
+                comment = f"This ticket was raised because ART automation found new issues in the " \
```
thegreyd, line 318: any reason why this goes as a comment and not in the ticket description?

ashwindasr: I didn't want to pollute the ticket with too much info. But do you think its better to keep in the description?

thegreyd (approved):

doozer/doozerlib/cli/scan_osh.py
thegreyd: Let's try it out like this

PR #216, opened 2023-11-30
Update #aos-art reference

locriandev:
Fix ART public channel name in messages like this: https://github.com/openshift/cluster-api/pull/188#issuecomment-1832276683


joepvd (approved):
/lgtm

PR #217, opened 2023-11-30
ART-8413 Exact jira title search

thegreyd:
https://issues.redhat.com/browse/ART-8413


joepvd:
Awesome!
/approve
Think we want at least to log when `open_issues` has length bigger than 1.

thegreyd:
> Awesome! /approve Think we want at least to log when `open_issues` has length bigger than 1.

Any strong reason? Looking at a search like [project = OCPBUGS And summary ~ "Update 4.15 ose-cluster-api-container image to be consistent with ART"](https://issues.redhat.com/issues/?jql=project%20%3D%20OCPBUGS%20And%20summary%20~%20%22Update%204.15%20ose-cluster-api-container%20image%20to%20be%20consistent%20with%20ART%22) , the other search results are irrelevant..  logging them wouldn't be very helpful. But if you think otherwise, can fix it in

PR #221, opened 2023-12-01
ART-8114 gen-assembly configures prerelease advisories

thegreyd:
https://issues.redhat.com/browse/ART-8114
Design doc: https://docs.google.com/document/d/1S_ivD8Sh85LuZKnClqvoRvIhn4hNJQA76jpfUaFZjUA/edit

## Summary
- Add `--prerelease` flag to gen-assembly, to indicate a candidate assembly is a prerelease 
  - https://github.com/openshift-eng/aos-cd-jobs/pull/4069
- ECs are always prerelease
- Inherit advisories and release jira from last EC in RC.0
- Do not inherit prerelease advisory
- Configure operator index mode to prerelease when indicated.

Prepare release will use these appropriately: https://github.com/openshift-eng/art-tools/pull/263

## Test
```
$ doozer --assembly stream --group openshift-4.15 --arches x86_64 release:gen-assembly --name rc.1 from-releases --nightly 4.15.0-0.nightly-2023-12-19-033450 --prerelease
...
      group:
        advisories:
          extras: 123982
          image: 123983
          metadata: 123984
          microshift: 123985
          prerelease: -1
          rpm: 123986
        operator_index_mode: pre-release
        release_jira: ART-8258
```

### DIFF
--- a/doozer/doozerlib/cli/release_gen_assembly.py
+++ b/doozer/doozerlib/cli/release_gen_assembly.py
@@ -477,6 +477,10 @@ class GenAssemblyCli:
             'metadata': -1,
         }
 
+        # ECs should always have an associated prerelease advisory
+        if self.assembly_type == AssemblyTypes.PREVIEW:
+            advisories['prerelease'] = -1
+
         # For OCP >= 4.14, also microshift advisory placeholder must be created
         major, minor = self.runtime.get_major_minor_fields()
         if (major, minor) >= (4, 14):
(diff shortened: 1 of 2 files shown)


thegreyd:
/hold 
see discussion on this behavior https://redhat-internal.slack.com/archives/GDBRP5YJH/p1701461075698179?thread_ts=1701459682.373229&cid=GDBRP5YJH

doozer/doozerlib/cli/release_gen_assembly.py
```diff
@@ -517,6 +546,8 @@ def _generate_assembly_definition(self) -> dict:
 
         if self.assembly_type not in [AssemblyTypes.CUSTOM, AssemblyTypes.PREVIEW]:
             group_info['advisories'], group_info["release_jira"] = self._get_advisories_release_jira()
+            if group_info['advisories']['prerelease'] == -1:
+                group_info['operator_index_mode'] = 'pre-release'
```
sosiouxme, line 550: should this be outdented? I think we'll only need this when `assembly_type` is `PREVIEW`

also `['prerelease']` won't always be defined so shouldn't it be `.get('prerelease')`?

sosiouxme (commented):
other than comment, release propagation looks great!

doozer/doozerlib/cli/release_gen_assembly.py
thegreyd: Ah good catch, I've updated the block to not run only for custom assembly

thegreyd:
I chose to not introduce separate job params for indicating pre/advance release right now. So if we want to do that, we would edit the assembly definition to introduce the advisory key and prepare-release would run their workflow https://github.com/openshift-eng/art-tools/pull/263 - I will update the docs to indicate this manual step

PR #224, opened 2023-12-04
[scan-osh] Skip if NVRs are same, minor fixes

ashwindasr:
Changes in this PR:
- If the previous successful scan is a duplicate scan from the same NVR, skip that
- Update `SCAN_RESULTS_URL_TEMPLATE` url since we need the URL without the raw suffix to comment on the ticket.


joepvd (approved):
/lgtm

PR #225, opened 2023-12-04
[scan-osh] after ocp4, do not trigger duplicate scans

ashwindasr:
The scheduled variant of SAST scan only triggers if there isn't already one. But the ocp4 job does not. This PR aims to fix that.


joepvd (approved):
/lgtm

PR #227, opened 2023-12-05
ART-8427: get rhcos version from release

Ximinhan:
rhel-coreos annotions can be empty, get rhcos version from release


locriandev (commented):
/lgtm

PR #231, opened 2023-12-05
use --load-disabled 

ashwindasr:
Rolls back https://github.com/openshift-eng/art-tools/pull/229

Use `--load-disabled` base doozer param instead of separate `--show-disabled` or `--include-disabled` param


sosiouxme (approved):
/lgtm

PR #233, opened 2023-12-05
[scan-osh] enable OCPBUGS workflow by default

ashwindasr:
group.yml will activate for all images. If we need to disable any particular image, we can override in that image config, by setting it to false

### DIFF
--- a/doozer/doozerlib/cli/scan_osh.py
+++ b/doozer/doozerlib/cli/scan_osh.py
@@ -407,13 +407,14 @@ class ScanOshCli:
         return nvrs_with_scan_issues
 
     @staticmethod
-    def is_sast_jira_enabled(meta):
+    def is_sast_jira_disabled(meta):
         """
         Check if the OCPBUGS workflow is enabled in ocp-build-data config
         """
         flag = meta.config.get("external_scanners", {}).get("sast_scanning", {}).get("jira_integration", {}).get("enabled")
 
-        return flag in [True, "True", "true", "yes"]
+        # Disable only if specifically defined in config
+        return flag in [False, "False", "false", "no"]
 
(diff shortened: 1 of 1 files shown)


doozer/doozerlib/cli/scan_osh.py
```diff
@@ -407,13 +407,14 @@ def get_packages_with_scan_issues(self, mapping: dict):
         return nvrs_with_scan_issues
 
     @staticmethod
-    def is_sast_jira_enabled(meta):
+    def is_sast_jira_disabled(meta):
         """
-        Check if the OCPBUGS workflow is enabled in ocp-build-data config
+        Check if the OCPBUGS workflow is enabled in ocp-build-data image meta config
         """
         flag = meta.config.get("external_scanners", {}).get("sast_scanning", {}).get("jira_integration", {}).get("enabled")
```
jupierce, line 414: With "group as the default for image metadata", I would have expected:
```
flag = meta.config.external_scanners.sast_scanner.jira_integration.enabled
if flash is Missing:
  flag = meta.runtime. group_config.external_scanners.sast_scanning.jira_integration.enabled
```

and no code like honoring group globally: https://github.com/openshift-eng/art-tools/blob/0fa8d84a7fcad697e168bcbcb5ecbfbe7e0a0eba/doozer/doozerlib/cli/scan_osh.py#L662

Meaning that you could enable a singular image in a given release without turning it on for everything in group. Given the nature of the scanner, it is unlikely that we will require that configuration in practice, so I offer this for future reference only and won't hold up this PR. lgtm.

ashwindasr: Noted, thanks for pointing this out!

I have to rewrite some of the code once osh-cli adds a feature that I requested. Will add this along with that change.

PR #236, opened 2023-12-06
[scan-osh] trigger only for scheduled job

ashwindasr:
We had been triggering the job after ocp4 since our initial requirement was to scan images that we were building. Since the requirement had evolved to scan everything in our candidate tags, a scheduled job was setup for that purpose. 

During the scheduled job, all the brew builds since the previous run will be collected and the workflow executed.


locriandev (approved):
/lgtm

PR #237, opened 2023-12-06
scan-sources pattern match for exempted packages

thegreyd:
- Introduce new config "exempt_rpms" intended to replace "exempted_packages". Right now both of them will be effective.
- Accept glob pattern matching 

After we have moved over all ocp-build-data branches from exempted_packages to exempt_rpms, 
I'll make a PR to remove the exempted_packages config from here


joepvd:
/unhold
This seems safe 👍🏼

PR #239, opened 2023-12-07
ART-8191 Deprecate unused code paths in create advisory

thegreyd:
Refactoring create advisory cli - https://issues.redhat.com/browse/ART-8191

## Summary
- Deprecate attaching bugs when creating advisory (this was never getting used)
- Deprecate setting impact and cve_names in internal func (we don't allow rhsa creation, so this was never getting used)
- Remove support for using values from outside of "boilerplate" config key in erratatool.yml - all our branches are updated now, error out if not.
- Require boilerplate_name explicitly, do not rely on "kind" for default
- Deprecate add_metadata_cli - the command and comment is not being used by anything 
- Deprecate using kind in placeholder bug - this will help us reduce complexity where it is not needed - placeholder bug title can be generic
- Deprecate impetus and kind when creating advisory and just rely on boilerplate name

After this goes - we can deprecate non boilerplate values in https://github.com/openshift-eng/ocp-build-data/blob/openshift-4.14/erratatool.yml (top level keys like synopsis)

## Test
```
./elliott --group openshift-4.15 create --art-advisory-key prerelease --type RHBA --assigned-to openshift-qe-errata@redhat.com --manager vlaad@redhat.com --package-owner lmeyer@redhat.com --date 2025-Nov-01
```


Ximinhan:
/lgtm
I think we can also deprecate the placeholder function, currently, it should be unused

PR #240, opened 2023-12-07
[ART-7926] [SAST] add support for RPMs, skip exluded components

ashwindasr:
Updates in this PR:
- Exclude components defined in ocp-build-data ([4.15 example](https://github.com/openshift-eng/ocp-build-data/blob/e2ec932f1594e3f5be49ab6d1b9003218c6f81ce/group.yml#L522)).
- Refactor to optimize logic for components managed and not managed by ART.
- Add support for RPMs.
- For RPMs and non-ART images, using distgit URL as upstream.
- RPMs that are version independent, only raise one ticket instead of one in each version.

_Requesting /lgtm only_


ashwindasr:
/hold

Want to time the merge.

PR #243, opened 2023-12-08
[ART-8256] Add Art-Dash Link to Health:Images Reports

MartinPl0:
This commit integrates Art-Dash links into the health:images reports. The goal is to enhance accessibility and traceability of the reports by providing direct links to Art-Dash for detailed build histories. This streamlines the process for analyzing build concerns and improves overall efficiency.

Key Changes:
- `generate_art_dash_history_link` to create Art-Dash URLs based on build records.
- Updated `images_health` to incorporate these Art-Dash links.
- Modified YAML output in `images_health` to handle URLs correctly and prevent YAML dump from adding unwanted quotes.

This enhancement is part of ongoing efforts to improve our build monitoring tools, simplifying the workflow for developers and release managers. A regex post-processing step has been added to ensure clarity and readability of the reports, maintaining user-friendly YAML formatting despite complex URL strings.

### DIFF
--- a/doozer/doozerlib/cli/images_health.py
+++ b/doozer/doozerlib/cli/images_health.py
@@ -1,11 +1,13 @@
 # stdlib
 import collections
 import datetime
-import yaml
 import time
+import urllib.parse
+import re
 
 # external
 import click
+import yaml
 
 # doozerlib
 from doozerlib.cli import cli, pass_runtime
@@ -16,6 +18,24 @@ BuildInfo = collections.namedtuple('BuildInfo', 'record_name, task_id task_state
 millis_hour = 1000 * 60 * 60
 millis_day = millis_hour * 24
 now_unix_ts = int(round(time.time() * 1000))  # millis since the epoch
+
+
+def generate_art_dash_history_link(dg_name, runtime):
+    base_url = "https://art-dash.engineering.redhat.com/dashboard/build/history"
+
+    # Validating essential parameters
+    if not dg_name or not runtime or not runtime.group_config or not runtime.group_config.name:
+        raise ValueError("Missing essential parameters for generating Art-Dash link")
+
+    formatted_dg_name = dg_name.split("/")[-1]
+
+    params = {
+        "group": runtime.group_config.name,
+        "dg_name": formatted_dg_name,
+    }
+
+    query_string = urllib.parse.urlencode(params)
+    return f"{base_url}?{query_string}"
 
 
 @cli.command("images:health", short_help="Create a health report for this image group (requires DB read)")
@@ -37,7 +57,14 @@ def images_health(runtime, limit, url_markup):
         runtime.logger.info('No concerns to report!')
         return
 
-    print(yaml.dump(concerns, default_flow_style=False, width=10000))
+    # Dump the YAML to a string
+    yaml_output = yaml.dump(concerns, default_flow_style=False, width=10000)
+
+    # Use a regular expression to remove single quotes from the start and end of lines starting with a hyphen
+    pattern = re.compile(r"^-\s+'(.*?)'$", re.MULTILINE)
+    modified_yaml_output = pattern.sub(r"- \1", yaml_output)
+
+    print(modified_yaml_output)
 
 
 def get_concerns(image, runtime, limit, url_markup):
@@ -66,7 +93,6 @@ def get_concerns(image, runtime, limit, url_markup):
     latest_success_bi_dt = ''
 
     for idx, record in enumerate(records):
-        # record=( jobid, state, timestamp, joburl)
         if record[1] == 'success':
             latest_success_idx = idx
             latest_success_bi = record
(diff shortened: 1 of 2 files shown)


doozer/doozerlib/cli/images_health.py
```diff
@@ -85,9 +132,10 @@ def url_text(url, text):
         if latest_success_idx == -1:
             # No success record was found
             msg += f'Failing for at least the last {len(records)} attempts / {oldest_attempt_bi_dt}'
+            msg += f'. See more details: {art_dash_link}'
```
Ximinhan, line 135: I think this line is not necessary

doozer/doozerlib/cli/images_health.py
```diff
@@ -85,9 +132,10 @@ def url_text(url, text):
...
         else:
             msg += f'Last {url_text(latest_success_bi_task_url, "success")} was {latest_success_idx} attempts ago on {latest_success_bi_dt}'
-
+        msg += f'. See more details: {art_dash_link}'
```
Ximinhan, line 138: use url_text to add hyperlink to the text instead of directly show the link url

doozer/doozerlib/cli/images_health.py
```diff
@@ -66,7 +112,8 @@ def url_text(url, text):
     latest_success_bi_dt = ''
 
     for idx, record in enumerate(records):
-        # record=( jobid, state, timestamp, joburl)
+        # Generate the art-dash link for each record
+        art_dash_link = generate_art_dash_history_link(image, record, runtime)
```
Ximinhan, line 116: why we need to get all records link? I think this loop is used for calculate index

doozer/doozerlib/cli/images_health.py
```diff
@@ -18,6 +20,43 @@
...
+def convert_to_iso_format(unix_timestamp_ms):
+    # Convert milliseconds to seconds
+    timestamp_seconds = unix_timestamp_ms / 1000.0
+    # Create a UTC datetime object from the timestamp
+    dt = datetime.datetime.utcfromtimestamp(timestamp_seconds)
+    # Format the datetime object to ISO 8601 format without fractional seconds
+    return dt.replace(microsecond=0).isoformat() + 'Z'
+
+
+def generate_art_dash_history_link(dg_name, build_record, runtime):
```
Ximinhan, line 32: I think only group and dg_name are needed, other params seem not useful

MartinPl0: Also correct, wanted to make sure the right build is always found so used all available parameters. If we dont need it then (as you say) I will change it.

doozer/doozerlib/cli/images_health.py
MartinPl0: I think we should provide art-dash link on build failure, no matter if the latest build is failing or not.

doozer/doozerlib/cli/images_health.py
MartinPl0: Ah i understand slack markup style now, your implementation is correct.

doozer/doozerlib/cli/images_health.py
```diff
@@ -77,17 +108,24 @@ def url_text(url, text):
...
+        art_dash_link = generate_art_dash_history_link(image, records[0], runtime)
+
         msg = f'Latest attempt {url_text(latest_attempt_task_url, "failed")} ({url_text(latest_attempt_build_url, "jenkins job")}); '
+
         # The latest attempt was a failure
         if latest_success_idx == -1:
             # No success record was found
-            msg += f'Failing for at least the last {len(records)} attempts / {oldest_attempt_bi_dt}'
+            msg += f"Failing for at least the last {len(records)} attempts / {oldest_attempt_bi_dt}"
+            msg += f'. See more details: {url_text(art_dash_link, "art-dashboard link")}'
```
Ximinhan, line 125: this msg can be put out of the if else

doozer/doozerlib/cli/images_health.py
Ximinhan: we need multi-builds for history not only one task, right?

doozer/doozerlib/cli/images_health.py
```diff
@@ -18,6 +20,28 @@
...
+    formatted_dg_name = dg_name.split("/")[-1]
+
+    params = {
+        "group": runtime.group_config.name,
+        "dg_name": formatted_dg_name,
+        "brew_task_id": task_id,
+    }
+
+    # Exclude any empty or irrelevant parameters
+    filtered_params = {k: v for k, v in params.items() if v}
```
Ximinhan, line 39: in which case item will be empty? if the param is empty we should return earlier and raise bad params, but when it would happened

MartinPl0: ohhh, right that makes sense. Will implement this.

MartinPl0: In normal operation, these parameters are expected to be present and correctly populated. However, the checks for empty parameters are a precautionary measure.

doozer/doozerlib/cli/images_health.py
MartinPl0: ah okay, so the art-dash link should not lead to just that build failure but to history of the build, this makes sense now and also follows Joeps strawhat proposal in the ticket.

PR #246, opened 2023-12-11
ART-8476 doozer images:rebase assert RHEL version equivalence in Dockerfiles

locriandev:
This PRs introduces 2 changes during images rebase:

1. Also rebase base images (rhel) as well as golang builders

2. If we're trying to match upstream config during rebase, add an extra layer to the Dockerfile to make sure that upstream RHEL version matches the one we have configured locally. Break the build otherwise.

This will let developers use their preferred builder versions before feature freeze, but will enforce that there is agreement upon an image RHEL version

### DIFF
--- a/doozer/doozerlib/distgit.py
+++ b/doozer/doozerlib/distgit.py
@@ -1528,7 +1528,7 @@ class ImageDistGitRepo(DistGitRepo):
     def _mapped_image_from_stream(self, image, original_parent, dfp):
         stream = self.runtime.resolve_stream(image.stream)
 
-        if not self._should_match_upstream():
+        if not self.should_match_upstream:
             # Do typical stream resolution.
             return stream.image
 
@@ -1558,11 +1558,11 @@ class ImageDistGitRepo(DistGitRepo):
             out, _ = exectools.cmd_assert(cmd, retries=3)
 
             # It does. Use this to rebase FROM directive
-            digest = json.loads(out)['digest']
+            mapped_image_data = json.loads(out)
 
-            mapped_image = f'{labels["name"]}@{digest}'
             # if upstream equivalent does not match ART's config, add a warning to the Dockerfile
-            if mapped_image != stream.image:
+            mapped_image = f'{labels["name"]}@{mapped_image_data["digest"]}'
+            if mapped_image_data['name'].split(':')[-1] != stream.image.split(":")[-1]:
                 dfp.add_lines_at(
                     0,
                     "",
@@ -1589,6 +1589,7 @@ class ImageDistGitRepo(DistGitRepo):
 
     def _rebase_from_directives(self, dfp):
         image_from = Model(self.config.get('from', None))
+        self.should_match_upstream = self._should_match_upstream()
 
         # Collect all the parent images we're supposed to use
         parent_images = image_from.builder if image_from.builder is not Missing else []
@@ -1825,6 +1826,16 @@ class ImageDistGitRepo(DistGitRepo):
                     filtered_content.append(line)
 
             df_lines = filtered_content
+
+            # ART-8476 assert rhel version equivalence
+            if self.should_match_upstream:
+                el_version = util.isolate_el_version_in_brew_tag(self.branch)
+                df_lines.extend([
+                    '',
+                    '# RHEL version in final image must match the one in ART\'s config',
+                    f'RUN source /etc/os-release && [[ "$PLATFORM_ID" == platform:el{el_version} ]]'
+                ])
+
             df_content = "\n".join(df_lines)
 
             if release:
--- a/doozer/doozerlib/distgit.py
+++ b/doozer/doozerlib/distgit.py
@@ -1612,26 +1612,26 @@ class ImageDistGitRepo(DistGitRepo):
         image_from = Model(self.config.get('from', None))
 
         # Collect all the parent images we're supposed to use
-        parent_images = image_from.builder if image_from.builder is not Missing else []
-        parent_images.append(image_from)
-        if len(parent_images) != len(dfp.parent_images):
+        downstream_parents = image_from.builder if image_from.builder is not Missing else []
+        downstream_parents.append(image_from)
+        if len(downstream_parents) != len(dfp.parent_images):
             raise IOError(
                 "Build metadata for {name} expected {count1} image parent(s), but the upstream Dockerfile "
                 "contains {count2} FROM statements. These counts must match. Detail: '{meta_parents}' vs "
                 "'{upstream_parents}'.".format(
                     name=self.config.name,
-                    count1=len(parent_images),
+                    count1=len(downstream_parents),
                     count2=len(dfp.parent_images),
-                    meta_parents=parent_images,
+                    meta_parents=downstream_parents,
                     upstream_parents=dfp.parent_images,
                 ))
         mapped_images = []
 
-        original_parents = dfp.parent_images
-        for i, image in enumerate(parent_images):
+        upstream_parents = dfp.parent_images
+        for i, image in enumerate(downstream_parents):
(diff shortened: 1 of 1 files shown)


doozer/doozerlib/util.py
```diff
@@ -420,6 +420,14 @@ def isolate_el_version_in_brew_tag(tag: Union[str, int]) -> Optional[int]:
     return int(el_version_match[1]) if el_version_match else None
 
 
+def isolate_rhel_version_in_distgit_branch(branch_name: str) -> int:
```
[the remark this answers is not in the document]
locriandev: Right, updated

locriandev:
/hold
Currently only `builder` stages are changed according to upstream. `member` stages (like ose-base) are left unchanged. This needs to be changed to, as these are usually the last build stages and determine the final rhel version

doozer/doozerlib/distgit.py
```diff
@@ -1394,10 +1394,66 @@ def _clean_repos(self, dfp):
...
+
+            # Verify whether the image exists
+            self.logger.debug('Checking for upstream equivalent existence, pullspec: %s', upstream_equivalent_pullspec)
+            cmd = f'oc image info {upstream_equivalent_pullspec} --filter-by-os linux/amd64 -o json'
+            out, _ = exectools.cmd_assert(cmd, retries=3)
+
+            # It does. Use this to rebase FROM directive
+            mapped_image_data = json.loads(out)
+            mapped_image_data_labels = mapped_image_data['config']['config']['Labels']
+            mapped_image = f'{labels["name"]}:{mapped_image_data_labels["version"]}-{mapped_image_data_labels["release"]}'
```
thegreyd, line 1425: This will always be the same as the values in line 1405?
`build_nvr = f'{labels["com.redhat.component"]}-{labels["version"]}-{labels["release"]}'`
so we can just use those

doozer/doozerlib/distgit.py
```diff
@@ -1611,7 +1627,7 @@ def _rebase_from_directives(self, dfp):
         for i, image in enumerate(parent_images):
             # Does this image inherit from an image defined in a different group member distgit?
             if image.member is not Missing:
-                mapped_images.append(self._mapped_image_from_member(image, original_parents, count))
+                mapped_images.append(self._mapped_image_from_member(image, original_parents[i], dfp))
```
thegreyd, line 1630: we can now delete `count`?

thegreyd, line 1630: I am slightly confused regarding the usage of  `original_parents`, `parent_images` and `image`. Seems like these were created so we don't modify and lose their original values? But they aren't a deepcopy ?
- on line 1637, image and original_parents[i] would be the same
- throughout the func original_parents and parent_images would be the same

doozer/doozerlib/distgit.py
```diff
@@ -1394,10 +1394,66 @@ def _clean_repos(self, dfp):
...
+
+            # Query Brew for build info
+            self.logger.debug('Retrieving info for Brew build %s', build_nvr)
+            with self.runtime.shared_koji_client_session() as koji_api:
+                if not koji_api.logged_in:
+                    koji_api.gssapi_login()
+                build = koji_api.getBuild(build_nvr, strict=True)
+
+            # Get the pullspec for the upstream equivalent
+            upstream_equivalent_pullspec = build['extra']['image']['index']['pull'][1]
```
thegreyd, line 1415: nitpick, we call this "upstream_equivalent_pullspec", imo it would be simpler to call it "downstream_image_pullspec"

doozer/doozerlib/distgit.py
```diff
@@ -1394,10 +1394,66 @@ def _clean_repos(self, dfp):
...
+            build_nvr = f'{labels["com.redhat.component"]}-{labels["version"]}-{labels["release"]}'
+
+            # Query Brew for build info
+            self.logger.debug('Retrieving info for Brew build %s', build_nvr)
+            with self.runtime.shared_koji_client_session() as koji_api:
+                if not koji_api.logged_in:
+                    koji_api.gssapi_login()
+                build = koji_api.getBuild(build_nvr, strict=True)
+
+            # Get the pullspec for the upstream equivalent
```
thegreyd, line 1414:
```suggestion
            # Get the pullspec for the upstream equivalent
            # registry-proxy.engineering.redhat.com/rh-osbs/openshift-golang-builder:v1.20.10-202310161945.el8.gdc4b478
```
Having an example value to look at would be helpful

doozer/doozerlib/distgit.py
```diff
@@ -1528,66 +1584,26 @@ def _should_match_upstream(self) -> bool:
...
+        if not parent:
             return stream.image
+        else:
+            return parent
 
     def _rebase_from_directives(self, dfp):
+        self.should_match_upstream = self._should_match_upstream()
+        if self.should_match_upstream:
+            # ART-8476 override image config with upstream
+            alt_upstream = self.config.get('alternative_upstream', None)
```
thegreyd, line 1602: If we're using https://gist.github.com/jupierce/1e9eaae40a47a3fd097d5d4fbcf3b93c#file-gistfile1-txt-L40-L44 as an example, it has a `- when: el8`, we would want to validate that here, since upstream could be using rhel9 but this would mean we use rhel8 config.

thegreyd: We can also separate out implementing this alt_upstream change in another PR, and let this PR through

locriandev: I think for the moment we can just an alternative config we use when `canonical_builders` is True, and blindly apply it if needed. To get more flexible, we sure need a couple of `when` clauses, but we need to implement a way to determine upstream rhel version. I'd do this as a separate ticket

doozer/doozerlib/distgit.py
locriandev: This is really the pullspec of the image that upstream wants to use, since it is deduced directly from the upstream Dockerfile. Naming it `downstream` would sound confusing to me

doozer/doozerlib/distgit.py
locriandev: Done in ea500e9de3b27acbc42e98c576995a353a09c1c8

doozer/doozerlib/distgit.py
locriandev: Labels slightly differ because `original_parent` is an image we mirror. Nonetheless, `release` and `version` appear to be the same:

```
10,11c10,11
<     "io.k8s.description": "ART equivalent image openshift-4.15-golang - rhel-8/golang",
<     "io.k8s.display-name": "builder-rhel-8-golang-1.20-openshift-4.15",
---
>     "io.k8s.description": "golang builder image for Red Hat internal builds",
>     "io.k8s.display-name": "Go Builder 1.20",
14,15d13
<     "io.openshift.build.name": "builder-rhel-8-golang-1.20-openshift-4.15--art-builder-474",
<     "io.openshift.build.namespace": "ci",
```

so I think it's fine to reuse the labels. `mapped_image` will need `name` instead of `com.redhat.component` though. 

Done in a809cf2bebdd0406c6d793da33902125afcaecd2

doozer/doozerlib/distgit.py
locriandev: `parent_images` is what we have in our yaml config; `original_parents` is the result of parsing Dockerfile FROM clauses. Maybe we could rename:
- `parent_images` into `downstream_parents`
- `original_parents` into `upstream_parents`

That should be enough to clarify. Done in 6fc899c6ca15b0093c5482be4d4bd8887bcb5ed6

thegreyd:
lgtm

locriandev:
/hold

small fix still needed

PR #247, opened 2023-12-11
[ART-7926] [scan-osh] Update  OCPBUGS workflow

ashwindasr:
For RPMs, we need to raise one ticket per rhel version, not ocp version. Hence updating the workflow to reflect that, while including optimizations. This will be a periodic scan set to run every hour.

How the new workflow will look like:
- Retrieve all builds from brew candidate tags, from [erratatool.yml](https://github.com/openshift-eng/ocp-build-data/blob/e2ec932f1594e3f5be49ab6d1b9003218c6f81ce/erratatool.yml#L8)
- Get all the latest builds from those tags
- If scan issues exist for those NVRs, raise a ticket (if there isn't already one)
- RPM tickets will be raised with title having rhel version as the prefix, while images will OCP version as the prefix.

Note:
Since this new workflow will immediatelly open tickets for a version, since all latest builds are checked (in the previous workflow, we were raising tickets for new builds only) need to deactivate for versions < 4.15 temporarily so that we can phase it out: 
https://github.com/openshift-eng/ocp-build-data/pull/4043

### DIFF
--- a/doozer/doozerlib/cli/scan_osh.py
+++ b/doozer/doozerlib/cli/scan_osh.py
@@ -14,7 +14,6 @@ from doozerlib.rpmcfg import RPMMetadata
 from typing import Optional
 from jira import JIRA, Issue
 from tenacity import retry, stop_after_attempt, wait_fixed
-from doozerlib.rpm_utils import parse_nvr
 
 SCAN_RESULTS_URL_TEMPLATE = "https://cov01.lab.eng.brq2.redhat.com/osh/task/{task_id}/log/{nvr}/scan-results-imp.js"
 
@@ -24,13 +23,15 @@ One or more software security scan issues have been detected for the package *{p
 *Important scan results*: [html|https://cov01.lab.eng.brq2.redhat.com/osh/task/{scan_id}/log/{nvr}/scan-results-imp.html] / [json|https://cov01.lab.eng.brq2.redhat.com/osh/task/{scan_id}/log/{nvr}/scan-results-imp.js] (Check 'defects')
 *All scan results*: https://cov01.lab.eng.brq2.redhat.com/osh/task/{scan_id}
 *Build record*: [{nvr}|https://brewweb.engineering.redhat.com/brew/buildinfo?buildID={brew_build_id}]
-*Upstream commit*: [{upstream_git_commit}|{upstream_repo_url}/commit/{upstream_git_commit}]
+*Upstream commit*: {upstream_commit}
 
 Once these issues are addressed, the Bug can be closed. But a new one will be opened if new issues are found in the next build.
 
 For information on triaging these issues and dispositioning this ticket, please review the [policy document for SAST Scanning Jira Tickets|https://docs.google.com/document/d/1GsT3tvw5qf3sOSAkTSVWnMIO6vzIuirhsVhhsbK_qf0].
 
 For other questions please reach out to @release-artists in #forum-ocp-art.
+
+_(Ticket title and description are maintained by the bot. Please do not modify)_
 """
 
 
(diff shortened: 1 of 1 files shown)


ashwindasr:
/hold

Until https://github.com/openshift-eng/ocp-build-data/pull/4043 and all its cherry-picks have merged.

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -64,6 +68,60 @@ def __init__(self, runtime: Runtime, last_brew_event: int, dry_run: bool, check_
...
+    def is_jira_workflow_component_disabled(meta):
+        """
+        Check if the OCPBUGS workflow is enabled in ocp-build-data image meta config
+        """
+        flag = meta.config.external_scanners.sast_scanning.jira_integration.enabled
+
+        # Disable only if specifically defined in config
+        return flag in [False, "False", "false", "no"]
+
+    def is_jira_workflow_group_disabled(self):
```
thegreyd, line 82:
```suggestion
    def is_jira_workflow_group_enabled(self):
```
This to me is more intuitive, since the default is to keep it `disabled`

doozer/doozerlib/cli/scan_osh.py
```diff
@@ -64,6 +68,60 @@ def __init__(self, runtime: Runtime, last_brew_event: int, dry_run: bool, check_
...
+        Check if JIRA workflow is disabled for this particular OCP version in group.yml
+        """
+        flag = self.runtime.group_config.external_scanners.sast_scanning.jira_integration.enabled
+        return flag not in [True, "True", "true", "yes"]
+
+    @staticmethod
+    def get_substring_regex(pattern, string):
+        match = re.search(pattern=pattern, string=string)
+        if match:
+            return match.group("search_key")
```
thegreyd, line 93: Why write a function for this? Also since we rely on search_key being in regex in the calling code,
it is better if we have both of them in the same place

doozer/doozerlib/cli/scan_osh.py
ashwindasr: updated in https://github.com/openshift-eng/art-tools/pull/247/commits/3153a81eca090e9df7930ea6c7fa6546572b4596

doozer/doozerlib/cli/scan_osh.py
ashwindasr: Well I needed to do string extractions using regex in more than one place, so I wrote the function to reduce redundancy. 

> Also since we rely on search_key being in regex in the calling code
Ahh yes, I missed this during refinement. Can I pass the search_key as a variable inside the function? Like

```
@staticmethod
    def get_substring_regex(pattern, string, search_key):
        match = re.search(pattern=pattern, string=string)
        if match:
            return match.group(search_key)
```

ashwindasr: like this https://github.com/openshift-eng/art-tools/pull/247/commits/4a74f2e74309158f496b9b17bad8a915442916b5

thegreyd: Imo it's okay to have this kind of redundancy, it helps with readability and maintainability. Defining a function to wrap an if statement, that we later check with an assert statement isn't reducing the cognitive burden to understand what's going on.

ashwindasr: makes sense, will make the change

thegreyd:
It's hard to see what parts have changed and what remain same / just have changed places, refactored slightly. Functions like get_tagged_latest, get_tagged_all, is_latest_scan_different_from_previous, get_untriggered_nvrs have barely changed? but appear interleaved in the diff. Is there a way to separate the refactoring changes from the actual changes?

doozer/doozerlib/cli/scan_osh.py
ashwindasr: updated in https://github.com/openshift-eng/art-tools/pull/247/commits/7a0cbcff199ae0e320344b186720e19dc0a292eb

ashwindasr:
> It's hard to see what parts have been removed/added and what remain same/changed places.

Apologies for that @thegreyd. I should have raised a separate PR for refactoring, after I made the changes to the workflow. I rewrote most of the code for ocpbugs workflow since we had to start from rhaos tags for RPMs

ashwindasr:
/unhold

https://github.com/openshift-eng/ocp-build-data/pull/4043 and all its cherry-picks have merged.

PR #248, opened 2023-12-11
Refactor doozer images:* commands

ashwindasr:
Migrating all `doozer images:*` commands to images.py for better readability.


locriandev (commented):
/lgtm

PR #260, opened 2023-12-15
ART-8476 handle when clause

locriandev:
With these changes, `alternative_upstream` should look like this:
```
alternative_upstream:
  - when: el8
    distgit:
      branch: rhaos-{MAJOR}.{MINOR}-rhel-8
    enabled_repos:
    - rhel-8-baseos-rpms
    - rhel-8-appstream-rpms
    name: openshift/ose-cluster-autoscaler-rhel8
```

### DIFF
--- a/doozer/doozerlib/distgit.py
+++ b/doozer/doozerlib/distgit.py
@@ -13,7 +13,7 @@ import shutil
 import sys
 import time
 import traceback
-from datetime import date, datetime
+from datetime import datetime
 from multiprocessing import Event, Lock
 from typing import Dict, List, Optional, Tuple, Union, Set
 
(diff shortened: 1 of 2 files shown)


doozer/doozerlib/distgit.py
```diff
@@ -1606,15 +1613,57 @@ def _mapped_image_from_stream(self, image, original_parent, dfp):
...
+                if alt_config['when'] == self.upstream_intended_el_version:
+                    self.logger.info('Merging %s alternative config to match upstream', self.upstream_intended_el_version)
+                    self.config.update(alt_config)
+                    matched = True
+                    break
+            if not matched:
+                # there's no 'when' clause matching upstream intended RHEL version: fallback to default ART's config
+                self.logger.warning('Could not match upstream because "%s" version is not mapped in alternative_config',
+                                    self.upstream_intended_el_version)
+                self.should_match_upstream = False
```
thegreyd, line 1662: This would mean that if there is no alt_config defined explicitly, then we won't attempt to use what's upstream .. I'm not sure we want to do that. I think we still want images to use their own upstream golang builder and with the recent change attempt to use their member, if builds fail then we fix it with alt config

locriandev: I have removed line `1662`. If there is no alt config defined, we will attempt to use upsteam but the build will most likely break, as golang and rhel versions won't match. We'll get a Slack notification and be able to react

PR #263, opened 2023-12-18
ART-8115 Modify prepare-release job for prereleases

thegreyd:
https://issues.redhat.com/browse/ART-8115

Summary
- Add and use `--permissive` mode to `elliott find-bugs:sweep` to avoid erroring out on invalid bugs when we prep ECs and RCs.
- Build and attach bundle builds to prerelease advisories
- Skip attach-cve-flaws when we're prepping a prerelease assembly
- Gather dependencies from other advisories when prepping prerelease advisory
- Use `rhose async auto` release for preGA advisories - with https://github.com/openshift-eng/ocp-build-data/pull/4101
- Bring all prep-release slack messages in a single thread

### DIFF
--- a/pyartcd/pyartcd/pipelines/prepare_release.py
+++ b/pyartcd/pyartcd/pipelines/prepare_release.py
@@ -19,6 +19,7 @@ from tenacity import retry, stop_after_attempt, wait_fixed
 from doozerlib.assembly import AssemblyTypes
 from elliottlib.assembly import assembly_group_config
 from elliottlib.errata import set_blocking_advisory, get_blocking_advisories
+from elliottlib.errata import get_brew_builds
 from elliottlib.model import Model
 from pyartcd import exectools
 from pyartcd.cli import cli, click_coroutine, pass_runtime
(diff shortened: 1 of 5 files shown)


pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -195,21 +195,26 @@ async def run(self):
...
-        self.sweep_bugs()
+        if assembly_type in (AssemblyTypes.PREVIEW, AssemblyTypes.CANDIDATE):
+            self.sweep_bugs(permissive=True)
+        else:
+            self.sweep_bugs()
 
-        _LOGGER.info("Processing attached Security Trackers")
-        for _, advisory in advisories.items():
-            self.attach_cve_flaws(advisory)
+        if 'prerelease' in advisories:
```
joepvd, line 212: Thinking that `assembly_type in (AssemblyTypes.PREVIEW, AssemblyTypes.CANDIDATE)` should skip flaw bug processing. There will be RCs that do not have a prerelease advisory, and I think we want to postpone CVE stuff until we're close to GA. 

Or is there a `software_lifecyckle.phase` that would match with the desired here?

thegreyd: Yeah that was one of the things I wasn't sure about from the doc if Luke meant us to run cve stuff intermittently or all at end. It makes sense to run all at end, while still reporting on errors with --permissive. I'll make the change

thegreyd: Done

joepvd (commented):
/approve

PR #267, opened 2023-12-19
Track Jenkins job run on alignment PRs

locriandev:
When opening an alignment PR, it can be useful to track the URL of the Jenkins run that caused this PR to be open, for debugging purposes. The URL will be provided in a separate comment, thus leaving the PR description unchanged.


doozer/doozerlib/cli/images_streams.py
```diff
@@ -1346,6 +1346,10 @@ def check_if_upstream_image_exists(upstream_image):
                     pr_title = f'Bug {bug}: {pr_title}'
                 try:
                     new_pr = public_source_repo.create_pull(title=pr_title, body=pr_body, base=public_branch, head=fork_branch_head, draft=draft_prs)
```
thegreyd, line 1348: Are we also updating the PR here? It might be that PR is updated by a latter job, not a big issue though

locriandev: No this only happens at first PR creation

PR #268, opened 2023-12-19
Protect build-sync for stream assembly with a Lock

locriandev:
[First build](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/dpaolell/job/build-sync/6/console) (lock acquired)

[Second build](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/dpaolell/job/build-sync/7/console) (trying to acquire the lock)

### DIFF
--- a/pyartcd/pyartcd/locks.py
+++ b/pyartcd/pyartcd/locks.py
@@ -15,6 +15,7 @@ class Lock(enum.Enum):
     BUILD = 'lock:build:{version}'
     MASS_REBUILD = 'lock:mass-rebuild-serializer'
     SIGNING = 'lock:signing:{signing_env}'
+    BUILD_SYNC = 'lock:build-sync:{version}'
 
 
 # This constant defines for each lock type:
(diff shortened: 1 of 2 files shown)


pyartcd/pyartcd/locks.py
```diff
@@ -15,6 +15,7 @@ class Lock(enum.Enum):
     BUILD = 'lock:build:{version}'
     MASS_REBUILD = 'lock:mass-rebuild-serializer'
     SIGNING = 'lock:signing:{signing_env}'
+    BUILD_SYNC = 'lock:build-sync:{version}'
```
thegreyd, line 18: So I think we need version-assembly here, running build-sync on 4.15 stream and say, 4.15.0-rc.1 parallely should be fine

thegreyd: ah I see we only use this when assembly is stream

PR #269, opened 2023-12-19
Scan-osh: Don't error out early

thegreyd:
There is a bug which is wiping descriptions in OCPBUGS on editing https://issues.redhat.com/browse/RH1-5087
which breaks our runs. Instead of erroring out early on individual tickets, wrap it and raise at end


vfreex (commented):
/lgtm

PR #275, opened 2023-12-20
ART-7803: Edit go.mod with ci build root

joepvd:


doozer/doozerlib/cli/images_streams.py
```diff
@@ -886,6 +887,31 @@ def images_streams_prs(runtime, github_access_token, bug, interstitial, ignore_c
...
+
+        if not golang:
+            return
+
+        gomods = [Path(repo_dir, f'{p}/go.mod') for p in image_config.cachito.packages.gomod]
+        if not gomods:
+            gomods = [Path(repo_dir, 'go.mod')]
+        for gomod in gomods:
+            if gomod.is_file():
+                exectools.cmd_assert(f'go mod edit -go={golang} {gomod}')
```
thegreyd, line 912: so this would mean a go binary running on buildvm.. 
Can we run this in hackspace / or something less risky but still tests this ?

thegreyd: I see it on buildvm so good with it

joepvd:
/hold

we probably do not want to open a pr if go.mod is the only change.

doozer/doozerlib/cli/images_streams.py
```diff
@@ -1174,6 +1201,12 @@ def check_if_upstream_image_exists(upstream_image):
 
                 exectools.cmd_assert(f'git add -f {str(ci_operator_config_path)}')
 
+                if desired_ci_build_root_image and 'golang' in desired_ci_build_root_image:
+                    try:
```
thegreyd, line 1205: Code looks good :+1:  Can we add a comment explaining why for this odd looking block, help future us and code readers

PR #278, opened 2023-12-21
Move duplicated functions to artcommon

locriandev:
Move these functions from doozer/elliott into artcommon as they're duplicated:
- go_arch_for_brew_arch
- go_suffix_for_arch
- brew_arch_for_go_arch
- brew_suffix_for_arch

Also move arch names/suffixes to artcommon. Improve unit tests


thegreyd:
/lgtm
Nice work! Would recommend squashing commits before merge

PR #285, opened 2024-01-03
gen-assembly: Ignore outdated payload tag

thegreyd:
To get around the issue of outdated payload tag in nightly imagestreams
`WARNING Ignoring payload tag cluster-config-api-rhel9 since payload_name=cluster-config-api is explicitly defined in image config`


vfreex (commented):
/lgtm
/approve

PR #292, opened 2024-01-05
[ART-8119] skip prerelease advisory from expected bugs check

ashwindasr:
Skip expected bugs check for tracker bugs related to builds attached to the prerelease advisory.


elliott/elliottlib/cli/verify_attached_bugs_cli.py
```diff
@@ -188,7 +188,10 @@ def verify_bugs_advisory_type(self, non_flaw_bugs, advisory_id_map, advisory_bug
         bugs_by_type = categorize_bugs_by_type(non_flaw_bugs, advisory_id_map, permitted_bug_ids=permitted_bug_ids)
         for kind, advisory_id in advisory_id_map.items():
             actual = {b for b in advisory_bug_map[advisory_id] if b.is_ocp_bug()}
-            expected = bugs_by_type[kind]
+            expected = bugs_by_type.get(kind)
+            if not expected:
+                # Looks like impetus is not part of the ones we care about. Skipping bug attachment
+                continue
```
thegreyd, line 194:
```suggestion
            # If impetus is not present, assume no bugs need to be attached 
            expected = bugs_by_type.get(kind, set())
```

elliott/elliottlib/cli/verify_attached_bugs_cli.py
```diff
@@ -188,7 +188,10 @@ def verify_bugs_advisory_type(self, non_flaw_bugs, advisory_id_map, advisory_bug
         bugs_by_type = categorize_bugs_by_type(non_flaw_bugs, advisory_id_map, permitted_bug_ids=permitted_bug_ids)
```
thegreyd, line 188:
```suggestion
        bugs_by_type = categorize_bugs_by_type(non_flaw_bugs, advisory_id_map, permitted_bug_ids=permitted_bug_ids, permissive=True)
```
Can you add this as well? Recently we added this param to allow ignoring bugs that don't look right (like cve trackers). We want to have this so that the verification runs fully, otherwise errors short-circuit our verification flow.

elliott/elliottlib/cli/verify_attached_bugs_cli.py
ashwindasr: > For "advance" operator advisory we will attach bugs to it
So for that we just need to add a line [here](https://github.com/openshift-eng/art-tools/blob/9331f7a357e88cda10f4c9f36dac5d51fa09ac06/elliott/elliottlib/cli/find_bugs_sweep_cli.py#L274) right, for `advance`.

elliott/elliottlib/cli/verify_attached_bugs_cli.py
ashwindasr: added in https://github.com/openshift-eng/art-tools/pull/292/commits/382f5459e2be12d8dd2ebdd9160efe47527af105

PR #297, opened 2024-01-08
[scan-osh] Fix RPM ticket update bug

ashwindasr:
Earlier builds were replacing later builds since for RPMs we do not consider OCP version.

### DIFF
--- a/doozer/doozerlib/cli/scan_osh.py
+++ b/doozer/doozerlib/cli/scan_osh.py
@@ -220,7 +220,7 @@ class ScanOshCli:
             self.runtime.logger.error("Regex match not found")
 
     @staticmethod
-    def get_scan_id_from_ticket(description):
+    def get_scan_details_from_ticket(description):
         pattern = r"\*All scan results\*: https://cov01.lab.eng.brq2.redhat.com/osh/task/(?P<task_id>[0-9]+)"
         match = re.search(pattern=pattern, string=description)
         task_id = None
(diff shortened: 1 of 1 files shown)


doozer/doozerlib/cli/scan_osh.py
```diff
@@ -579,6 +580,18 @@ def create_update_jira_ticket(self, component):
                 # Looks like it's the same NVR
                 return
 
+            if kind == BuildType.RPM:
+                # Get the build in the currently "open" ticket
+                _, build_nvr_on_ticket = self.get_scan_details_from_ticket(description=issue.fields.description)
+                build_on_ticket = self.koji_session.getBuild(build_nvr_on_ticket)
+
+                if build_on_ticket["build_id"] > build["build_id"]:
```
thegreyd, line 588:
```suggestion
                if int(build_on_ticket["build_id"]) > int(build["build_id"]):
```
small suggestion

thegreyd:
1 suggestion, overall lgtm

doozer/doozerlib/cli/scan_osh.py
ashwindasr: When I checked the API ,it returned int. But makes sense for us to enforce. Updated in https://github.com/openshift-eng/art-tools/pull/297/commits/dd02866f7b3a87861b285c8b57652c2168ccc4f0

PR #300, opened 2024-01-09
[scan-osh] update regex for microshift nightly

ashwindasr:
The characters `~` and `_` need to be included since microshift nightly NVR contains it. Eg: `microshift-4.16.0~0.nightly_2024_01_05_154400-202401052128.p0.g0467b5a.assembly.microshift.el9`


joepvd (approved):
/lgtm

PR #308, opened 2024-01-11
exclude builds from attached to advisory

ashwindasr:


thegreyd:
This is stale, let's follow up with a new ticket & PR

PR #309, opened 2024-01-12
support bug close status with Obsolete

Ximinhan:
OCPBUGS-26416 status was closed with Obsolete, we should not block this with invalid status


joepvd (approved):
/lgtm

PR #313, opened 2024-01-12
skip verify-flaws for preview and prerelease

ashwindasr:
Skip `--verify-flaws` flag for elliott verify-attached-bugs for preview and prerelease

### DIFF
--- a/pyartcd/pyartcd/pipelines/promote.py
+++ b/pyartcd/pyartcd/pipelines/promote.py
@@ -255,8 +255,13 @@ class PromotePipeline:
                 #     logger.info("%s is GA'd. Blocking Bug check will be enforced.", next_minor)
                 logger.info("Verifying attached bugs...")
                 advisories = list(filter(lambda ad: ad > 0, impetus_advisories.values()))
+
+                verify_flaws = True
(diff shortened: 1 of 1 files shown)


pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -881,7 +886,8 @@ def verify_advisory_status(self, advisory_info: Dict):
         if advisory_info["status"] not in {"QE", "REL_PREP", "PUSH_READY", "IN_PUSH", "SHIPPED_LIVE"}:
             raise VerificationError(f"Advisory {advisory_info['id']} should not be in {advisory_info['status']} state.")
 
-    async def verify_attached_bugs(self, advisories: Iterable[int], no_verify_blocking_bugs: bool):
+    async def verify_attached_bugs(self, advisories: Iterable[int], no_verify_blocking_bugs: bool,
+                                   skip_verify_flaws: bool):
```
thegreyd, line 890:
```suggestion
                                  verify_flaws: bool = True):
```

thegreyd: Would recommend having a verify_flaws param that is set to true by default and then we set it to False when we deem it necessary

ashwindasr: done! repushed

PR #317, opened 2024-01-15
ART-8688 Enable mass rebuild queue

thegreyd:
https://issues.redhat.com/browse/ART-8688

Summary
- Store a `mass-rebuld-queue` in redis to store versions in queue. Pop it based on priority and then acquire mass rebuild  lock.
- Move to official redis client async implementation (we were using https://github.com/deepsentinel/aioredis which was a fork of https://github.com/aio-libs-abandoned/aioredis-py which moved in the official redis client). This is mainly because calling commands with params is different between these 2 and we are better off relying on official client lib.
- TODO: implement an initial wait time when lock isn't acquired

Test build
https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/sidsharm-aos-cd-jobs/job/build%252Focp4/33/console

### DIFF
--- a/pyartcd/pyartcd/locks.py
+++ b/pyartcd/pyartcd/locks.py
@@ -1,10 +1,13 @@
 import enum
 import logging
+import tenacity
 from types import coroutine
 
 from aioredlock import Aioredlock, LockError
 
 from artcommonlib import redis
+
+LOGGER = logging.getLogger('pyartcd')
 
 
 # Defines the pipeline locks managed by Redis
@@ -16,6 +19,10 @@ class Lock(enum.Enum):
     MASS_REBUILD = 'lock:mass-rebuild-serializer'
     SIGNING = 'lock:signing:{signing_env}'
     BUILD_SYNC = 'lock:build-sync:{version}'
+
+
+class Keys(enum.Enum):
+    MASS_REBUILD_QUEUE = 'appdata:mass-rebuild-queue'
 
 
 # This constant defines for each lock type:
@@ -141,6 +148,57 @@ class LockManager(Aioredlock):
 
         self.logger.info('Retrieving locks matching pattern "%s"', pattern)
         return await redis.get_keys(pattern)
+
+
+def mass_rebuild_score(ocp_version: str) -> int:
+    """Given an ocp_version (e.g. `4.16`) return an integer score value
+    Higher the score, higher the priority
+    """
+    return int(float(ocp_version) * 100)  # '4.16' -> 416
+
+
+async def run_with_mass_rebuild_lock(coro: coroutine, lock: Lock, lock_name: str, ocp_version: str, lock_id: str):
+    if lock != lock.MASS_REBUILD or lock_name != lock.MASS_REBUILD.value:
+        raise ValueError("expected lock to be of mass_rebuild type")
+
+    queue = Keys.MASS_REBUILD_QUEUE.value
+    score = mass_rebuild_score(ocp_version)
+    mapping = {ocp_version: score}
+
+    # add yourself to the queue
+    # if queue does not exist, it will be created with the value
+    await redis.call('zadd', queue, mapping)
+
+    LOGGER.info('Queued in for mass rebuild lock')
+
+    lock_manager = LockManager.from_lock(lock)
+
+    @tenacity.retry(
+        wait=tenacity.wait_fixed(600),  # wait for 10 minutes between retries
+        stop=tenacity.stop_after_attempt(600 * 24),  # wait for 24 hours
+        retry=tenacity.retry_if_exception_type(),
+    )
+    async def run_when_i_have_priority():
+        if not await lock_manager.is_locked(lock_name):
+            # TODO: use a redis tx here
+            # fetch the first element in the reverse sorted set by score (item with the max score)
+            result = await redis.call('zrange', queue, 0, 0, desc=True)
+            # there should be at least 1 entry, if not error out
+            version_on_top = result[0]
+            if version_on_top == ocp_version:
+                # remove yourself from the queue
+                items = await redis.call('zpopmax', queue)
+                assert items[0] == (ocp_version, score)
+                return await run_with_lock(coro, lock, lock_name, lock_id)
+
+        LOGGER.info(f"Do not have priority for {lock_name} Waiting ..")
+        raise tenacity.TryAgain()
+
+    # if we reach timeout, quit but don't remove yourself from the queue
+    # it will prevent other versions from acquiring the lock, until you get back in the queue
+    # for accidental mass rebuild requests, we will need to delete the keys manually for now
+
+    return await run_when_i_have_priority()
 
 
 async def run_with_lock(coro: coroutine, lock: Lock, lock_name: str, lock_id: str = None, skip_if_locked: bool = False):
--- a/pyartcd/pyartcd/pipelines/ocp4.py
+++ b/pyartcd/pyartcd/pipelines/ocp4.py
@@ -872,10 +872,11 @@ class Ocp4Pipeline:
             if self.mass_rebuild:
                 await self._slack_client.say(
                     f':loading-correct: Enqueuing mass rebuild for {self.version.stream} :loading-correct:')
-                await locks.run_with_lock(
+                await locks.run_with_mass_rebuild_lock(
                     coro=self._rebase_and_build_images(),
                     lock=Lock.MASS_REBUILD,
                     lock_name=Lock.MASS_REBUILD.value,
+                    ocp_version=self.version.stream,
                     lock_id=self.lock_identifier
                 )
             else:
--- a/artcommon/artcommonlib/redis.py
+++ b/artcommon/artcommonlib/redis.py
@@ -57,6 +57,16 @@ def handle_connection_sync(func):
 
 
 @handle_connection
+async def call(conn: redis.asyncio.client.Redis, func_name: str, *args, **kwargs):
+    """
+    Call any redis client function with opening and closing of connection auto handled
+    call('zadd', name, mapping) -> redis_connection.zadd(name, mapping)
+    """
+    func = getattr(conn, func_name)
+    return await func(*args, **kwargs)
+
+
+@handle_connection
 async def get_value(conn: redis.asyncio.client.Redis, key: str):
     """
     Returns value for a given key
@@ -101,7 +111,7 @@ def set_value_sync(conn: redis.client.Redis, key: str, value):
 @handle_connection
 async def get_keys(conn: redis.asyncio.client.Redis, pattern: str):
     """
-    Returns a list of keys (string) matching pattern (e.g. "*.count")
+    Returns a list of keys (string) matching pattern (e.g. "count*")
     """
 
     keys = await conn.keys(pattern)


locriandev (changes requested):

pyartcd/pyartcd/locks.py
```diff
@@ -18,6 +21,10 @@ class Lock(enum.Enum):
     BUILD_SYNC = 'lock:build-sync:{version}'
 
 
+class Keys(enum.Enum):
+    MASS_REBUILD_QUEUE = 'mass-rebuild-queue'
```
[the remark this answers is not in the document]
thegreyd: yes agreed :+1:

pyartcd/pyartcd/redis.py
```diff
@@ -4,12 +4,12 @@
...
 
-import aioredis
+import redis.asyncio as redis
 
 logger = logging.getLogger(__name__)
 
-# Redis instance template, to be renderes with env vars
-redis = Template('${protocol}://:${redis_password}@${redis_host}:${redis_port}')
+# Redis instance template, to be renders with env vars
+redis_url_temp = Template('${protocol}://:${redis_password}@${redis_host}:${redis_port}')
```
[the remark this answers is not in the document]
thegreyd: ha yeah I was going back and forth on this. Changed

pyartcd/pyartcd/locks.py
```diff
@@ -143,6 +150,46 @@ async def get_locks(self, version: str = None):
         return await redis.get_keys(pattern)
 
 
+async def request_mass_rebuild(coro: coroutine, lock: Lock, lock_name: str, ocp_version: str, lock_id: str):
```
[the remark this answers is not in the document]
thegreyd: Hm I don't know if it's worth abstracting this right now. We would assume to have a queue (which would be a sorted set), we would need to encode how we need to pop the value out of the set, we would need to pass in the score with the item - all of which is specific to each case. I like having this which is concise and specific to mass rebuild. If we have another usecase for queuing later, I would like us to do this then. Wdyt?

pyartcd/pyartcd/locks.py
```diff
@@ -143,6 +150,46 @@ async def get_locks(self, version: str = None):
...
+    queue = Keys.MASS_REBUILD_QUEUE.value
+    score = int(float(ocp_version) * 100)  # '4.16' -> 416
+    item = {ocp_version: score}
+    await redis.call('zadd', queue, item)
+
+    LOGGER.info('Queued in for mass rebuild lock')
+
+    lock_manager = LockManager.from_lock(lock)
+
+    @tenacity.retry(
```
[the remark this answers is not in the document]
thegreyd: Would like to keep this specific to mass rebuild / ocp4 right now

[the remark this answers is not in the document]
locriandev: I really don't like mixing up locks related stuff with ocp4 specifics. I am sure we can find a way to separate them out with little effort. Maybe `run_when_i_have_priority()` can be at locks module level, taking a LockManager instance as param, while `request_mass_rebuild()` can be at ocp4 module level?

pyartcd/pyartcd/locks.py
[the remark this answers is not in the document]
locriandev: Makes sense to me, as long as we manage to keep the separation of concerns (see next comment)

pyartcd/pyartcd/locks.py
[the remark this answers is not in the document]
thegreyd: Refactored now

pyartcd/pyartcd/locks.py
```diff
@@ -1,11 +1,14 @@
 import enum
 import logging
+import tenacity
 from types import coroutine
 
 from aioredlock import Aioredlock, LockError
 
 from artcommonlib import redis
 
+LOGGER = logging.getLogger('pyartcd')
```
locriandev, line 10: This is redundant, as we already have `lock_manager.logger` available
```suggestion
```

pyartcd/pyartcd/locks.py
```diff
@@ -143,6 +150,34 @@ async def get_locks(self, version: str = None):
...
+        # fetch the first element in the reverse sorted set by score (item with the max score)
+        result = await redis.call('zrange', version_queue_name, 0, 0, desc=True)
+        # there should be at least 1 entry, if not error out
+        version_on_top = result[0]
+        if version_on_top == ocp_version:
+            # remove yourself from the queue
+            await redis.call('zpopmax', version_queue_name)
+            return await run_with_lock(coro, lock, lock_name, lock_id)
+
+    LOGGER.info(f"Do not have priority for {lock_name} Waiting ..")
```
locriandev, line 177:
```suggestion
    lock_manager.logger.info(f"Do not have priority for {lock_name} Waiting ..")
```

pyartcd/pyartcd/pipelines/ocp4.py
```diff
@@ -842,6 +849,31 @@ def _report_success(self):
...
+
+        # add yourself to the queue
+        # if queue does not exist, it will be created with the value
+        await redis.call('zadd', queue, mapping)
+        self.runtime.logger.info('Queued in for mass rebuild lock')
+
+        # if we reach timeout, quit but don't remove yourself from the queue
+        # it will prevent other versions from acquiring the lock, until you get back in the queue
+        # for accidental mass rebuild requests, we will need to delete the keys manually for now
+        return await locks.enqueue_for_lock(coro, lock, lock_name, lock_id, ocp_version, queue)
```
locriandev, line 875:
```suggestion
        return await locks.enqueue_for_lock(
            coro=self._rebase_and_build_images(),
            lock=Lock.MASS_REBUILD,
            lock_name=Lock.MASS_REBUILD.value,
            lock_id=self.lock_identifier,
            ocp_version=self.version.stream,
            queue=locks.Keys.MASS_REBUILD_QUEUE.value
        )
```

pyartcd/pyartcd/pipelines/ocp4.py
```diff
@@ -842,6 +849,31 @@ def _report_success(self):
...
+        return round(float(self.version.stream) * 100)  # '4.16' -> 416
+
+    async def _request_mass_rebuild(self):
+        coro = self._rebase_and_build_images()
+        ocp_version = self.version.stream
+        lock_id = self.lock_identifier
+        lock = Lock.MASS_REBUILD
+        lock_name = Lock.MASS_REBUILD.value
+        queue = locks.Keys.MASS_REBUILD_QUEUE.value
+        mapping = {ocp_version: self._mass_rebuild_score()}
```
locriandev, line 865:
```suggestion
```

pyartcd/pyartcd/pipelines/ocp4.py
```diff
@@ -842,6 +849,31 @@ def _report_success(self):
...
+
+        # add yourself to the queue
+        # if queue does not exist, it will be created with the value
+        await redis.call('zadd', queue, mapping)
```
locriandev, line 869:
```suggestion
        await redis.call('zadd', locks.Keys.MASS_REBUILD_QUEUE.value, {self.version.stream: self._mass_rebuild_score()})
```

locriandev:
Overall LGTM, with minor bits. Great work!

pyartcd/pyartcd/locks.py
```diff
@@ -143,6 +150,34 @@ async def get_locks(self, version: str = None):
...
+)
+async def _enqueue_for_lock(coro: coroutine, lock_manager, lock: Lock, lock_name: str, lock_id: str,
+                            ocp_version: str, version_queue_name):
+    if not await lock_manager.is_locked(lock_name):
+        # TODO: use a redis tx here
+        # fetch the first element in the reverse sorted set by score (item with the max score)
+        result = await redis.call('zrange', version_queue_name, 0, 0, desc=True)
+        # there should be at least 1 entry, if not error out
+        version_on_top = result[0]
+        if version_on_top == ocp_version:
```
Ximinhan, line 172: this only contains ocp_version, what will happen if trigger two 4.13 jobs will it follow the job's order?

pyartcd/pyartcd/locks.py
```diff
@@ -143,6 +150,34 @@ async def get_locks(self, version: str = None):
         return await redis.get_keys(pattern)
 
 
+async def enqueue_for_lock(coro: coroutine, lock: Lock, lock_name: str, lock_id: str,
+                           ocp_version: str, version_queue_name):
```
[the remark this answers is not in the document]
thegreyd: Yes that is the key name is redis which stores the queue (sorted set in redis). No it is not supposed to have version in name, it is a sorted set (basically a dictionary) with `{ocp_version: str -> score: int}` . See it here https://redis-ui-redis.apps.artc2023.pc3z.p1.openshiftapps.com/

pyartcd/pyartcd/locks.py
thegreyd: Yeah, this was something of a design decision. Since this is a sorted set, the item that we insert is `{string: int/float}`. We can make the string `ocp_version#jobid` or similar and make sure the job that got in the queue is the job that gets popped. But that introduces us to situations where when the job reaches timeout (which will frequently happen since these are very long running jobs), we will need to pop it out of the queue in some way. But then we also lose our place in the queue. If between the interval of one ocp4 4.14 job reaching timeout popping itself, and then next instance of it inserting itself, the mass rebuild lock can be acquired by a 4.13 job still in the queue.

Rather than that, having generic not job specific ids, let us keep our place in queue.

pyartcd/pyartcd/pipelines/ocp4.py
thegreyd: I would like to keep this for readability

pyartcd/pyartcd/locks.py
```diff
@@ -143,6 +148,36 @@ async def get_locks(self, version: str = None):
         return await redis.get_keys(pattern)
 
 
+async def enqueue_for_lock(coro: coroutine, lock: Lock, lock_name: str, lock_id: str,
```
locriandev, line 151: Is there a reason for us to have this function that only serves as a wrapper to `_enqueue_for_lock`? Can't this be called directly?

pyartcd/pyartcd/locks.py
```diff
@@ -143,6 +148,36 @@ async def get_locks(self, version: str = None):
...
+                           ocp_version: str, version_queue_name):
+    lock_manager = LockManager.from_lock(lock)
+    return await _enqueue_for_lock(coro, lock_manager, lock, lock_name, lock_id, ocp_version, version_queue_name)
+
+
+@tenacity.retry(
+    wait=tenacity.wait_fixed(600),  # wait for 10 minutes between retries
+    stop=tenacity.stop_after_attempt(600 * 24),  # wait for 24 hours
```
locriandev, line 159: Is 24 hours long enough to account for times when all versions want to do mass rebuilds? What happens if the timeout is exceeded?

pyartcd/pyartcd/locks.py
thegreyd: This is to cache lock manager instance. I wanted to cache redis connection as well, but that will take some refactoring

joepvd:
Let's get it in:
/lgtm
/approve

pyartcd/pyartcd/locks.py
thegreyd: tenacity tryagain exception is raised
I think it's good, but we can tweak it in future

PR #318, opened 2024-01-16
Revert "Omit disabled repos from .oit/{signed,unsigned}.repo during rebase"

joepvd:
Reverts openshift-eng/art-tools#281


joepvd:
/hold
This will create mass rebuilds.

joepvd:
/close
let's flight forward rather than revert

PR #325, opened 2024-01-16
Migrate ocp-build-data-validator

ashwindasr:
Migrating validator from https://github.com/openshift-eng/ocp-build-data-validator to art-tools


vfreex (commented):
/lgtm

PR #326, opened 2024-01-17
Sync images always

joepvd:


joepvd:
/close
real fix will be part of https://issues.redhat.com/browse/ART-7129

PR #327, opened 2024-01-17
Fix GhApi owner param

ashwindasr:
For `pyartcd:WARNING No assembly PRs were found with head=thegreyd:art8712` issue

### DIFF
--- a/pyartcd/pyartcd/pipelines/build_sync.py
+++ b/pyartcd/pyartcd/pipelines/build_sync.py
@@ -24,8 +24,6 @@ from ghapi.all import GhApi
 
 TRACER = trace.get_tracer(__name__)
 GEN_PAYLOAD_ARTIFACTS_OUT_DIR = 'gen-payload-artifacts'
-GITHUB_OWNER = "openshift-eng"
-
 
 class BuildSyncPipeline:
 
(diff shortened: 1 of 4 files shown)


pyartcd/pyartcd/pipelines/build_sync.py
```diff
@@ -24,6 +24,7 @@
 
 TRACER = trace.get_tracer(__name__)
 GEN_PAYLOAD_ARTIFACTS_OUT_DIR = 'gen-payload-artifacts'
+GITHUB_OWNER = "openshift-eng"
```
locriandev, line 27: Already defined [here](https://github.com/openshift-eng/art-tools/blob/main/pyartcd/pyartcd/pipelines/build_microshift.py#L431). Can we move it to a common place? e.g. https://github.com/openshift-eng/art-tools/blob/main/pyartcd/pyartcd/constants.py

locriandev (changes requested):

pyartcd/pyartcd/pipelines/build_sync.py
ashwindasr: Thanks! Updated

locriandev (commented):
/lgtm

PR #329, opened 2024-01-18
Refactor redis module

locriandev:
Move it to artcommon
Use redis official repo instead of aioredis
Add sync functions to be used by synchronous code


artcommon/artcommonlib/redis.py
```diff
@@ -65,19 +88,40 @@ async def set_value(conn: aioredis.commands.Redis, key: str, value):
...
+    return keys
+
+
+@handle_connection_sync
+def get_keys_sync(conn: redis.client.Redis, pattern: str):
+    """
+    Same as get_keys(), but synchronous
+    """
+
+    keys = [key.decode() for key in conn.keys(pattern)]
```
thegreyd, line 118: Can remove the `decode()`
And if a key is not found, I think None will be returned? So maybe we return a dict?

thegreyd: Or list is fine as well, I was thinking of set first and then thought of dict. No strong opinion

locriandev: yeah I took that out after setting `decode_responses=True`

PR #334, opened 2024-01-19
Update images health message

ashwindasr:
Mention just the count on the main message and dump the details in the thread.

Using reference from [art attention](https://github.com/openshift-eng/aos-cd-jobs/blob/32ee80a778ec40e4044e133b89b2c09ad18b00f4/scheduled-jobs/scanning/unresolved-art-threads/unresolved-thread-notification.py#L123-L138)


joepvd (approved):
/lgtm

PR #350, opened 2024-01-25
Catch LockError in locks.unlock()

locriandev:


pyartcd/pyartcd/locks.py
```diff
@@ -124,9 +124,12 @@ async def lock(self, resource, *args, **kwargs):
...
     async def unlock(self, lock):
-        self.logger.info('Releasing lock "%s"', lock.resource)
-        await super().unlock(lock)
-        self.logger.info('Lock released')
+        try:
+            self.logger.info('Releasing lock "%s"', lock.resource)
+            await super().unlock(lock)
+            self.logger.info('Lock released')
+        except LockError:
+            self.logger.warning('Failed releasing lock %s', lock.resource)
```
thegreyd, line 132: Let's log the error message from the exception, and raise again?

locriandev: The LockError doesn't quite carry any useful information, I've tried in the past but seemed unrelevant

thegreyd: But why not raise again, when this happens, as it would previously

locriandev: Right, updated

PR #351, opened 2024-01-25
Extend lock timeouts

thegreyd:
We have possibility of silently releasing locks, which can end us up in
 very bad states in our build jobs. So extend them, until we implement the
"abort or extend on lock timeout" approach


locriandev (commented):
/lgtm

PR #353, opened 2024-01-26
[ART-8116] add hotfix and prerleease check for advisory

Ximinhan:
ref: https://issues.redhat.com/browse/ART-8116

enhance verify-attached-operators cli to add check advisory content
 * double check greenwave cvp test on the advisory
 * if the advisory is a hotfix advisory, bundle build need to have hotfix label
 * if the advisory is a prerelease advisory, image build need to have prerelease label

### DIFF
--- a/elliott/elliottlib/cli/verify_attached_operators_cli.py
+++ b/elliott/elliottlib/cli/verify_attached_operators_cli.py
@@ -114,7 +114,7 @@ def _analyze_image_builds(runtime, advisories: List, omit_shipped: bool, omit_at
                                       where they are needed
     """
     brew_session = koji.ClientSession(runtime.group_config.urls.brewhub or constants.BREW_HUB)
-    image_builds = _get_attached_image_builds(brew_session, advisories)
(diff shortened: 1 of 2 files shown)


pyartcd/pyartcd/util.py
```diff
@@ -49,6 +49,18 @@ def isolate_el_version_in_branch(branch_name: str) -> Optional[int]:
...
+    a OCP major.minor version. If it does, it returns the version value as (int, int).
+    If it is not found, (None, None) is returned.
+    """
+    match = re.fullmatch(r"openshift-(\d+).(\d+)", group_name)
+    if not match:
+        return None, None
+    return int(match[1]), int(match[2])
+
+
 def is_greenwave_all_pass_on_advisory(advisory_id: int) -> bool:
```
sosiouxme, line 64: since you've copied this to elliott, why not just use it from there?
```suggestion
from elliottlib.errata import is_greenwave_all_pass_on_advisory
```
or, remove it from here entirely and just do that import in `pyartcd/pyartcd/pipelines/prepare_release.py`

sosiouxme (changes requested):

PR #355, opened 2024-01-26
Allow pipeline user to trigger tekton pipelines

ashwindasr:
Should fix `Error: failed to list PipelineResources from namespace art-cd: pipelineresources.tekton.dev is forbidden: User "system:serviceaccount:art-cd:pipeline" cannot list resource "pipelineresources" in API group "tekton.dev" in the namespace "art-cd"
`


sosiouxme (approved):
/lgtm
we might find at some point that we should have created a separate serviceaccount but we can deal with that if so

PR #359, opened 2024-01-29
Do not overwrite scores in mass rebuild queue

thegreyd:
ref: https://redis-py.readthedocs.io/en/stable/commands.html#redis.commands.core.CoreCommands.zadd


joepvd (approved):
/lgtm

PR #361, opened 2024-01-29
[ART-6890] Allow stream aliases and multi-branch image builds

jupierce:
For [ART-6890](https://issues.redhat.com//browse/ART-6890), it is going to be useful to be able to:
- refer to streams from image metadata files via an unchanging alias.
- moves BrewBuildInspector out of doozerlib.images into its own package for easier use.
-  specify non-embargoed `repos` in the CI image metadatas. At present, this fails because there is no content_set value for the non-embargoed repos: https://github.com/openshift-eng/ocp-build-data/blob/2554dcb3b80d6deccd212613b1f7733a6aea4731/group.yml#L408 .  This change allows non-embargoed repos to be used to generate repo files during rebase.
- two different image metadatas (e.g. `ci-openshift-base.yml` and `ci-openshift-base.rhel8.yml`) can now refer to the same distgit repository but build on different branches. The release field for images has been updated to include `elX` so that tags in different branches will not conflict and that finding the latest image can distinguish between images built for different RHEL versions.
- add simple caching for `oc image info` invocations.

Example streams entry using aliases:
```yaml
rhel-9-golang:
  aliases:
  - rhel-9-golang-{GO_LATEST}
  - rhel-9-golang-{GO_EXTRA}
  image: openshift/golang-builder:v1.21.3-202401221732.el9.g00c615b
  mirror: true
  transform: rhel-9/golang
...
```


vfreex (commented):
/lgtm

doozer/doozerlib/distgit.py
```diff
@@ -411,6 +412,13 @@ async def push_async(self):
         timeout = str(self.runtime.global_opts['rhpkg_push_timeout'])
         await exectools.cmd_assert_async(["timeout", f"{timeout}", "git", "push", "--follow-tags"], cwd=self.distgit_dir, retries=3)
 
+    def get_branch_el(self) -> Optional[int]:
```
locriandev, line 415: This replaced `isolate_rhel_major_from_version()` that is no longer used and can be removed

doozer/doozerlib/distgit.py
```diff
@@ -1672,48 +1679,31 @@ def _determine_upstream_rhel_version(self, source_path) -> Optional[int]:
...
-        version = redis.get_value_sync(redis_key)
-        if version:
-            # Already tracked in Redis
-            return int(version)
 
-        # Not tracked yet: let's read it from os-release
+        version = None
         try:
-            # Create a temp dir with a random name to store the os-release file
-            with TemporaryDirectory() as tmpdir:
```
locriandev, line 1686: `TemporaryDirectory` imported symbol is no longer used

doozer/doozerlib/distgit.py
```diff
@@ -1672,48 +1679,31 @@ def _determine_upstream_rhel_version(self, source_path) -> Optional[int]:
...
-
-                # Parse it
-                with open(f'{tmpdir}/os-release') as os_release:
-                    # os-release will contain a line like this: VERSION_ID="9.2"
-                    # We need to eval the result to turn '"9.2"' into '9.2'
-                    version_id = eval([line for line in os_release if 'VERSION_ID' in line][0].split('=')[1])
-                    version = isolate_rhel_major_from_version(version_id)
-                    self.logger.info('Intended upstream RHEL version for %s: %s', self.name, version)
+            # We will infer the rhel version from the last build layer in the upstream Dockerfile
+            last_layer = dfp.parent_images[-1]
```
locriandev, line 1686: Already done at line [1681](https://github.com/openshift-eng/art-tools/pull/361/files#diff-d6994e859542b479607bede445901c422a82513def8423a189876ef9ca601b6cL1674)

doozer/doozerlib/distgit.py
```diff
@@ -1672,48 +1679,31 @@ def _determine_upstream_rhel_version(self, source_path) -> Optional[int]:
...
-            #   - an IndexError when /etc/os-release is not as expected
+        except Exception as e:
+            # Swallow exception as this is not fatal. We just can't determine the
+            # correct canonical upstream RHEL version to use.
             self.logger.warning('Failed determining upstream rhel version: %s', e)
 
         if not version:
             self.logger.warning('Could not determine rhel version from upstream %s', self.name)
 
-        try:
```
locriandev, line 1712: If something goes wrong while retrieving `version` casting None to int will throw an unhandled TypeError causing a rebase error

locriandev (changes requested):

doozer/doozerlib/distgit.py
jupierce: Removed the import.

doozer/doozerlib/distgit.py
jupierce: Removed the import.

doozer/doozerlib/distgit.py
jupierce: Good catch. Late night merge conflict resolution should be avoided :) .

doozer/doozerlib/distgit.py
jupierce: True, but I don't think there is any unsafe casting in the the BrewBuildImageInspector version. `.get_rhel_base_version()` directly returns an `int` or `None`.

jupierce:
@locriandev 
> this PR was probably conceived before https://github.com/openshift-eng/art-tools/pull/331 merged, with which we started storing the upstream RHEL version in Redis. If it's there already we'd avoid calling oc image info and Brew API, which most likely takes more time than retrieving a key from redis. Can we consider keep that part in place?

In the last commit, I added a simple, in-memory cache with `util.oc_image_info__cached`. Given the small number of upstream images, I think this will perform admirably, help users that don't have redis creds installed, and help with execution times across doozer by reducing redundant `info` gathering calls (of which I think there were many offenders).

doozer/doozerlib/distgit.py
locriandev: I think `isolate_rhel_major_from_version()` can be wiped out, nothing is using it as this PR is using `isolate_el_version_in_brew_tag()` instead

locriandev:
> @locriandev
> 
> > this PR was probably conceived before #331 merged, with which we started storing the upstream RHEL version in Redis. If it's there already we'd avoid calling oc image info and Brew API, which most likely takes more time than retrieving a key from redis. Can we consider keep that part in place?
> 
> In the last commit, I added a simple, in-memory cache with `util.oc_image_info__cached`. Given the small number of upstream images, I think this will perform admirably, help users that don't have redis creds installed, and help with execution times across doozer by reducing redundant `info` gathering calls (of which I think there were many offenders).

@jupierce while in-memory caching works within the same Doozer invocation, what I wanted to achieve with Redis is a permanent storage of any pullspec-rhel version pair. This would let distinct Doozer invocations read from the same cache, saving time in future runs. If the only issue you're seeing is the need to make these credentials available to the users, I was already thinking about making them optional: if they're not defined in the execution environment, we can simply fall backback to using `oc` (and make good use of the in-memory caching). `buildvm` will always have them.

OTOH, if you think that for a given pullspec we might see changes in the RHEL version, the Redis approach could be dangerous unless we set an expiration date on all cached  values. But is this the case? e.g. for `openshift-enterprise-base-rhel9` we know el9 is the rhel version, once resolved and stored we can pretty much forget about it

jupierce:
We talked in slack about the risks of using redis. Removing for now and removing do-not-merge.

locriandev (approved):

PR #363, opened 2024-01-30
log severity change

mbiarnes:
when in build-sync logs, after scanning all images and doing checks the output is
permitted: false and so viable: false the job will fail and it is  misleading to have a warning here.
**WARNING Assembly is not permitted. Will not apply changes to imagestreams.**


vfreex (commented):
/lgtm
/approve

PR #364, opened 2024-01-30
Use canonical builders when rhel versions match

locriandev:
If there is no matching "when" clause but ART and upstream rhel versions match, apply canonical builders

### DIFF
--- a/artcommon/artcommonlib/util.py
+++ b/artcommon/artcommonlib/util.py
@@ -151,6 +151,17 @@ def isolate_rhel_major_from_version(version: str) -> Optional[int]:
     return None
 
 
+def isolate_rhel_major_from_distgit_branch(branch: str) -> Optional[int]:
+    """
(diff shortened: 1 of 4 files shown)


artcommon/artcommonlib/util.py
```diff
@@ -151,6 +151,17 @@ def isolate_rhel_major_from_version(version: str) -> Optional[int]:
     return None
 
 
+def isolate_rhel_major_from_distgit_branch(branch: str) -> Optional[int]:
+    """
+    E.g. 'rhaos-4.16-rhel-9' => 9
+    """
+
+    match = re.fullmatch(r"^rhaos-(\d+).(\d+)-rhel-(\d+)", branch)
```
joepvd, line 159: Please escape the `.`. Maybe also do not put capture groups around major and minor. as this match is irrelevant. 

`"^rhaos-\d+\.\d+-rhel-(\d+)`

PR #366, opened 2024-01-30
Set default lock timeout to a BIG value

thegreyd:
In our current locks when lock timeout is reached they silently expire, no exception is raised. This is an expected pattern in redis, the lib we use implements https://redis.io/docs/manual/patterns/distributed-locks/ and sets keys with an expiration time.

But it's unexpected for us. We've had a pattern of using BIG job-wide locks, since we started using locks in jenkins. From what I remember there was no expiry time in jenkins locks, so as long as the `lock() {do something}` ran, it kept the lock active.

We're in the middle of these 2 approaches, and bad things can happen without us noticing. 
Therefore set lock timeouts to a BIG value, effectively not relying on them (until we change our lock use pattern)

### DIFF
--- a/pyartcd/pyartcd/locks.py
+++ b/pyartcd/pyartcd/locks.py
@@ -23,6 +23,12 @@ class Keys(enum.Enum):
     MASS_REBUILD_QUEUE = 'appdata:mass-rebuild-queue'
 
 
+# Use a BIG timeout value so that locks do not silently expire.
+# Effectively we do not want to rely on lock timeout being reached
+# since we usually start a long-running job/operation & after acquiring lock do not
+# check its validity again (if this pattern changes then we should change this).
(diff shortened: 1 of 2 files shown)


pyartcd/pyartcd/locks.py
```diff
@@ -195,10 +205,5 @@ async def run_with_lock(coro: coroutine, lock: Lock, lock_name: str, lock_id: st
 
         async with await lock_manager.lock(resource=lock_name, lock_identifier=lock_id):
             return await coro
-
-    except LockError as e:
-        lock_manager.logger.error('Failed acquiring lock %s: %s', lock_name, e)
```
thegreyd, line 200: This is being logged when unlock() raises an exception, so `Failed acquiring lock` isn't accurate. Move logging error to the lock method.

locriandev (approved):
/lgtm

PR #368, opened 2024-01-31
[ART-8800] update alignment PR wording

sosiouxme:
... because most of these are proposed changes, not just alignment with what ART already does.

### DIFF
--- a/doozer/doozerlib/cli/images_streams.py
+++ b/doozer/doozerlib/cli/images_streams.py
@@ -714,8 +714,8 @@ def reconcile_jira_issues(runtime, pr_map: Dict[str, PullRequest.PullRequest], d
     for distgit_key, pr in pr_map.items():
         image_meta: ImageMetadata = runtime.image_map[distgit_key]
         potential_project, potential_component = image_meta.get_jira_info()
-        summary = f"Update {release_version} {image_meta.get_component_name()} image to be consistent with ART"
-        old_summary_format = f"Update {release_version} {image_meta.name} image to be consistent with ART"
+        summary = f"ART requests updates to {release_version} image {image_meta.get_component_name()}"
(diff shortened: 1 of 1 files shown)


doozer/doozerlib/cli/images_streams.py
```diff
@@ -714,8 +714,8 @@ def reconcile_jira_issues(runtime, pr_map: Dict[str, PullRequest.PullRequest], d
     for distgit_key, pr in pr_map.items():
         image_meta: ImageMetadata = runtime.image_map[distgit_key]
         potential_project, potential_component = image_meta.get_jira_info()
-        summary = f"Update {release_version} {image_meta.get_component_name()} image to be consistent with ART"
-        old_summary_format = f"Update {release_version} {image_meta.name} image to be consistent with ART"
+        summary = f"ART requests updates to {release_version} image {image_meta.get_component_name()}"
```
thegreyd, line 717: I think we might be relying on this wording in some doozer commands (images:streams), so changing this could break them

doozer/doozerlib/cli/images_streams.py
```diff
@@ -1267,9 +1274,11 @@ def check_if_upstream_image_exists(upstream_image):
   introduced with a separate PR opened by the component team. Once the repository is aligned,
   this PR will be closed automatically.
 - Patch-manager or those with sufficient privileges within this repository may add
-  any required labels to ensure the PR merges once tests are passing. Downstream builds
-  are *already* being built with these changes. Merging this PR only improves the fidelity
-  of our CI.
+  any required labels to ensure the PR merges once tests are passing. In cases where ART config is
+  canonical, downstream builds are *already* being built with these changes, and merging this PR
```
thegreyd, line 1278: Should we mention the expected behavior of our canonical feature? 
preGA: usually canonical 
postGA: not canonical

PR #370, opened 2024-02-01
Fix rebase with a deep_merge() function

locriandev:
`util.merge_objects()` is failing when we try and merge configs during rebase, by throwing this Exception:
```
AttributeError: 'dict' object has no attribute 'move_to_end'
```

This could be solved by casting the object into an OrderedDict, still it has problems when overriding existing keys: it will not preseve tested leaves. More details in [this thread](https://redhat-internal.slack.com/archives/GDBRP5YJH/p1706796879603389?thread_ts=1706774096.594939&cid=GDBRP5YJH)

### DIFF
--- a/artcommon/artcommonlib/util.py
+++ b/artcommon/artcommonlib/util.py
@@ -4,6 +4,8 @@ from artcommonlib.constants import RELEASE_SCHEDULES
 import requests
 import re
 
+from artcommonlib.model import Model
+
 
 def remove_prefix(s: str, prefix: str) -> str:
     if s.startswith(prefix):
@@ -170,3 +172,24 @@ def get_ocp_version_from_group(group):
     if not match:
         raise ValueError(f"Invalid group name: {group}")
     return int(match[1]), int(match[2])
+
+
+def deep_merge(dict1, dict2):
+    """
+    Recursively merge two dictionaries.
+
+    Returns:
+    A new dictionary with merged values.
+    """
+
+    merged = dict1.copy()
+
+    for key, value in dict2.items():
+        if isinstance(merged.get(key), dict) and isinstance(value, dict):
+            # If both values are dictionaries, merge them recursively
+            merged[key] = deep_merge(merged[key], value)
+        else:
+            # Otherwise, simply update the value
+            merged[key] = value
+
+    return merged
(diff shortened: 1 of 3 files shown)


artcommon/artcommonlib/util.py
```diff
@@ -170,3 +172,25 @@ def get_ocp_version_from_group(group):
...
+    Returns:
+    A new dictionary with merged values.
+    """
+
+    merged = Model(dict1.copy())
+    model2 = Model(dict2)
+
+    for key, value in dict2.items():
+        if isinstance(merged.get(key), dict) and isinstance(value, dict):
+            # If the key is in preserve_keys, and both values are dictionaries, merge them recursively
```
sosiouxme, line 190: comment is confusing

artcommon/artcommonlib/util.py
```diff
@@ -170,3 +172,25 @@ def get_ocp_version_from_group(group):
...
+            merged[key] = deep_merge(merged[key], value)
+        else:
+            # Otherwise, simply update the value
+            merged[key] = value
+
+    return Model(merged)
```
sosiouxme, line 196: it being a Model doesn't seem important to the function, and the callers re-create it as a Model anyway (which does a deep copy), so why bother here?
```suggestion
    merged = dict1.copy()

    for key, value in dict2.items():
        if isinstance(merged.get(key), dict) and isinstance(value, dict):
            # If both values are dictionaries, merge them recursively
            merged[key] = deep_merge(merged[key], value)
        else:
            # Otherwise, simply update the value
            merged[key] = value

    return merged
```

doozer/doozerlib/distgit.py
```diff
@@ -1722,12 +1722,12 @@ def _update_image_config(self):
         """
 
         # Check if there is an alternative configuration matching upstream RHEL version
-        alt_configs = self.config.get('alternative_upstream', [])
+        alt_configs = self.config.alternative_upstream
         matched = False
         for alt_config in alt_configs:
```
sosiouxme, line 1727: this will break if it's not there, can't iterate MissingModel
```suggestion
        for alt_config in alt_configs or []:
```

sosiouxme (changes requested):
otherwise, looks good

sosiouxme (approved):
/lgtm

PR #376, opened 2024-02-05
report nothing when list of inconsistencies is empty

mbiarnes:
in case the inconsistencies are only about  AssemblyIssueCode.OUTDATED_RPMS_IN_STREAM_BUILD the msgs will be empty


doozer/doozerlib/cli/release_gen_payload.py
```diff
@@ -1687,6 +1687,10 @@ def build_inconsistency_annotations(inconsistencies: Iterable[AssemblyIssue]) ->
         exclude_assembly_issues = [AssemblyIssueCode.OUTDATED_RPMS_IN_STREAM_BUILD]
 
         msgs = sorted([i.msg for i in inconsistencies if i.code not in exclude_assembly_issues])
+
+        if not msgs:
+            return
```
locriandev, line 1692: I think this function should return a dict, which is expected from the caller. Returning None could cause trouble. See line #1684

locriandev (changes requested):

locriandev (commented):
/lgtm

PR #381, opened 2024-02-06
Remove go.real symlink

vfreex:
This symlink prevents ArgoCD from working with art-tools repo.


vfreex:
/hold
@jupierce Could you check if the symlink is still in use?

PR #382, opened 2024-02-06
Allow name attribute in distgit

joepvd:
To support https://github.com/openshift-eng/ocp-build-data/pull/4312


locriandev (commented):
/lgtm

PR #390, opened 2024-02-08
fix date format

Ximinhan:
fix the error of https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fgen-assembly/802/console
we convert the date format for advisory usage, need to convert back when doing date calculation


joepvd (approved):
/lgtm

PR #392, opened 2024-02-08
[ART-8905] Fixes for OCP 3.11 pipeline

thiagoalessio:
Test run: https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/talessio-aos-cd-jobs/job/build%252Focp3/13/console

### DIFF
--- a/doozer/doozerlib/cli/config_plashet.py
+++ b/doozer/doozerlib/cli/config_plashet.py
@@ -252,8 +252,11 @@ def get_brewroot_arch_base_path(config, nvre, signed):
     package_version = parsed_nvr["version"]
     package_release = parsed_nvr["release"]
 
+    # Default to rhel 7 if no version is found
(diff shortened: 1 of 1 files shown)


doozer/doozerlib/cli/config_plashet.py
```diff
@@ -252,8 +252,11 @@ def get_brewroot_arch_base_path(config, nvre, signed):
     package_version = parsed_nvr["version"]
     package_release = parsed_nvr["release"]
 
+    # Default to rhel 7 if no version is found
+    el_version = isolate_el_version_in_release(package_name) or 7
```
locriandev, line 256: I don't think this should make it into our production code, as there might be situations where we expect None instead. This hack appears to be needed only by ocp3, so I'd prefer to keep it in a hack branch

locriandev (commented):
/hold

PR #394, opened 2024-02-09
Use .el<version> suffix to find Brew builds

locriandev:
Follows changes in https://github.com/openshift-eng/art-tools/pull/361

Fixes error like
```
OSError: No builds detected for using prefix: 'ci-openshift-golang-builder-latest-container-v4.16.', extra_pattern: '*', assembly: 'stream', build_state: 'COMPLETE', el_target: 'None'
```
that is caused by, for example, `.el8` build being the latest one, and we're looking for el9 builds. We'll filter out the build because of [this check](https://github.com/openshift-eng/art-tools/blob/main/doozer/doozerlib/metadata.py#L483)


joepvd (approved):
/lgtm

PR #398, opened 2024-02-12
build-sync: fix runtime initialization

locriandev:
After the introduction of `.el<version>` suffixes in Brew builds, we are getting empty builds lists after [filtering them out](https://github.com/openshift-eng/art-tools/blob/main/doozer/doozerlib/metadata.py#L485).

When doing so, `branch_el_target()` reads from the image metadata, which must be updated to use the `alternative_upstream` stanza when using canonical builders. This el target must match the one indicated by the build suffix. If we don't consider the alternative_upstream config, we'll get el9 branches (read from the image main config) for .el8 builds, and we'll filter these builds out thus not finding any.

Unfortunately, to do so we need to clone distgits and do the upstream intended rhel version check.


thiagoalessio (approved):
/lgtm

PR #400, opened 2024-02-13
Update validator

locriandev:
Current schema allows `ci_alignment.enabled` property, which is misleading as setting it to False will not prevent CI alignment PRs to be created.

Follows https://github.com/openshift-eng/ocp-build-data/pull/4356

### DIFF
--- a/ocp-build-data-validator/validator/json_schemas/image_content.base.schema.json
+++ b/ocp-build-data-validator/validator/json_schemas/image_content.base.schema.json
@@ -21,7 +21,6 @@
         "ci_alignment": {
           "type": "object",
           "properties": {
-            "enabled-": {},
             "final_user": {
               "description": "Parameter for the transform Dockerfile to set this user when complete",
               "oneOf": [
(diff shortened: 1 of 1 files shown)


ocp-build-data-validator/validator/json_schemas/image_content.base.schema.json
```diff
@@ -21,16 +21,6 @@
...
-              "description": "Whether to enable CI alignment. Default value: true",
-              "type": "boolean"
-            },
-            "enabled?": {
-              "$ref": "#/properties/source/properties/ci_alignment/properties/enabled"
-            },
-            "enabled!": {
-              "$ref": "#/properties/source/properties/ci_alignment/properties/enabled"
-            },
             "enabled-": {},
```
sosiouxme, line 24: nuke this too...

sosiouxme:
```suggestion
```

PR #410, opened 2024-02-16
[ART-8970] Mirror all images to private quay

ashwindasr:
As the first step for filtering out embargoed builds.

### DIFF
--- a/doozer/doozerlib/cli/release_gen_payload.py
+++ b/doozer/doozerlib/cli/release_gen_payload.py
@@ -20,6 +20,7 @@ from artcommonlib import rhcos
 from artcommonlib.rhcos import RhcosMissingContainerException
 import openshift_client as oc
 from opentelemetry import trace
+from enum import Enum
 
 
 from doozerlib.rpm_utils import parse_nvr
@@ -43,6 +44,11 @@ from doozerlib.util import find_manifest_list_sha
 TRACER = trace.get_tracer(__name__)
 
 
+class RepositoryType(Enum):
+    PRIVATE = 0
+    PUBLIC = 1
+
+
 @cli.command("release:gen-payload", short_help="Mirror release images to quay and release-controller")
 @click.option("--is-name", metavar="NAME", required=False,
               help="ImageStream .metadata.name value. For example '4.2-art-latest'")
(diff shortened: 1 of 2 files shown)


doozer/doozerlib/cli/release_gen_payload.py
```diff
@@ -52,6 +58,8 @@
               help="Quay ORGANIZATION to mirror into.\ndefault=openshift-release-dev")
 @click.option("--repository", metavar="REPO", required=False, default="ocp-v4.0-art-dev",
               help="Quay REPOSITORY in ORGANIZATION to mirror into.\ndefault=ocp-v4.0-art-dev")
+@click.option("--private-repository", metavar="REPO", required=False, default="ocp-v4.0-art-dev-priv",
+              help="Private Quay REPOSITORY in ORGANIZATION to mirror into.\ndefault=ocp-v4.0-art-dev-priv")
```
Ximinhan, line 62: I think the repo name ocp-v4.0-art-dev-priv won't change so no need to be a parameter, can be a constant value or a hardcode

doozer/doozerlib/cli/release_gen_payload.py
```diff
@@ -592,12 +605,15 @@ def generate_payload_entries(self, assembly_inspector: AssemblyInspector) -> Dic
 
             entries: Dict[str, PayloadEntry]  # Key of this dict is release payload tag name
             issues: List[AssemblyIssue]
-            entries, issues = PayloadGenerator.find_payload_entries(assembly_inspector, arch,
-                                                                    self.full_component_repo())
+            entries, issues = PayloadGenerator.find_payload_entries(assembly_inspector, arch, self.full_component_repo(repo_type=RepositoryType.PUBLIC))
+
+            # Exclude issues for now, for private repo
+            private_entries, _ = PayloadGenerator.find_payload_entries(assembly_inspector, arch, self.full_component_repo(repo_type=RepositoryType.PUBLIC))
```
Ximinhan, line 611: the repo_type should be private?

doozer/doozerlib/cli/release_gen_payload.py
ashwindasr: I followed the same pattern for the public quay. So I think we can raise a follow up PR if we want to convert them to constants.

doozer/doozerlib/cli/release_gen_payload.py
ashwindasr: You're right! Thanks for the catch! Fixed in https://github.com/openshift-eng/art-tools/pull/410/commits/2cfe5c0b8cd05446df56a2e4420c85b80d61d53f

sosiouxme:
/lgtm
i think this should be OK to test in a hackspace too

ashwindasr:
/hold

Until buildvm oc mirror issue is sorted

sosiouxme:
updated it so the sync files show up in artifacts with public/private names instead of reusing filenames and thus overwriting. [test run](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fbuild-sync/42221/console) with [artifacts](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fbuild-sync/42221/artifact/gen-payload-artifacts/)
/lgtm
this does double the amount of syncing build-sync has to do, but hopefully that's acceptable.
/hold cancel

PR #413, opened 2024-02-19
Use multiple search patterns when looking for builds

locriandev:
Fixes https://github.com/openshift-eng/art-tools/pull/402. Using `branch_el_target()` go compose the search pattern lead to an error, as this is drawn from group.yml and can differ from the actual rhel target we're interested in. So for example in 4.14 it was always finding el8 builds for `openshift-clients`, that in turn has both rhel 8 and 9 build targets.


locriandev (commented):
/approve

PR #414, opened 2024-02-20
ART-8763 create batch when prepare release

Ximinhan:
add create_batch function and update_advisory_function with batch API
During prepare release create a batch before creating advisory and link the advisory to the batch id after advisory created
During drop advisory clear the batch of the advisory before drop advisory

the batch will be created like https://errata.devel.redhat.com/batches/2828


elliott/elliottlib/errata.py
```diff
@@ -764,6 +764,40 @@ def get_file_meta(advisory_id) -> List[dict]:
     return ErrataConnector()._get(f'/api/v1/erratum/{advisory_id}/filemeta')
 
 
+def create_batch(release_version, release_date):
+    """Create batch for a release
+    https://errata.devel.redhat.com/documentation/developer-guide/api-http-api.html#batches
+    :param release_name: release name, e.g. 4.14.12
```
joepvd, line 770: doc does not match arg

elliott/elliottlib/errata.py
```diff
@@ -764,6 +764,40 @@ def get_file_meta(advisory_id) -> List[dict]:
...
+
+    """
+    data = {
+        "name": f"OCP {release_version}",
+        "release_name": "RHOSE ASYNC - AUTO",
+        "release_date": release_date,
+        "description": f"OCP {release_version}",
+        "is_active": True
+    }
+    response = ErrataConnector()._post("/api/v1/batches", data=data)
```
joepvd, line 782: error checking and retries would be great

pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -151,6 +153,10 @@ async def run(self):
             advisory_type = "RHEA" if is_ga else "RHBA"
             for ad in advisories:
                 if advisories[ad] < 0:
+                    if not batch_id:
```
joepvd, line 156: How do we want to do pre-releases? 
Think this logic works for GA releases, but if we are trying to ship an operator-prerelease, then not.

elliott/elliottlib/errata.py
```diff
@@ -764,6 +764,40 @@ def get_file_meta(advisory_id) -> List[dict]:
...
+    """Change the batch details for an advisory.
+    POST /api/v1/erratum/{id}/change_batch
+    Request body may contain:
+        batch_id - id of the batch (integer)
+        batch_name - name of the batch (string) - alternative to batch_id
+        clear_batch - removes erratum from batch (boolean)
+        is_batch_blocker - indicates if advisory blocks batch (boolean)
+    """
+    if clear_batch:
+        ErrataConnector()._post(f'/api/v1/erratum/{advisory_id}/change_batch', data={"clear_batch": True})
```
joepvd, line 796: error handling / retries

joepvd (dismissed):

elliott/elliottlib/errata.py
Ximinhan: added

elliott/elliottlib/errata.py
Ximinhan: updated

PR #424, opened 2024-02-26
Simplify runtime init logic

Ximinhan:
move canonical_builders_from_upstream check into runtime init function


Ximinhan:
need to check the canonical_builders_from_upstream value type, because in 4.12 its no, 4.13 its auto, 4.14 its off, 4.15 its false

PR #435, opened 2024-02-29
[ART-8910] Always inject doozer metadata environment variables on rebase

jupierce:
Always inject metadata environment variables into Dockerfiles even when set_build_variables is set to false. This changes ensures that metadata config `envs` are written into Dockerfiles as well as some standard metadata: `__doozer_group`, `__doozer_key`, and `__doozer_version`.

Ultimately, this is to enable work for [ART-8910](https://issues.redhat.com//browse/ART-8910) which requires the ability to set environment variables without using `set_build_variables` .


vfreex (commented):
/lgtm
/approve

PR #422, opened 2024-02-22
Honor canonical builder in elliott find-builds

thegreyd:
Follow up to https://github.com/openshift-eng/art-tools/pull/419

PR #436, opened 2024-02-29
Revert "Honor canonical builder in elliott find-builds"

vfreex:
Reverts openshift-eng/art-tools#422

@locriandev elliott never clones distgit. Does canonical builder feature affect how elliott finds builds?

### DIFF
--- a/doozer/doozerlib/runtime.py
+++ b/doozer/doozerlib/runtime.py
@@ -368,8 +368,7 @@ class Runtime(GroupRuntime):
         if disabled is not None:
             self.load_disabled = disabled
 
-        if not self.logger:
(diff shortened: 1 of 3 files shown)


locriandev:
> Reverts #422
> 
> @locriandev elliott never clones distgit. Does canonical builder feature affect how elliott finds builds?

@vfreex yes it does. We tried digging, and it turned out to be very messy for how the canonical builder feature is currently implemented. In 4.16, elliott was attaching old builds. To fix this, I've run prepare-release from a hacked art-tools: https://github.com/openshift-eng/art-tools/compare/main...locriandev:hack-find-builds

PR #438, opened 2024-02-29
Fix elliott

joepvd:


elliott/elliottlib/cli/create_cli.py
```diff
@@ -1,6 +1,7 @@
 import click
 import datetime
 import elliottlib
```
locriandev, line 3: Think this is unused now and can be removed

PR #440, opened 2024-02-29
Fix elliott find-builds

locriandev:
For pre-releases, canonical builders will be enabled and `alternative_upstream` configs will be defined for many components. This feature requires distgits to be cloned (discussion [here](https://redhat-internal.slack.com/archives/C06625LHXQS/p1708975115881139?thread_ts=1708610009.547359&cid=C06625LHXQS)) in order to take the alternative_upstream into consideration, but Elliott knows nothing about them and it is not trivial to teach him.

As an alternative, we can hack `find-builds` like this: if canonical builders are enabled and `alternative_upstream` is defined for an image, it most likely means that upstream is behind ART's main config with respect to the rhel target. Therefore, when looking for builds, Elliott can consider a downgraded (-1) rhel target to find, for example, `.el8` builds instead of the el9 it normally would.

Tested with:
```
$ elliott --group=openshift-4.16 --assembly=stream --rpms=openshift-clients find-builds --kind=rpm --member-only
openshift-clients-4.16.0-202402272209.p0.g91ba77a.assembly.stream.el8
openshift-clients-4.16.0-202402272209.p0.g91ba77a.assembly.stream.el9

$ elliott --group=openshift-4.16 --assembly=stream --images=azure-file-csi-driver-operator find-builds --kind=image --payload
ose-azure-file-csi-driver-operator-container-v4.16.0-202402281710.p0.g74590de.assembly.stream.el9

$ elliott --group=openshift-4.16 --assembly=stream --images=csi-livenessprobe find-builds --kind=image --payload
csi-livenessprobe-container-v4.16.0-202403010139.p0.ga5a99fc.assembly.stream.el8
```

### DIFF
--- a/artcommon/artcommonlib/build_util.py
+++ b/artcommon/artcommonlib/build_util.py
@@ -1,6 +1,12 @@
+from datetime import datetime
 from typing import List, Optional, Iterable, Dict
 
+from artcommonlib import logutil
+from artcommonlib.model import Missing
 from artcommonlib.release_util import isolate_assembly_in_release
+from doozerlib.release_schedule import ReleaseSchedule
+
+LOGGER = logutil.get_logger(__name__)
 
 
 def find_latest_build(builds: List[Dict], assembly: Optional[str]) -> Optional[Dict]:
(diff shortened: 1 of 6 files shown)


joepvd:
/lgtm
/approve

This is better than we have. Let's go with it.

PR #448, opened 2024-03-05
Match upstream builders by finding a suitable entry in streams.yml

locriandev:
Ref. [ART-7129](https://issues.redhat.com/browse/ART-7129)

Currently we are trusting upstream with whatever `FROM` clause they use. If, for example, a 4.16 component has 
```
FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.21-openshift-4.15
```
we would use `oc` to infer the exact pullspec for a 4.15 builder, and use that in the downstream Dockerfile. Even with matching {MAJOR}.{MINOR} builders, we would simply get the latest pullspec associated with that floating tag, which could lead to unexpected behavior.

This PR changes the way we rebase upstream desired builders: for a given CI pullspec, we determine the builder X.Y and the RHEL version, and look for a matching entry in `streams.yml`. For example:

- `rhel-8-golang-1.21-openshift-4.15` for a 4.16 image would match the `golang` stream, that points to `openshift/golang-builder:v1.21.3-202311221709.el8.gf79bf9c`. 
- `rhel-8-golang-1.19-openshift-4.6` for a 4.16 image would match the `etcd_golang` stream, that points to `openshift/golang-builder:v1.19.13-202310161907.el8.g0d095f7`
-  `rhel-8-golang-1.18-openshift-4.16` for a 4.16 image wouldn't match any entry in `streams.yml`, and we would override the build stage with ART configured builder.

This way, we will make sure only ART selected builder NVRs shall be used

### DIFF
--- a/doozer/doozerlib/distgit.py
+++ b/doozer/doozerlib/distgit.py
@@ -29,7 +29,7 @@ from artcommonlib.assembly import AssemblyTypes
 from artcommonlib.format_util import yellow_print
 from artcommonlib.model import Missing, Model, ListModel
 from artcommonlib.pushd import Dir
-from artcommonlib.release_util import isolate_assembly_in_release
+from artcommonlib.release_util import isolate_assembly_in_release, isolate_el_version_in_release
 from doozerlib import constants, state, util
 from doozerlib.brew import BuildStates
 from doozerlib.dblib import Record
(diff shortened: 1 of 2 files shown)


joepvd:
/approve

There are merge conflicts

PR #453, opened 2024-03-07
Use secure xml parser

adobes1:
Switch to `defusedxml` to prevent issues reported by Snyk. 
`xml.etree.ElementTree` import in `doozer/doozerlib/repodata.py` is kept because `defusedxml.ElementTree` does not have `Element` class which is used in type annotations.

### DIFF
--- a/doozer/doozerlib/repodata.py
+++ b/doozer/doozerlib/repodata.py
@@ -3,7 +3,8 @@ import gzip
 import lzma
 import io
 import logging
-import xml.etree.ElementTree as ET
+import xml.etree.ElementTree
+import defusedxml.ElementTree as ET
 from dataclasses import dataclass, field
 from logging import Logger
 from typing import Any, Dict, List, Optional, Set, Tuple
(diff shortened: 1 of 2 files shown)


doozer/doozerlib/repodata.py
```diff
@@ -78,7 +79,7 @@ def from_dict(nvrea_dict: Dict):
         )
 
     @staticmethod
-    def from_metadata(metadata: ET.Element):
+    def from_metadata(metadata: xml.etree.ElementTree.Element):
```
locriandev, line 82: What does `defusedxml.ElementTree.fromString` return? We should update the type hint accordingly, or remove it at all, so that references to `xml.etree.ElementTree` can go away.

adobes1: According to the project [documentation](https://pypi.org/project/defusedxml/), it should still return the original `xml.etree.ElementTree.Element`.
> All functions and parser classes accept three additional keyword arguments. They return either the same objects as the original functions or compatible subclasses.

And defusedxml's [source code](https://github.com/tiran/defusedxml/blob/main/defusedxml/ElementTree.py#L151) seems to support it.

locriandev (commented):
/lgtm

ashwindasr:
I see that we have a lgtm. @adobes1 anything left in this PR or shall we merge it?

adobes1:
There's nothing left, it can be merged.

PR #475, opened 2024-03-19
Disable embargoed plashets for now

thegreyd:
Disable embargoed plashets until plashet creation issues resolve
see: https://redhat-internal.slack.com/archives/GDBRP5YJH/p1710762665903979


joepvd:
If anything, let's take out unembargoed repos

/hold

PR #479, opened 2024-03-19
plashet: Use queryRPMSigs API to check if rpm signed

vfreex:
This new method should be much faster on new buildvm.


joepvd (approved):
/lgtm

vfreex:
/hold cancel
simple local tests passed.

PR #486, opened 2024-03-20
Do not fetch the release date for custom assemblies

locriandev:
If `gen-assembly` is run with `IN_FLIGHT_PREV = none` and a custom assembly (e.g. `art1234`), pyartcd will try to infer the inflight prev from the release schedule, and fail to do so. It is now throwing an exception, which prevents us from running gen-assembly with custom assemblies.

The proposed fix is to skip fetching the release date for custom assemblies, where IN_FLIGHT_PREV and PREVIOUS shall be None

[Test run ](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fgen-assembly/13/console) (broke because `koji.GenericError: No such build: 'ubi8-container-416.94.202403121338-0-1136'`)

### DIFF
--- a/artcommon/artcommonlib/util.py
+++ b/artcommon/artcommonlib/util.py
@@ -1,10 +1,10 @@
 from typing import OrderedDict, Optional, Tuple
 from datetime import datetime
-from artcommonlib.constants import RELEASE_SCHEDULES
-import requests
 import re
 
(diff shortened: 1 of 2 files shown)


joepvd (approved):
/lgtm

PR #493, opened 2024-03-22
Do not count on exception when extracting non-existing binaries

joepvd:


joepvd:
/close
Superseded by https://github.com/openshift-eng/art-tools/pull/492

PR #508, opened 2024-03-27
[ART-9249] Set reconciliation bugs as RH employee

ashwindasr:
ref https://issues.redhat.com/browse/ART-9249

Setting this will prevent the bug from showing up in the customer portal.


sosiouxme:
I think there's no problem with doing this (nobody outside needs to see alignment bugs), but I'm not sure the premise is correct... I suspect bugs that ship still go public, one way or another.

joepvd (approved):
/lgtm

PR #509, opened 2024-03-27
[ART-8482] Add labels for better JIRA tracking

ashwindasr:
ref https://issues.redhat.com/browse/ART-8482

### DIFF
--- a/doozer/doozerlib/cli/images_streams.py
+++ b/doozer/doozerlib/cli/images_streams.py
@@ -796,7 +796,7 @@ Jira mapping: https://github.com/openshift-eng/ocp-build-data/blob/main/product.
         fields = {
             'project': {'key': project},
             'issuetype': {'name': 'Bug'},
-            'labels': ['art:reconciliation', f'art:distgit:{distgit_key}'],
+            'labels': ['art:reconciliation', f'art:package:{image_meta.get_component_name()}'],
             'versions': [{'name': release_version}],  # Affects Version/s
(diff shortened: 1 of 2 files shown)


doozer/doozerlib/cli/images_streams.py
```diff
@@ -796,6 +796,7 @@ def search_issues(query):
         fields = {
             'project': {'key': project},
             'issuetype': {'name': 'Bug'},
+            'labels': ["art:reconciliation", f"art:distgit:{distgit_key}"],
```
sosiouxme, line 799: this is petty but good style is to use consistent quoting
```suggestion
            'labels': ['art:reconciliation', f'art:distgit:{distgit_key}'],
```
(good style is also to use double quotes, not single, but that's more controversial)

sosiouxme (commented):
/lgtm

doozer/doozerlib/cli/images_streams.py
```diff
@@ -796,6 +796,7 @@ def search_issues(query):
         fields = {
             'project': {'key': project},
             'issuetype': {'name': 'Bug'},
+            'labels': ['art:reconciliation', f'art:distgit:{distgit_key}'],
```
joepvd, line 799: I think `imagemeta` should have `get_component_name()`, which returns the brew component. Think there's some value in uniformity here.

ashwindasr: I made a wrong assumption that get_component_name retrieves the component in product.yaml page.

Updated

joepvd (approved):
/lgtm

PR #512, opened 2024-03-28
also sync noarch rpm

Ximinhan:
test job: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fsync-for-ci/4829/console

### DIFF
--- a/doozer/doozerlib/cli/__main__.py
+++ b/doozer/doozerlib/cli/__main__.py
@@ -285,6 +285,7 @@ installonly_limit=3
                    f'--repoid {repo.name} '
                    'reposync '
                    f'--arch {arch} '
+                   '--arch noarch '
                    '--delete '
                    '--download-metadata '
                    f'--metadata-path {metadata_dir} '


doozer/doozerlib/cli/__main__.py
```diff
@@ -280,6 +280,8 @@ def beta_reposync(runtime, output, cachedir, arch, repo_type, names, dry_run):
                 continue
 
             color_print('Syncing repo {}'.format(repo.name), 'blue')
+            if arch == "x86_64":
```
vfreex, line 283: Could you explain why noarch rpms should only be included in x86_64 repos? I think repos for all arches should include noarch rpms as well.

doozer/doozerlib/cli/__main__.py
```diff
@@ -280,6 +280,8 @@ def beta_reposync(runtime, output, cachedir, arch, repo_type, names, dry_run):
...
+                arch = arch + ' --arch noarch '
```
vfreex, line 284: The naming of the variable will be misleading: It is named "arch", but it is actually part of command line options.

doozer/doozerlib/cli/__main__.py
Ximinhan: updated

joepvd (approved):
Awesome!
/lgtm

PR #513, opened 2024-03-28
Fix doozerlib

Ximinhan:
init el_target if not set
test job: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4/399/parameters/


joepvd:
/hold
Looks scary to me, think a full understanding of what is going on and what circumstance triggers the bug to occur for the limited set of images that experience it, is needed.

PR #514, opened 2024-03-28
latest_build_list(): log the search pattern when no builds are found

locriandev:
Currently we only get a single warning that goes like this:
```
No builds detected for using prefix: 'openshift-base-nodejs-container-v4.16.', extra_pattern: '*', assembly: 'test', build_state: 'COMPLETE', el_target: 'None'
```
This is the final message of an iterative process that attempts to find builds by progressively broadening the search pattern. It would be useful to see what queries are being attempted. Since it's happened in the past that `None` made it as a string into the search pattern, let us also assert that this is never happening. Crash otherwise (so we notice and fix).

### DIFF
--- a/doozer/doozerlib/metadata.py
+++ b/doozer/doozerlib/metadata.py
@@ -463,6 +463,7 @@ class Metadata(object):
                 # Include * after pattern_suffix to tolerate other release components that might be introduced later.
                 # Also include a .el<version> suffix to match the new build pattern
                 rhel_pattern = f'{pattern_prefix}{extra_pattern}{pattern_suffix}.{f"el{el_target}" if el_target else ""}*'
+                assert 'None' not in rhel_pattern
                 builds = koji_api.listBuilds(packageID=package_id,
                                              state=None if build_state is None else build_state.value,
                                              pattern=rhel_pattern,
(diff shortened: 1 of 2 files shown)


doozer/doozerlib/metadata.py
```diff
@@ -463,6 +463,7 @@ def latest_build_list(pattern_suffix):
                 # Include * after pattern_suffix to tolerate other release components that might be introduced later.
                 # Also include a .el<version> suffix to match the new build pattern
                 rhel_pattern = f'{pattern_prefix}{extra_pattern}{pattern_suffix}.{f"el{el_target}" if el_target else ""}*'
+                assert 'None' not in rhel_pattern
```
joepvd, line 466: Not sure I am happy about this assert, as a non-payload weirdness could prevent us from creating 4.16 nightlies. I'd lgtm without this assert.

locriandev: If we're using a pattern like `openshift-base-nodejs-container-v4.16.*.stream.None` we're definitely doing something wrong, and we won't find any builds. This assert is there to ensure we did not screw up in translating a missing el target into the pattern. If we did, we better know asap, and failing the job is the fastest way

joepvd: makes sense, thanks!

joepvd (approved):
/lgtm

PR #521, opened 2024-04-02
[ART-7995] Sign component & release images via sigstore

sosiouxme:
introduces a new command `artcd sigstore-sign` that uses sigstore's `cosign` to sign an entire release (payload image and components) and publish the signatures back to the same registry.


pyartcd/pyartcd/oc.py
```diff
@@ -11,6 +11,27 @@
 logger = logging.getLogger(__name__)
 
 
+async def get_image_info(pullspec: str, raise_if_not_found: bool = False):
```
Ximinhan, line 14: suggest to use function [get_release_image_info_from_pullspec](https://github.com/openshift-eng/art-tools/blob/main/pyartcd/pyartcd/oc.py#L70) and add show-multiarch to its params, octool should be able to handle the errors

sosiouxme: that name is confusing. i'll try renaming and modifying.

i'm not sure there's really much benefit to using `openshift_client` since it just wraps `oc` anyway but... i guess it's a bit less code here.

sosiouxme: now i remember why i needed a separate method; it needs to be async.

PR #525, opened 2024-04-04
Block embargoed content only for assembly type stream

joepvd:


thegreyd:
https://github.com/openshift-eng/art-tools/pull/527 is addressing this, closing in favor of that

PR #526, opened 2024-04-04
[ART-8971] Mirror non-embargoed images to public quay

ashwindasr:
Reverts openshift-eng/art-tools#523


ashwindasr:
/close 

in favor of https://github.com/openshift-eng/art-tools/pull/527

PR #527, opened 2024-04-04
[ART-9330] [ART-8971] new embargo process

ashwindasr:
Ref https://issues.redhat.com/browse/ART-8080

The goal of this PR is to setup a new flow so that we can create embargoed nightlies (visible in the private release controller), so that QE and other stakeholders can test the payload well in advance of the embargo being lifted. Currently, we only start building after the embargo is lifted, and we immediately pin the build(s) in the payload and promote. Having the private builds earlier in the private release controller will let us follow the usual pattern of release, but with public nightlies. The flow can be summarized as follow:

- For STREAM:
    - Only public builds are synced to public image streams (thereby to public nightlies), but both public and private builds are synced to private nightlies. Since we rebase from openshift to openshift-priv before building from the latter, we make sure that the private builds always contain the latest public content.
  
- For non-STREAM:
    - A two key approach is adopted.
    - A parameter in build_sync has to be enabled as well. This param will set `--embargo-permit-ack`. (key 1)
    - If a private build is seen, it will fail viability, with code `EMBARGOED_CONTENT`. We need to permit this in the assembly definition (key 2)
     - Both the above condition has to be satisfied. Due to the nature of embargoes, the accidental release of embargoed content, before lift, is severe. Hence this level of gating is proposed.
     - gen-assembly can be run with private nightlies as params, instead of public nightlies.

### DIFF
--- a/doozer/doozerlib/cli/release_gen_payload.py
+++ b/doozer/doozerlib/cli/release_gen_payload.py
@@ -77,12 +77,14 @@ class RepositoryType(Enum):
               help="Also create a release payload for multi-arch/heterogeneous clusters.")
 @click.option("--moist-run", default=False, is_flag=True,
               help="Mirror and determine tags but do not actually update imagestreams.")
+@click.option("--embargo-permit-ack", default=False, is_flag=True,
+              help="Allow embargoed builds to sync publicly in named assemblies")
 @click_coroutine
 @pass_runtime
 async def release_gen_payload(runtime: Runtime, is_name: str, is_namespace: str, organization: str,
                               repository: str, private_repository: str, release_repository: str, output_dir: str,
                               exclude_arch: Tuple[str, ...], skip_gc_tagging: bool, emergency_ignore_issues: bool,
-                              apply: bool, apply_multi_arch: bool, moist_run: bool):
+                              apply: bool, apply_multi_arch: bool, moist_run: bool, embargo_permit_ack: bool):
     """
 Computes a set of imagestream tags which can be assembled into an OpenShift release for this
 assembly. The tags may not be valid unless --apply or --moist-run triggers mirroring.
@@ -143,7 +145,7 @@ read and propagate/expose this annotation in its display of the release image.
         output_dir,
         exclude_arch,
         skip_gc_tagging, emergency_ignore_issues,
-        apply, apply_multi_arch, moist_run
+        apply, apply_multi_arch, moist_run, embargo_permit_ack
     ).run()
 
 
@@ -269,7 +271,7 @@ class GenPayloadCli:
                  repository: str = None, private_repository: str = None, release_repository: str = None,
                  output_dir: str = None, exclude_arch: Tuple[str] = None, skip_gc_tagging: bool = False,
                  emergency_ignore_issues: bool = False, apply: bool = False, apply_multi_arch: bool = False,
-                 moist_run: bool = False):
+                 moist_run: bool = False, embargo_permit_ack: bool = False):
 
         self.runtime = runtime
         self.logger = runtime.logger if runtime else MagicMock()  # in tests, blackhole logs by default
@@ -300,6 +302,8 @@ class GenPayloadCli:
         self.privacy_modes: List[bool] = [False]
         # do we proceed with this payload after weighing issues against permits?
         self.payload_permitted = False
+        # Allows embargoed builds to be released
+        self.embargo_permit_ack = embargo_permit_ack
 
     @start_as_current_span_async(TRACER, "releases:gen-payload")
     async def run(self):
@@ -603,15 +607,22 @@ class GenPayloadCli:
                 continue
             # No adjustment for private or public; the assembly's canonical payload content is the same.
 
-            entries: Dict[str, PayloadEntry]  # Key of this dict is release payload tag name
-            issues: List[AssemblyIssue]
-            entries, issues = PayloadGenerator.find_payload_entries(assembly_inspector, arch, self.full_component_repo(repo_type=RepositoryType.PUBLIC))
+            # entries: Dict[str, PayloadEntry]  # Key of this dict is release payload tag name
+            # issues: List[AssemblyIssue]
+            public_entries, public_payload_issues = PayloadGenerator.find_payload_entries(assembly_inspector, arch, self.full_component_repo(repo_type=RepositoryType.PUBLIC))
+            entries_for_arch[arch] = public_entries
+            self.assembly_issues.extend(public_payload_issues)
 
-            # Exclude issues for now, for private repo
-            private_entries, _ = PayloadGenerator.find_payload_entries(assembly_inspector, arch, self.full_component_repo(repo_type=RepositoryType.PRIVATE))
-            entries_for_arch[arch] = entries
+            # Always make sure that embargoed builds do not flow to public image stream
+            # If releasing after embargo lift, this can be permitted using 'EMBARGOED_CONTENT' code
+            embargo_issues = PayloadGenerator.embargo_issues_for_payload(public_entries, arch)
+            self.assembly_issues.extend(embargo_issues)
+
+            private_entries, private_payload_issues = PayloadGenerator.find_payload_entries(assembly_inspector, arch, self.full_component_repo(repo_type=RepositoryType.PRIVATE))
             private_entries_for_arch[arch] = private_entries
-            self.assembly_issues.extend(issues)
+            # We do not want to add private_payload_issues into the assembly issues list because it will just be the same issues as public, so we'll get duplicates
+            # This can be uncommented if it changes in the future
+            # self.assembly_issues.extend(private_payload_issues)
 
         return entries_for_arch, private_entries_for_arch
 
@@ -706,17 +717,29 @@ class GenPayloadCli:
         Check whether the issues found are permitted,
         and collect issues per component into a serializable report
         """
-
+        permitted_non_acknowledged_embargoed_issues = []
         payload_permitted = True
         assembly_issues_report: Dict[str, List[Dict]] = dict()
         for ai in self.assembly_issues:
             permitted = assembly_inspector.does_permit(ai)
+
+            if self.runtime.assembly_type is not AssemblyTypes.STREAM and ai.code == AssemblyIssueCode.EMBARGOED_CONTENT and permitted:
+                # If this is not a STREAM run, check to see if there are any permitted embargoed content
+                # If there is, --embargo-permit-ack must be specified in build-sync params, as the second and final gate
+                # before letting embargoed builds from being synced to public quay
+                if not self.embargo_permit_ack:
+                    permitted_non_acknowledged_embargoed_issues.append(ai.component)
+
             payload_permitted &= permitted  # If anything not permitted, payload not permitted
             assembly_issues_report.setdefault(ai.component, []).append(dict(
                 code=ai.code.name,
                 msg=ai.msg,
                 permitted=permitted
             ))
+
+        if permitted_non_acknowledged_embargoed_issues:
+            raise IOError(f"WARNING: Embargoed components found: {set(permitted_non_acknowledged_embargoed_issues)} "
+                          "--embargo-permit-ack must be specified if is intended")
 
         return payload_permitted, assembly_issues_report
 
@@ -761,24 +784,28 @@ class GenPayloadCli:
             False: dict()
         }
 
-        # Filter out embargoed builds to mirror to public quay
+        # Check to see if there is an embargoed build and whether --embargo-permit-ack is specified
         filtered_payload_entries_for_arch = {}
         for arch, payload_entries in self.payload_entries_for_arch.items():
             filtered_payload_entries = {}
             for payload_tag_name, payload_entry in payload_entries.items():
-                if payload_entry.rhcos_build is None and payload_entry.build_inspector.is_under_embargo():
-                    # We do not want embargoed builds to get leaked to public quay
-                    # RHCOS builds are considered non-embargoed for now
+                if payload_entry.rhcos_build is not None:
+                    # RHCOS builds can't be embargoes for now
+                    # Filter them out
                     continue
+
+                if payload_entry.build_inspector.is_under_embargo() and self.runtime.assembly_type is AssemblyTypes.STREAM:
+                    # We filter out the embargoed builds from syncing to public quay.
+                    # Once its lifted, we let it through, by adding a permit
+                    continue
+
                 filtered_payload_entries[payload_tag_name] = payload_entry
 
             if filtered_payload_entries:
                 # Sync if there are any builds for this arch, otherwise omit the arch entirely
                 filtered_payload_entries_for_arch[arch] = filtered_payload_entries
 
-        # Ensure that all payload images have been mirrored before updating
-        # the imagestream. Otherwise, the imagestream will fail to import the
-        # image.
+        # filtered_payload_entries_for_arch includes embargoed content when self.embargo_permit_ack is set
         tasks = []
         for arch, payload_entries in self.private_payload_entries_for_arch.items():
             tasks.append(self.mirror_payload_content(arch, payload_entries, True))
@@ -1909,4 +1936,23 @@ class PayloadGenerator:
             tasks.append(PayloadGenerator._check_nightly_consistency(assembly_inspector, nightly, arch))
         results = await asyncio.gather(*tasks)
         issues.extend([issue for result in results for issue in result])
+        return issues
+
+    @staticmethod
+    def embargo_issues_for_payload(entries: Dict[str, PayloadEntry], arch: str) -> List[AssemblyIssue]:
+        """
+        Check to see if there are any embargoed builds in public image stream tags
+        """
+        issues = []
+        for payload_tag, entry in entries.items():
+            if not entry.build_inspector:
+                # Probably RHCOS
+                continue
+
+            if entry.build_inspector.is_under_embargo():
+                issues.append(AssemblyIssue(
+                    f"Found embargoed build {entry.build_inspector.get_nvr()} in payload entries for arch {arch}",
+                    component=entry.image_meta.name,
+                    code=AssemblyIssueCode.EMBARGOED_CONTENT
+                ))
         return issues
--- a/doozer/doozerlib/cli/release_gen_payload.py
+++ b/doozer/doozerlib/cli/release_gen_payload.py
@@ -599,7 +599,7 @@ class GenPayloadCli:
         Returns a dict of dicts, keyed by architecture, then by payload component name.
         """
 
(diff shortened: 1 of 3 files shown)


doozer/doozerlib/cli/release_gen_payload.py
```diff
@@ -767,37 +773,63 @@ async def sync_payloads(self):
...
             for payload_tag_name, payload_entry in payload_entries.items():
                 if payload_entry.rhcos_build is None and payload_entry.build_inspector.is_under_embargo():
+                    if self.runtime.assembly_type is AssemblyTypes.STANDARD and not self.cve_exclusion_ack:
+                        raise IOError(f"Embargoed build detected for payload tag {payload_tag_name} with "
+                                      f"payload_entry: {payload_entry} \n --cve-exclusion-ack must be specified, "
+                                      f"if this was intended")
+                    # Otherwise
                     # We do not want embargoed builds to get leaked to public quay
                     # RHCOS builds are considered non-embargoed for now
                     continue
```
sosiouxme, line 783: I think this can be simplified if at this point, we just don't consider the build private, and let it be included in the payload entries from the start, while noting that there are private entries, e.g.


```
                    # (note RHCOS builds can't be embargoes for now)
                    if self.runtime.assembly_type is not AssemblyTypes.STANDARD:
                        # Unless releasing, we do not want embargoed builds synced to public quay
                        continue
                    if self.cve_exclusion_ack:
                        # consciously releasing private content
                        self.includes_embargo = True
                        # probably create an AssemblyIssue somehow here to be permitted by the assembly?
                    else:
                        raise IOError(f"Embargoed build detected for payload tag {payload_tag_name} with "
                                                f"payload_entry: {payload_entry} \n --cve-exclusion-ack must be specified, "
                                                f"if this was intended")
```

ashwindasr: > probably create an AssemblyIssue somehow here to be permitted by the assembly?
I was thinking we can have a flag in build sync?

artcommon/artcommonlib/assembly.py
```diff
@@ -381,3 +381,7 @@ def assembly_metadata_config(releases_config: Model, assembly: typing.Optional[s
             config_dict = _merger(component_entry.metadata.primitive(), config_dict)
 
     return Model(dict_to_model=config_dict)
+
+
+def contains_embargoed_build(releases_config: Model, assembly: str):
```
thegreyd, line 386: I'm not sure what the utility of having it defined in assembly is .. or what we would be preventing.

artcommon/artcommonlib/assembly.py
```diff
@@ -381,3 +381,7 @@ def assembly_metadata_config(releases_config: Model, assembly: typing.Optional[s
...
+    return releases_config.releases[assembly].contains_embargoed_build in [True, 'true', 'True', 'yes']
```
jupierce, line 387: I would suggest a change here. `contains_embargoed_build` reads as if the stanza is asserting the payload does or does not contain a embargoed fix whereas I believe it is trying to imply "I know this payload contains something embargoed and we are ready to make it public". 
Permits are our typical method of capturing "I know this payload <is unusual for some reason>" semantics. So `code: UNEMBARGO` would make sense to me. 
Alternatively, a boolean like `acknowledge_unembargo` in the assembly, but I'd like to understand the rationale.

doozer/doozerlib/cli/release_gen_payload.py
```diff
@@ -761,25 +767,50 @@ async def sync_payloads(self):
...
+                if payload_entry.rhcos_build is None and payload_entry.build_inspector.is_under_embargo():
+                    # (note RHCOS builds can't be embargoes for now)
+                    if self.runtime.assembly_type is not AssemblyTypes.STANDARD:
+                        # Unless releasing, we do not want embargoed builds synced to public quay
+                        continue
+                    if self.embargo_permit_ack and self.runtime.contains_embargo:
+                        # consciously releasing private content
+                        self.includes_embargo = True
+                    else:
+                        raise IOError(f"Embargoed build detected for payload tag {payload_tag_name} with "
```
jupierce, line 784: Iterate through all results before throwing the error. Imagine if there were 2 embargoed images in the payload: one that was ready to be unembargoed and one that was not. You'd want the artist to know about both before making the decision to permit the unembargo.

jupierce (dismissed):

artcommon/artcommonlib/assembly.py
sosiouxme: I thought when I looked previously that permits are not observed in build-sync, but it seems like they are (via `assembly_inspector.does_permit`). `release:gen-payload` could add an `AssemblyIssue` with code `EMBARGOED_CONTENT` whenever there's private content on a standard assembly, and then we could `permit` it in the assembly. I guess that would be nice to avoid adding a new attr.

doozer/doozerlib/cli/release_gen_payload.py
sosiouxme: easy if it's just adding an `AssemblyIssue`. but this may require a little rearrangement; this sub doesn't currently contribute to assembly issues, and i don't recall where we actually check for viability. we want to make sure that's given a chance to evaluate and stop issues before we've actually copied any content out.

doozer/doozerlib/cli/release_gen_payload.py
```diff
@@ -1893,3 +1933,20 @@ async def check_nightlies_consistency(assembly_inspector: AssemblyInspector) ->
         results = await asyncio.gather(*tasks)
         issues.extend([issue for result in results for issue in result])
         return issues
+
+    @staticmethod
+    def does_payload_entry_include_embargo(entries) -> List[AssemblyIssue]:
```
sosiouxme, line 1938: sounds like a boolean, should describe what it's returning
```suggestion
    def embargo_issues_for_payload(entries) -> List[AssemblyIssue]:
```

sosiouxme (changes requested):
just some surface-level things

i'm not sure what to think about dividing into private/public sets of issues, let's talk about that

ashwindasr:
Review issues fixed in https://github.com/openshift-eng/art-tools/pull/527/commits/d45438debbe81e6532393db3600f90cb59f154b4

ashwindasr:
/unhold

Our automation has access to private nightlies as well

ashwindasr:
/hold

Until testing is completed

doozer/doozerlib/cli/release_gen_payload.py
```diff
@@ -761,25 +782,44 @@ async def sync_payloads(self):
...
+        filtered_payload_entries_for_arch = {}
+        for arch, payload_entries in self.payload_entries_for_arch.items():
+            filtered_payload_entries = {}
+            for payload_tag_name, payload_entry in payload_entries.items():
+                if payload_entry.rhcos_build is not None:
+                    # RHCOS builds can't be embargoes for now
+                    # Filter them out
+                    continue
+
+                if self.runtime.assembly_type is AssemblyTypes.STREAM:
```
sosiouxme, line 795: think you only wanted to filter embargoed ones here... this filters everything for `STREAM`

doozer/doozerlib/cli/release_gen_payload.py
```diff
@@ -761,25 +782,44 @@ async def sync_payloads(self):
...
+                    # We filter out the embargoed builds from syncing to public quay.
+                    # Once its lifted, we let it through, by adding a permit
+                    continue
+
+            if filtered_payload_entries:
```
sosiouxme, line 800: nothing ever sets it, so how can it have any content?

i think you wanted to add the entries after the `if` "filters"

doozer/doozerlib/cli/release_gen_payload.py
```diff
@@ -603,15 +607,20 @@ def generate_payload_entries(self, assembly_inspector: AssemblyInspector) -> (Di
...
-            entries_for_arch[arch] = entries
+            # Always make sure that embargoed builds do not flow to public image stream
+            # If releasing after embargo lift, this can be permitted using 'EMBARGOED_CONTENT' code
+            embargo_issues = PayloadGenerator.embargo_issues_for_payload(public_entries)
+            self.assembly_issues.extend(embargo_issues)
+
+            private_entries, private_payload_issues = PayloadGenerator.find_payload_entries(assembly_inspector, arch, self.full_component_repo(repo_type=RepositoryType.PRIVATE))
             private_entries_for_arch[arch] = private_entries
-            self.assembly_issues.extend(issues)
+            self.assembly_issues.extend(private_payload_issues)
```
sosiouxme, line 623: i think we do not want to add these into the assembly issues list because it will just be the same issues as public, so we'll get duplicates (that's my memory looking at the logic this is replacing; and looks like issues only come from RHCOS at this point). the assembly issues, in addition to being checked for permits, are also added to annotations on the imagestream, so we don't want to bloat them any further.

sosiouxme (commented):
sorry i'm out of time to review properly, will try to get back to it.
test method is pretty simple:
1. for an old release like 4.11, rebuild a p0 build as p1 (clone the distgit, change the release, brew container-build)
2. run build-sync to see if it correctly detects that this is an embargo
3. on a fork create a bogus release and build-sync it to see that it correctly handles the flag and permit being there or not.

doozer/tests/cli/test_gen_payload.py
```diff
@@ -209,15 +211,17 @@ def test_detect_inconsistent_images(self):
         self.assertEqual(gpcli.assembly_issues, ["stuff"])
 
     @patch("doozerlib.cli.release_gen_payload.PayloadGenerator.find_payload_entries")
-    def test_generate_payload_entries(self, pg_fpe_mock):
+    @patch("doozerlib.cli.release_gen_payload.PayloadGenerator.embargo_issues_for_payload")
```
sosiouxme, line 214: previously there was nothing in this method to inspect the output of `find_payload_entries`, which was just mocked with dummy responses; you added a call to `embargo_issues_for_payload` which actually look at those, so either need to come up with better responses, or mock that too.

doozer/doozerlib/cli/release_gen_payload.py
```diff
@@ -706,18 +717,30 @@ def summarize_issue_permits(self, assembly_inspector: AssemblyInspector) -> (boo
...
             payload_permitted &= permitted  # If anything not permitted, payload not permitted
             assembly_issues_report.setdefault(ai.component, []).append(dict(
                 code=ai.code.name,
                 msg=ai.msg,
                 permitted=permitted
             ))
 
+        if permitted_non_acknowledged_embargoed_issues:
+            raise IOError(f"WARNING: Embargoed components found: {set(permitted_non_acknowledged_embargoed_issues)} "
+                          "--embargo-permit-ack must be specified if is intended")
```
sosiouxme, line 742:
```suggestion
            raise IOError(f"Embargoed components found: {set(permitted_non_acknowledged_embargoed_issues)}; "
                          "must specify --embargo-permit-ack flag to publish private content")
```

doozer/doozerlib/cli/release_gen_payload.py
```diff
@@ -603,15 +607,22 @@ def generate_payload_entries(self, assembly_inspector: AssemblyInspector) -> (Di
                 continue
             # No adjustment for private or public; the assembly's canonical payload content is the same.
 
-            entries: Dict[str, PayloadEntry]  # Key of this dict is release payload tag name
-            issues: List[AssemblyIssue]
-            entries, issues = PayloadGenerator.find_payload_entries(assembly_inspector, arch, self.full_component_repo(repo_type=RepositoryType.PUBLIC))
+            # entries: Dict[str, PayloadEntry]  # Key of this dict is release payload tag name
+            # issues: List[AssemblyIssue]
+            public_entries, public_payload_issues = PayloadGenerator.find_payload_entries(assembly_inspector, arch, self.full_component_repo(repo_type=RepositoryType.PUBLIC))
+            entries_for_arch[arch] = public_entries
```
sosiouxme, line 613:
```suggestion
            entries_for_arch[arch] = {k: v for k, v in public_entries.items() if not v.is_under_embargo()}
```

doozer/doozerlib/cli/release_gen_payload.py
```diff
@@ -761,25 +784,46 @@ async def sync_payloads(self):
...
+                if payload_entry.build_inspector.is_under_embargo() and self.runtime.assembly_type is AssemblyTypes.STREAM:
+                    # We filter out the embargoed builds from syncing to public quay.
+                    # Once its lifted, we let it through, by adding a permit
+                    continue
+
+                filtered_payload_entries[payload_tag_name] = payload_entry
+
+            if filtered_payload_entries:
+                # Sync if there are any builds for this arch, otherwise omit the arch entirely
+                filtered_payload_entries_for_arch[arch] = filtered_payload_entries
```
sosiouxme, line 806: get rid of filtering at this point

sosiouxme (changes requested):

ashwindasr:
/unhold

Testing complete. Slack [thread](https://redhat-internal.slack.com/archives/GDBRP5YJH/p1715888032404899)

sosiouxme (changes requested):
check my assertion that there would be a bug here...

doozer/doozerlib/cli/release_gen_payload.py
```diff
@@ -603,15 +607,38 @@ def generate_payload_entries(self, assembly_inspector: AssemblyInspector) -> (Di
...
+
+            self.assembly_issues.extend(public_payload_issues)
+
+            # Always make sure that embargoed builds do not flow to public image stream
+            # If releasing after embargo lift, this can be permitted using 'EMBARGOED_CONTENT' code
+            embargo_issues = PayloadGenerator.embargo_issues_for_payload(public_entries, arch)
+            self.assembly_issues.extend(embargo_issues)
+
+            private_entries, private_payload_issues = PayloadGenerator.find_payload_entries(assembly_inspector, arch, self.full_component_repo(repo_type=RepositoryType.PRIVATE))
             private_entries_for_arch[arch] = private_entries
```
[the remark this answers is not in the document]
ashwindasr: you're right, the original list was being processed again, instead of the filtered one

doozer/doozerlib/cli/release_gen_payload.py
```diff
@@ -603,15 +607,37 @@ def generate_payload_entries(self, assembly_inspector: AssemblyInspector) -> (Di
...
+
+                if v.build_inspector.is_under_embargo() and self.runtime.assembly_type == AssemblyTypes.STREAM:
+                    # It's an embargoed build. Filter it out if its stream
+                    continue
+
+                public_entries[k] = v
 
-            # Exclude issues for now, for private repo
-            private_entries, _ = PayloadGenerator.find_payload_entries(assembly_inspector, arch, self.full_component_repo(repo_type=RepositoryType.PRIVATE))
             entries_for_arch[arch] = entries
```
thegreyd, line 628: This should be public_entries i think since entries_for_arch is how the payload will be constructed
```suggestion
            entries_for_arch[arch] = public_entries
```

doozer/doozerlib/cli/release_gen_payload.py
```diff
@@ -603,15 +607,37 @@ def generate_payload_entries(self, assembly_inspector: AssemblyInspector) -> (Di
...
+            # If releasing after embargo lift, these can be permitted using 'EMBARGOED_CONTENT' code
+            embargo_issues = PayloadGenerator.embargo_issues_for_payload(public_entries, arch)
+            self.assembly_issues.extend(embargo_issues)
+
+            private_entries, private_payload_issues = PayloadGenerator.find_payload_entries(assembly_inspector, arch, self.full_component_repo(repo_type=RepositoryType.PRIVATE))
             private_entries_for_arch[arch] = private_entries
-            self.assembly_issues.extend(issues)
+            # We do not want to add private_payload_issues into the assembly issues list because it will just be the same issues as public, so we'll get duplicates
+            # This can be uncommented if it changes in the future
+            # self.assembly_issues.extend(private_payload_issues)
```
thegreyd, line 640: I think this will miss If there is an issue with a private build, so this should get us only the private build issues
```suggestion
            private_only_issues = set(private_payload_issues) - set(payload_issues)
            self.assembly_issues.extend(list(private_only_issues))
```

doozer/doozerlib/cli/release_gen_payload.py
thegreyd: can we also rename this to `public_entries_for_arch`

doozer/doozerlib/cli/release_gen_payload.py
```diff
@@ -773,13 +811,14 @@ async def sync_payloads(self):
 
         await asyncio.sleep(120)
 
-        # Update the imagestreams being monitored by the release controller.
-        tasks = []
-        for arch, payload_entries in self.payload_entries_for_arch.items():
-            for private_mode in self.privacy_modes:
+        # Updating public and private image streams
+        for private_mode, payload_entries_for_each_arch_iter in [(False, self.payload_entries_for_arch), (True, self.private_payload_entries_for_arch)]:
```
thegreyd, line 817: let's rename the attr self.payload_entries_for_arch -> self.public_payload_entries_for_each
so it's crystal clear that these are supposed to be only public entries

thegreyd:
Let's write a couple of unit tests, testing
- public and private builds are being returned in separate buckets
- when private build is in public bucket, the assembly issue code is being raised
- when assembly issue code is raised, we complain if ack flag is not found (and a case testing vice versa)

doozer/doozerlib/cli/release_gen_payload.py
```diff
@@ -604,16 +608,40 @@ def generate_payload_entries(self, assembly_inspector: AssemblyInspector) -> (Di
...
+
+            private_entries, private_payload_issues = PayloadGenerator.find_payload_entries(assembly_inspector, arch, self.full_component_repo(repo_type=RepositoryType.PRIVATE))
             private_entries_for_arch[arch] = private_entries
-            self.assembly_issues.extend(issues)
 
-        return entries_for_arch, private_entries_for_arch
+            # Check to see if there are private only issues, and add them to the list of assembly issues
+            private_only_issues = set(private_payload_issues) - set(payload_issues)
+
+            self.assembly_issues.extend(list(private_only_issues))
```
sosiouxme, line 642: i debated whether it made sense to add these at all. seems like it could be confusing if there were any and they showed up on the public payloads. OTOH they would show issues that could become public if the payload were promoted so... i guess it makes sense to include them.

sosiouxme (approved):
thanks, looks clearer. now. sid has some requests but you have a thumbs up from me.

sosiouxme:
/hold
in case you want to address sid's requests before merging...
or that could be another PR :shrug:

ashwindasr:
/hold

Until the team decides when to merge this to prod

ashwindasr:
/unhold

due to a emergency

PR #528, opened 2024-04-04
Make sure assembly name matches found microshift build

thegreyd:
Our logic doesn't validate assembly name for found nvrs, so it just goes with unshipped builds.
This can result in cases like this one https://github.com/openshift-eng/ocp-build-data/pull/4614/files 
where 4.15.7 build was pinned and attached to 4.15.8 

Tested with https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fbuild-microshift/81/console


locriandev (commented):
/lgtm

PR #537, opened 2024-04-09
Elliott find builds do not look at el target

locriandev:


doozer/tests/cli/test_gen_payload.py
```diff
@@ -576,7 +576,7 @@ async def test_create_multi_manifest_list(self, open_mock, exec_mock, fmlsha_moc
         await gpcli.create_multi_manifest_list("spam", arch_to_payload_entry, "ocp-multi")
         self.assertEqual(
             exec_mock.call_args[0][0],
-            "manifest-tool  push from-spec /tmp/ocp-multi.spam.manifest-list.yaml")
+            "manifest-tool --docker-cfg=/run/user/1000/containers/auth.json push from-spec /tmp/ocp-multi.spam.manifest-list.yaml")
```
vfreex, line 579: Does the path of docker-cfg depend on the env var where the unit test is run? We should make it a constant for unit tests

vfreex (commented):
One nit. Otherwise LGTM

doozer/tests/cli/test_gen_payload.py
locriandev: Fixed

PR #544, opened 2024-04-11
Populate jira mappings at runtime

ashwindasr:


ashwindasr:
/close

in favor of https://github.com/openshift-eng/art-tools/pull/545/

PR #547, opened 2024-04-12
Do not post traceback to slack

thegreyd:
Due to our jira instance being quite volatile we've been getting a lot of jira related job failures,
each of these job failures posts a lengthy traceback in slack which doesn't help.
Even otherwise traceback.format_exc from pyartcd aren't helpful and do not get to the actual error, so remove them


joepvd (approved):
/lgtm

PR #549, opened 2024-04-16
[ART-9446] Populate jira mappings 

ashwindasr:
If anything goes wrong with the request to JIRA API at compile time, pyartcd won't compile. even if the pipeline run is not doing anything with elliott or jira.

So lets override it later


vfreex (commented):
/lgtm
/approve

PR #550, opened 2024-04-17
get jira field from function

Ximinhan:
test job: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fcheck-bugs/29/parameters/

### DIFF
--- a/elliott/elliottlib/bzutil.py
+++ b/elliott/elliottlib/bzutil.py
@@ -29,7 +29,27 @@ from elliottlib.errata_async import AsyncErrataAPI
 from elliottlib.metadata import Metadata
 from elliottlib.util import isolate_timestamp_in_release, chunk
 
+
 logger = logutil.get_logger(__name__)
+
+
+def get_jira_client() -> JIRA:
+    token_auth = os.environ.get("JIRA_TOKEN")
+    if not token_auth:
+        raise ValueError(f"elliott requires login credentials for {constants.JIRA_SERVER_URL}. Set a JIRA_TOKEN env var ")
+    client = JIRA(constants.JIRA_SERVER_URL, token_auth=token_auth)
+    return client
+
+
+def get_jira_field_id(client, field_name):
+    """
+    Returns the field id for a given field name
+    """
+    fields = client.fields()
+    for field in fields:
+        if field['name'] == field_name:
+            return field['id']
+    raise ValueError(f"Could not find field {field_name} in {constants.JIRA_SERVER_URL}")
 
 
 # This is easier to patch in unit tests
@@ -210,6 +230,7 @@ class BugzillaBug(Bug):
 class JIRABug(Bug):
     def __init__(self, bug_obj: Issue):
         super().__init__(bug_obj)
+        self.jira_client = None
 
     @property
     def id(self):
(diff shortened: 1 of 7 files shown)


elliott/elliottlib/bzutil.py
```diff
@@ -29,9 +29,29 @@
 from elliottlib.metadata import Metadata
 from elliottlib.util import isolate_timestamp_in_release, chunk
 
+
 logger = logutil.get_logger(__name__)
 
 
+def get_jira_client() -> JIRA:
```
thegreyd, line 36: Why have this? We already have JiraBugTracker class to encapsulate all of the Jira logic

elliott/elliottlib/bzutil.py
```diff
@@ -29,9 +29,29 @@
...
+    token_auth = os.environ.get("JIRA_TOKEN")
+    if not token_auth:
+        raise ValueError(f"elliott requires login credentials for {constants.JIRA_SERVER_URL}. Set a JIRA_TOKEN env var ")
+    client = JIRA(constants.JIRA_SERVER_URL, token_auth=token_auth)
+    return client
+
+
+def get_jira_field_id(client, field_name):
```
thegreyd, line 44: Why have a separate func? This should go in JiraBugTracker class

elliott/elliottlib/bzutil.py
```diff
@@ -210,6 +230,7 @@ def creation_time_parsed(self):
 class JIRABug(Bug):
     def __init__(self, bug_obj: Issue):
         super().__init__(bug_obj)
+        self.jira_client = None
```
thegreyd, line 233: Why do we want to add this..?

elliott/elliottlib/constants.py
```diff
@@ -128,4 +128,4 @@
 errata_post_erratum_url = errata_url + "/api/v1/erratum"
 errata_get_advisories_for_bug_url = errata_url + "/bugs/{id}/advisories.json"
 
-JIRA_API_FIELD = "https://issues.redhat.com/rest/api/2/field"
+JIRA_SERVER_URL = "https://issues.redhat.com"
```
thegreyd, line 131: why encode this here when we have bug.yml config and JiraBugTracker class to handle that?

elliott/elliottlib/bzutil.py
thegreyd: Okay I think I understand the intent. To get access to the custom field info, we need access to jira client. This is fine then

elliott/elliottlib/bzutil.py
```diff
@@ -290,7 +311,9 @@ def version(self):
 
     @property
     def blocked_by_bz(self):
-        url = getattr(self.bug.fields, JIRABugTracker.field_blocked_by_bz)
+        if not self.jira_client:
+            self.jira_client = get_jira_client()
+        url = getattr(self.bug.fields, get_jira_field_id(self.jira_client, "Blocked by Bugzilla Bug"))
```
thegreyd, line 316: I don't particularly like this pattern. We should fail if there is no jira client stored in bug
And in all methods of JiraBugTracker where we are fetching bugs, we store jira client in every jira bug

PR #556, opened 2024-04-19
[ART-9446] skip next minor when it doesn't exist

ashwindasr:
Looks like code is looking for `openshift-4.18`

```
stderr>>error: pathspec 'openshift-4.18' did not match any file(s) known to git
```

logs[ here](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fcheck-bugs/33/)


Ximinhan:
I think check-bug should not go into 4.18. the issue is the input version list is from ocp4Versions in commonlib.groovy, check bug should only check those versions in active release. those EOL and pre-release version should not be passed in the check list.

ashwindasr:
/close 

In favor of https://github.com/openshift-eng/art-tools/pull/559

PR #561, opened 2024-04-24
use branch in image config

Ximinhan:
When building the bundle, the current branch and target rely on the global configuration, but some images have been migrated to RHEL 9 while the global configuration remains set to RHEL 8. To ensure consistency and accuracy in the build process, it is imperative to honor the specific configuration of each image. Therefore, this pull request aims to optimize build system to correctly identify and apply the specific configuration for each image instead of solely relying on the global configuration.

test job: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Folm_bundle/271/console

### DIFF
--- a/doozer/doozerlib/olm/bundle.py
+++ b/doozer/doozerlib/olm/bundle.py
@@ -610,12 +610,12 @@ class OLMBundle(object):
 
     @property
     def target(self):
-        target_match = re.match(r'.*-rhel-(\d+)(?:-|$)', str(self.branch()))
+        target_match = re.match(r'.*-rhel-(\d+)(?:-|$)', self.branch)
         if target_match:
             el_target = int(target_match.group(1))
             return self.runtime.get_default_candidate_brew_tag(el_target=el_target) or '{}-candidate'.format(self.branch)
         else:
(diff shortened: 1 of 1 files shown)


joepvd:
/hold
Before merging this, I would like to understand which bundle builds in which versions would start to be built in rhel9 rather than rhel8 branch, and if we can expect problems with either building or shipping.

thegreyd:
Not having this in is causing trouble..
When new images are onboarded for rhel9, it and it's bundle image packages are added to our rhel9 build tag by HoneyBadger. And when we try to default to what's defined in group "rhel8", then builds fail since they are not added to the rhel8 build tag, and we have to open up requests like these https://issues.redhat.com/browse/CWFCONF-9545 which takes a bunch of manual effort for us and rel-eng folks.

From what I see the operator and bundle building and shipping is being handled together by HB and is consistent.

See example (and their related HB pipeline runs):
https://issues.redhat.com/browse/RCMAUTO-6490
https://issues.redhat.com/browse/RCMAUTO-6489 

So I think this is good to merge

PR #569, opened 2024-04-25
sudo podman to see root images

ashwindasr:
Since otherwise we are looking at only the images in non-root context


joepvd (approved):
/lgtm

PR #574, opened 2024-04-26
doozer gen-payload: Assert that latest is latest

thegreyd:
Follow up to https://github.com/openshift-eng/art-tools/pull/562

See slack thread: https://redhat-internal.slack.com/archives/GDBRP5YJH/p1713954580446079

https://issues.redhat.com/browse/RHELBLD-15024

Brew api is returning us outdated builds, this is to verify it in assembly_inspector which gets called via gen-payload when we create nightlies (which only runs for --stream without a brew event passed in)

Brew call - `listBuilds(completeBefore=brew_event_timestamp)`

## Test
```
$ doozer --assembly=stream --group=openshift-4.17 -i ose-installer-artifacts release:gen-payload --output-dir=gen-payload-artifacts
...
ValueError: 
Latest build ose-installer-artifacts-container-v4.17.0-202404251943.p0.g388a35c.assembly.stream.el8 
is not the true latest build ose-installer-artifacts-container-v4.17.0-202404260414.p0.g59cc24d.assembly.stream.el8 
for image ose-installer-artifacts.
Brew event: 57677105 timestamp: 1714153567.77471 2024-04-26 17:46:07.774710 UTC. 
True latest build completion ts: 1714140511.85667 2024-04-26 14:08:31.856669
```

PR #578, opened 2024-04-29
Workaround for RHELBLD-15024

thegreyd:
Fix tested in combination with 
https://github.com/openshift-eng/art-tools/pull/574
and https://gist.github.com/thegreyd/311eabef91689d6abbcf7b1c7fe4bf2e

### DIFF
--- a/doozer/doozerlib/brew.py
+++ b/doozer/doozerlib/brew.py
@@ -8,6 +8,7 @@ import json
 import threading
 import time
 import traceback
+from datetime import datetime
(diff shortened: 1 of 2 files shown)


ashwindasr:
/lgtm

Noting for posterity:
```
ts = koji_session.getLastEvent()['ts']
dt = datetime.utcfromtimestamp(ts)
latest_build_obj = koji_session.listBuilds(completeBefore=str(dt), **kwargs)[0])
```

PR #584, opened 2024-05-01
Move and make use of canonical builders config read in runtime

thegreyd:
We have a method that we use to parse the "canonical_builders_from_upstream"
config key, it has special meaning and modes. But recently we started reading 
and depending on it's raw value when init'ing various commands. So move that 
method to runtime and rely on that


locriandev (commented):
/lgtm
/hold

`is_canonical_builders_config_enabled()` is getting more complex due to nesting functions, can we have unit tests for those?

joepvd:
gonna:
/close
...as the value of this has decreased after the move to rhel9 was completed, and when we need to move to rhel10, our situation is probably different.

PR #594, opened 2024-05-02
Regression alerts only if next version is GA

thegreyd:
Make sure next version is GA for doing regression checks
Relies on `software_lifecycle`  config in group.yml

Test: elliott --assembly 4.15.12 -g openshift-4.15 verify-attached-bugs

### DIFF
--- a/elliott/elliottlib/runtime.py
+++ b/elliott/elliottlib/runtime.py
@@ -13,6 +13,7 @@ import click
 import yaml
 
 from artcommonlib import logutil
+from artcommonlib import exectools
 from artcommonlib.assembly import AssemblyTypes, assembly_type, assembly_basis_event, assembly_group_config
 from artcommonlib.model import Model, Missing
 from artcommonlib.runtime import GroupRuntime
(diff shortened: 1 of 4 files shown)


elliott/elliottlib/runtime.py
```diff
@@ -444,6 +445,20 @@ def get_releases_config(self):
...
+            raise ValueError("Do not use this method to access files from the current branch")
+
+        # fetch branch first
+        cmd = f"git -C {self.data_dir} fetch origin {branch}:{branch}"
+        exectools.cmd_assert(cmd)
+
+        cmd = f"git -C {self.data_dir} show {branch}:{filename}"
+        rc, out, err = exectools.cmd_gather(cmd)
+        if rc != 0:
+            raise IOError(f"Failed to get {filename} from {branch} branch")
```
thegreyd, line 459:
```suggestion
            raise IOError(f"Failed to get {filename} from {branch} branch: {err}")
```

Ximinhan:
generally lgtm, but I think we can use https://art-dash-server-art-dashboard-server.apps.artc2023.pc3z.p1.openshiftapps.com/api/v1/ga-version?format=json to get the latest ga version, which will make it simpler

Ximinhan:
> I thought we were already skipping regression checks when next version is not GA.
for check bug its true as you mentioned, what we need to do is in promote job https://github.com/openshift-eng/art-tools/blob/main/pyartcd/pyartcd/pipelines/promote.py#L251-L264
I created another pr https://github.com/openshift-eng/art-tools/pull/631 for this

thegreyd:
I would not like us to depend on art-dash api call to figure this out. if art-dash is down we lose this basic check

locriandev (commented):
/lgtm

PR #600, opened 2024-05-02
elliott set batch command

thegreyd:
Set batch by id for find if one exists for named assembly and use that.
Note: It does not automatically create a batch

Test:
```
$ elliott --assembly 4.14.24 -g openshift-4.14 errata batch set --auto
...
2024-05-02 20:05:54,481 INFO Advisory 131291 - changing batch from 2842 ➔ 2881
2024-05-02 20:05:55,492 INFO Advisory 131293 - changing batch from 2842 ➔ 2881
2024-05-02 20:05:56,621 INFO Advisory 131295 - changing batch from 2842 ➔ 2881
2024-05-02 20:05:57,524 INFO Advisory 131296 - changing batch from 2842 ➔ 2881
2024-05-02 20:05:58,445 INFO Advisory 131297 - changing batch from 2842 ➔ 2881
```


Ximinhan:
I was thinking put the batch id we created in the assembly config and treat it as the source of truth, then advisory in the same assembly know which batch to align with instead of search it by name.

PR #608, opened 2024-05-03
Select existing bundle image for assembly if delivery labels match

thegreyd:
When preparing prerelease assembly like ECs, we set `operator_index_mode: pre-release` 
in assembly definition. This gets used by olm-bundle:rebase-and-build to set the appropriate 
labels on the bundle image. These are different for ECs and normal stream builds that we do.
So we cannot reuse normal stream builds that we build everyday via olm_bundle job for ECs,
or any assembly that has something different for `operator_index_mode` config than what's in
group.yml for the group. 

To get around this we opted to force rebuild bundle images for prerelease everytime, (except
when advisory was already populated which then we skipped). But this isn't perfect. We've been
seeing the bundle rebase and build command failures where it would fail if distgit branch doesn't
exist or if a rebase/build fails. This leads those builds to get lost and leading us and pipeline to 
force build them multiple times. This build time is significant so it costs us time and attention.

Instead we can try to match labels that we expect bundle image to have and make sure latest
build has those. This would mean when we run this command in quick successions, even with
failures, existing builds would be found and reused.

Note that we are not matching all labels but only delivery_labels which can change for assembly.

## Test
```
./doozer --group=openshift-4.16 --assembly ec.6 olm-bundle:rebase-and-build --dry-run
```

### DIFF
--- a/doozer/doozerlib/olm/bundle.py
+++ b/doozer/doozerlib/olm/bundle.py
@@ -2,6 +2,7 @@ import glob
 import io
 import os
 import re
+import json
 import shutil
 import threading
 from pathlib import Path
@@ -13,7 +14,6 @@ from dockerfile_parse import DockerfileParser
 from koji import ClientSession
 
 from artcommonlib import pushd, exectools
-from artcommonlib.build_util import find_latest_build
 from doozerlib import brew, util
 from doozerlib.runtime import Runtime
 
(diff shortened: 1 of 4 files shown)


doozer/doozerlib/olm/bundle.py
```diff
@@ -581,3 +591,15 @@ def target(self):
...
+        cmd = f"oc image info {bundle_image_pullspec} -o json"
+        rc, out, err = exectools.cmd_gather(cmd)
+        if rc != 0:
+            raise ValueError(f"Error running {cmd}: {err}")
+        image_info = json.loads(out)
+        image_labels = image_info['config']['config']['Labels']
+        for label, value in self.redhat_delivery_tags.items():
+            if image_labels.get(label, '') != value:
+                return False
+        return True
```
sosiouxme, line 605: fine, but clearer to me:
```suggestion
        image_labels = json.loads(out)['config']['config']['Labels']
        return all(image_labels.get(label) == val for label, val in self.redhat_delivery_tags.items())
```

sosiouxme (approved):
I like it. Only rebuild what you don't already have, and actually simplifies the prepare-release logic.

Just to be sure... `--force` should still always rebuild, for the case where something has changed about bundle construction. Don't think this interferes, just pointing it out.

sosiouxme:
weird, it's adding both `approved` and `lgtm` on a positive review. I'm fine with this code actually, just wanted to give a chance to consider my suggestion before merging automatically.

PR #609, opened 2024-05-03
[ART-7995] pyartcd.promote: call sigstore signing

sosiouxme:
uses the same signatory code from sigstore-sign to sign a release.
for now it will only sign component images, not release images due to OTA-1267.

Must merge with https://github.com/openshift-eng/aos-cd-jobs/pull/4146 - either alone breaks promotion


joepvd (approved):
/lgtm

thegreyd:
Tests are breaking. What I'll do is merge and then followup with a fix

PR #612, opened 2024-05-06
for ec and rc use current date to get in flight release

Ximinhan:
fix issue like https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fgen-assembly/63/consoleFull
don't break when assembly is ec or rc, when release not found in pp use current date as default value.


joepvd:
/hold
This is needed to get the test green: https://github.com/openshift-eng/art-tools/pull/613. Let's not add machine-os-content.

Ximinhan:
we need to select prerelease for rc when gen-assembly, then ec or rc won't get into this get_inflight function

PR #614, opened 2024-05-06
Use art automation address for art image build fail emails

thegreyd:
Right now we're getting emails for build failures of images that we own (ci golang builders)
on our team address. Reroute these to art automation address.

### DIFF
--- a/pyartcd/pyartcd/util.py
+++ b/pyartcd/pyartcd/util.py
@@ -597,9 +597,9 @@ The following logs are just the container build portion of the OSBS build:
 
         # Send email to owners of failed image builds
         # If art is the only owner of image (example for our ci golang builder images) send instead to our default automation email
-        owner = failure['owners']
+        owner = (failure['owners']
                 if (failure['owners'] and failure['owners'] != ["aos-team-art@redhat.com"])
-                else default_owner
+                else default_owner)
         mail_client.send_mail(
             to=['aos-art-automation+failed-ocp-build@redhat.com', owner],
             subject=f'Failed OCP build of {failure["image"]}:{failure["version"]}',
--- a/pyartcd/pyartcd/util.py
+++ b/pyartcd/pyartcd/util.py
@@ -595,9 +595,13 @@ The following logs are just the container build portion of the OSBS build:
                             f'{failure["message"]}\n' \
                             f'{container_log}'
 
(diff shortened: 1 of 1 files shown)


thiagoalessio (approved):
/lgtm

pyartcd/pyartcd/util.py
```diff
@@ -595,9 +595,13 @@ def mail_build_failure_owners(failed_builds: dict, doozer_working: str, mail_cli
                             f'{failure["message"]}\n' \
                             f'{container_log}'
 
+        # Send email to owners of failed image builds
+        # If art is the only owner of image (example for our ci golang builder images) send instead to our default automation email
+        owner = failure['owners']
+                if (failure['owners'] and failure['owners'] != ["aos-team-art@redhat.com"])
```
ashwindasr, line 601: Probably needs `\` at EOLs right?

ashwindasr (changes requested):

pyartcd/pyartcd/util.py
```diff
@@ -595,9 +595,13 @@ def mail_build_failure_owners(failed_builds: dict, doozer_working: str, mail_cli
...
+                else default_owner
         mail_client.send_mail(
-            to=['aos-art-automation+failed-ocp-build@redhat.com',
-                failure['owners'] if failure['owners'] else default_owner],
+            to=['aos-art-automation+failed-ocp-build@redhat.com', owner],
```
ashwindasr, line 604: Also `owner` can be `None`, since `default_owner` can be none, since its a non required param in ocp4. Not sure if that will cause issues with the mail_client?

pyartcd/pyartcd/util.py
thegreyd: Ah you're right, thanks for catching that

pyartcd/pyartcd/util.py
thegreyd: default_owner is passed via what's set in Jenkinsfile param which is our automation email. So it works fine for now, but if you think it's important to test/improve validations, we can work on that in a separate PR

pyartcd/pyartcd/util.py
thegreyd: fixed now

ashwindasr (approved):

PR #615, opened 2024-05-06
Cleanup prepare-release logs

thegreyd:
Our job logs have considerably increased in length, making them hard to parse through.
- Make the cmd_gather `time elapsed` information debug
- Don't double warn in elliott find-bugs sweep

### DIFF
--- a/elliott/elliottlib/cli/find_bugs_sweep_cli.py
+++ b/elliott/elliottlib/cli/find_bugs_sweep_cli.py
@@ -368,14 +368,10 @@ def categorize_bugs_by_type(bugs: List[Bug], advisory_id_map: Dict[str, int],
 
     not_found = set(tracker_bugs) - found
     if not_found:
-        not_found_with_component = [(b.id, b.whiteboard_component) for b in not_found]
-        logger.warning('No attached builds found in advisories for tracker bugs (bug, package):'
-                       f' {not_found_with_component}.')
-
         still_not_found = not_found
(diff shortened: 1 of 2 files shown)


elliott/elliottlib/cli/find_bugs_sweep_cli.py
```diff
@@ -368,14 +368,10 @@ def categorize_bugs_by_type(bugs: List[Bug], advisory_id_map: Dict[str, int],
 
     not_found = set(tracker_bugs) - found
     if not_found:
-        not_found_with_component = [(b.id, b.whiteboard_component) for b in not_found]
-        logger.warning('No attached builds found in advisories for tracker bugs (bug, package):'
-                       f' {not_found_with_component}.')
```
ashwindasr, line 373: So I guess we don't care about the case in which `not_found` is `True` and `permitted_bug_ids` is false?

If so,
/lgtm

Ximinhan: these two lines will log the original `not_found` bug list, after removing the `permitted_bug_ids` is `still_not_found`

thegreyd: Yeah, this logging was little bit like a debug thing since I wrote this piece of code - to confirm we are filtering correctly. Now it's unecessary and is noisy to see the same log message twice

PR #582, opened 2024-04-30
Get missing bundle branches first

thegreyd:
If distgit branch doesn't exist, fail fast
```
./doozer --group=openshift-4.16 --assembly ec.6 olm-bundle:rebase-and-build --dry-run

ERROR One or more bundle branches do not exist: ['containers/ose-smb-csi-driver-operator-bundle/rhaos-4.16-rhel-8', 'containers/dpu-operator-bundle/rhaos-4.16-rhel-8']
```

PR #621, opened 2024-05-07
Fix olm bundle invocations

thegreyd:
Follow up to https://github.com/openshift-eng/art-tools/pull/582/files
This also deprecates `olm-bundle:build` and `olm-bundle:rebase` commands.
We don't use these anywhere in the pipeline since we always want to build and
rebase as a single operation. Supporting these now takes extra effort (with refactoring)
so remove them.


joepvd (approved):
/lgtm

PR #629, opened 2024-05-09
Error out when verify attached operator fails

ashwindasr:
Since this causes issues with IIB

### DIFF
--- a/pyartcd/pyartcd/pipelines/prepare_release.py
+++ b/pyartcd/pyartcd/pipelines/prepare_release.py
@@ -268,7 +268,7 @@ class PrepareReleasePipeline:
                                                          advisories['metadata'])
             except Exception as ex:
                 _LOGGER.warning(f"Unable to verify attached operators: {ex}")
-                await self._slack_client.say_in_thread("Unable to verify attached operators. Details in log.")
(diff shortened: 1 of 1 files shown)


thegreyd:
@ashwindasr this will still not fail the command. This is where we are catching the exception https://github.com/ashwindasr/art-tools/blob/26bcbe42dd659df1114793b833d915c1e4386bb8/pyartcd/pyartcd/pipelines/prepare_release.py#L269 - so tweak that

pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -268,7 +268,7 @@ async def run(self):
                                                          advisories['metadata'])
             except Exception as ex:
                 _LOGGER.warning(f"Unable to verify attached operators: {ex}")
-                await self._slack_client.say_in_thread("Unable to verify attached operators. Details in log.")
+                await self._slack_client.say_in_thread("Unable to verify attached operators. Details in log.", reaction=":art-attention:")
```
thegreyd, line 271:
```suggestion
                await self._slack_client.say_in_thread("`elliott verify-attached-operators` failed, details in log. Please retry", reaction=":art-attention:")
```

PR #630, opened 2024-05-10
allow multi ovn23.06 in repo

Ximinhan:


joepvd:
/close
The check does a "starts with" so ovn23.06 AND ovn 24.03 will all match.

PR #631, opened 2024-05-10
skip blocking bug check for latest minor release

Ximinhan:
Regression alerts only if next version is GA when promote


thegreyd:
I do not think we need this after https://github.com/openshift-eng/art-tools/pull/594
atleast not it in its current form, this could be an additional check in the styley of ^

PR #632, opened 2024-05-10
gen-assembly: notify when non-latest green amd64 nightly is selected

locriandev:
Depending on timing logics, it may happen that latest amd64 nightly does not have an equivalent set across all arches. In this case, release-artists might want to investigate and see if this is acceptable, or if a manual intervention is required to capture latest nightly changes.


thiagoalessio (approved):
/lgtm

PR #643, opened 2024-05-20
allow EMBARGOED_CONTENT in standard permits

ashwindasr:
Permits are usually not allowed in standard assemblies. But after embargo lift, we would like to explicitly add the permit so that we make consciously publish private builds to public image streams.

### DIFF
--- a/artcommon/artcommonlib/assembly.py
+++ b/artcommon/artcommonlib/assembly.py
@@ -234,7 +234,6 @@ def assembly_permits(releases_config: Model, assembly: typing.Optional[str]) -> 
     """
     defined_permits = assembly_config_struct(releases_config, assembly, 'permits', [])
 
-    non_embargo_permits = []
     # Do some basic validation here to fail fast
     if assembly_type(releases_config, assembly) == AssemblyTypes.STANDARD:
         if any(permit.code != AssemblyIssueCode.EMBARGOED_CONTENT.name for permit in defined_permits):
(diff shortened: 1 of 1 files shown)


artcommon/artcommonlib/assembly.py
```diff
@@ -230,10 +234,15 @@ def assembly_permits(releases_config: Model, assembly: typing.Optional[str]) ->
...
     # Do some basic validation here to fail fast
     if assembly_type(releases_config, assembly) == AssemblyTypes.STANDARD:
-        if defined_permits:
-            raise ValueError(f'STANDARD assemblies like {assembly} do not allow "permits"')
+        for permit in defined_permits:
+            if permit.code != AssemblyIssueCode.EMBARGOED_CONTENT.name:
+                non_embargo_permits.append(permit)
+
+        if non_embargo_permits:
+            raise ValueError(f'STANDARD assemblies like {assembly} only allow {AssemblyIssueCode.EMBARGOED_CONTENT.name} in "permits"')
```
sosiouxme, line 245: if you're not going to do anything with `non_embargo_permits` then there's no real need to record it.
```suggestion
        if any(permit.code != AssemblyIssueCode.EMBARGOED_CONTENT.name for permit in defined_permits):
            raise ValueError(f'STANDARD assemblies like {assembly} only allow {AssemblyIssueCode.EMBARGOED_CONTENT.name} in "permits"')
```

sosiouxme (approved):
/lgtm

PR #644, opened 2024-05-21
Skip metadata population during pre-release

ashwindasr:
Using the same logic as advance advisory prep. If pre-release is detected, drop all builds from metadata advisory and attach builds to pre-release advisory only (skipping metadata advisory prep).

### DIFF
--- a/pyartcd/pyartcd/pipelines/prepare_release.py
+++ b/pyartcd/pyartcd/pipelines/prepare_release.py
@@ -109,6 +109,8 @@ class PrepareReleasePipeline:
 
         # This will be set to True if advance operator advisory is detected
         self.advance_release = False
+        # This will be set to True if operator pre-release advisory is detected
+        self.pre_release = False
 
     async def run(self):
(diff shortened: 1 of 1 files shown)


pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -238,6 +240,19 @@ async def run(self):
                 _LOGGER.info(f"'advance' advisory {advisory_info['id']} is not editable. Defaulting bundle advisory"
                              " to 'metadata'")
 
+        if "prerelease" in advisories.keys():
```
thegreyd, line 243: We can add a check `if self.pre_release and self.advance_release` and error out

thegreyd (changes requested):
This is one more place where we can add similar condition for prerelease https://github.com/openshift-eng/art-tools/blob/9bf8f5c4ffc2c5be707dce02c602a7e7fc597be9/pyartcd/pyartcd/pipelines/prepare_release.py#L308

PR #649, opened 2024-05-22
Update certificate for validator

Ximinhan:
test run : https://github.com/openshift-eng/ocp-build-data/pull/4872


vfreex:
/hold
 This will prevent requests library to connect to public https URLs.

PR #659, opened 2024-05-30
Respect cachito global config

thegreyd:
This proposes that we only look at `cachito.enabled` config as the source of truth,
first in the meta config, then in the global config. And ignore the pkg_managers 
config if the global config is disabled.

Should go with config changes, https://github.com/openshift-eng/ocp-build-data/pull/4900

### DIFF
--- a/doozer/doozerlib/distgit.py
+++ b/doozer/doozerlib/distgit.py
@@ -580,7 +580,7 @@ class ImageDistGitRepo(DistGitRepo):
             if self.runtime.group_config.cachito.enabled:
                 cachito_enabled = True
             elif isinstance(self.config.content.source.pkg_managers, ListModel):
-                self.logger.warning(f"Cachito config for {self.name} has no effect since cachito is not enabled in "
+                self.logger.warning(f"pkg_managers directive for {self.name} has no effect since cachito is not enabled in "
                                     "image meta or group config.")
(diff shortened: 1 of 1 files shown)


doozer/doozerlib/distgit.py
```diff
@@ -570,15 +570,18 @@ def _generate_osbs_image_config(self, version: str) -> Dict:
...
         # https://osbs.readthedocs.io/en/latest/users.html#remote-sources
         cachito_enabled = False
         if self.config.cachito.enabled:
             cachito_enabled = True
         elif self.config.cachito.enabled is Missing:
-            if isinstance(self.config.content.source.pkg_managers, ListModel) or self.runtime.group_config.cachito.enabled:
+            if self.runtime.group_config.cachito.enabled:
                 cachito_enabled = True
+            elif isinstance(self.config.content.source.pkg_managers, ListModel):
+                self.logger.warning(f"Cachito config for {self.name} has no effect since cachito is not enabled in "
```
locriandev, line 583:
```suggestion
                self.logger.warning(f"pkg_managers directive for {self.name} has no effect since cachito is not enabled in "
```

locriandev (commented):
/lgtm

locriandev (approved):
/lgtm

PR #676, opened 2024-06-05
[ART-9795] Konflux rebase and build

ashwindasr:
Changes in this PR:
- Use quay.io registry urls in `FROM` since Konflux builds from an to Quay. Use `quay.io/rh_ee_asdas/konflux-test` for testing, for now and then switch to team registry.


vfreex:
Will be pulled into a separate PR

PR #679, opened 2024-06-06
Fix get rhcos version at ocp4_scan time

joepvd:
The `-o json` option invalidates the `--template` in `oc get`. With this change, ocp4_scan should be looking at the right data.


locriandev (approved):
/lgtm

PR #684, opened 2024-06-10
[art cluster] Fix Dockerfile install error

ashwindasr:
Since `requests` was installed by an RPM, it wasn't letting `pip install` upgrade it to the newer version


ashwindasr:
Green [build](https://console-openshift-console.apps.artc2023.pc3z.p1.openshiftapps.com/k8s/ns/art-cd/builds/art-cd-base-32/)

PR #687, opened 2024-06-11
Do not inherit advance advisory

thegreyd:


doozer/doozerlib/cli/release_gen_assembly.py
```diff
@@ -554,17 +553,14 @@ def _get_advisories_release_jira(self) -> Tuple[Dict[str, int], str]:
...
-            if self.pre_ga_mode == 'prerelease':
-                advisories['prerelease'] = -1
-            else:
-                advisories.pop('prerelease')
+        # preGA advisories (prerelease/advance) associated with an assembly should not be reused
+        # they should be shipped or dropped if not shipping
+        for key in preGA_advisory_type:
+            previous_advisories.pop(key, None)
+
+        advisories.update(previous_advisories)
```
ashwindasr, line 563: I don't think the update method removes keys from the dict its trying to update right?

ashwindasr (changes requested):

doozer/doozerlib/cli/release_gen_assembly.py
```diff
@@ -554,17 +553,14 @@ def _get_advisories_release_jira(self) -> Tuple[Dict[str, int], str]:
...
-        # see https://docs.google.com/document/d/1S_ivD8Sh85LuZKnClqvoRvIhn4hNJQA76jpfUaFZjUA/edit
-        if 'prerelease' in advisories:
-            if self.pre_ga_mode == 'prerelease':
-                advisories['prerelease'] = -1
-            else:
-                advisories.pop('prerelease')
+        # preGA advisories (prerelease/advance) associated with an assembly should not be reused
+        # they should be shipped or dropped if not shipping
+        for key in preGA_advisory_type:
+            previous_advisories.pop(key, None)
```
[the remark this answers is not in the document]
thegreyd: yes but that would be the exception not the rule, so by default we don't inherit. If we decide to, then we manually edit the assembly

doozer/doozerlib/cli/release_gen_assembly.py
thegreyd: Yes removal is done in the previous line `.pop`

ashwindasr: `advisories` are initialized as 
```
advisories = {
            'image': -1,
            'rpm': -1,
            'extras': -1,
            'metadata': -1,
        }
```
so we're good

PR #690, opened 2024-06-12
Set prerelease label in bundle build always

thegreyd:
Follow up to https://github.com/openshift-eng/art-tools/pull/608
When trying to match labels, we missed the case when a label would be present in 
found image that we do not want or want it to be false. To get around this always set prerelease label to 
whatever state it is in (true/false)

### DIFF
--- a/doozer/doozerlib/olm/bundle.py
+++ b/doozer/doozerlib/olm/bundle.py
@@ -553,9 +553,8 @@ class OLMBundle(object):
         labels = {
             'com.redhat.delivery.operator.bundle': 'true',
             'com.redhat.openshift.versions': versions.format(**self.runtime.group_config.vars),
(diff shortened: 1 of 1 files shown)


doozer/doozerlib/olm/bundle.py
```diff
@@ -553,9 +553,8 @@ def redhat_delivery_tags(self):
         labels = {
             'com.redhat.delivery.operator.bundle': 'true',
             'com.redhat.openshift.versions': versions.format(**self.runtime.group_config.vars),
+            'com.redhat.prerelease': str(self.operator_index_mode == 'pre-release'),
```
thegreyd, line 556:
```suggestion
            'com.redhat.prerelease': str(self.operator_index_mode == 'pre-release').lower(),
```

PR #693, opened 2024-06-13
add pre-check when set advisory batch

Ximinhan:


vfreex (commented):
This is incorrect.

If an advisory is already on a batch, calling `/api/v1/erratum/{advisory_id}/change_batch`   will be able to change it to another batch. The reason why we saw error in https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fprepare-release/166/ is that the unexpected batch is locked.

I am gonna propose a new PR https://github.com/openshift-eng/art-tools/pull/697 to explicitly set the batch ID.

vfreex:
Included in https://github.com/openshift-eng/art-tools/pull/697

PR #695, opened 2024-06-14
promote: Fix invalid json

joepvd:
promote can fail while reading the json that artcd produces. The json includes logs from aws s3 interactions. This ensures that stdout of the aws gets redirected to stderr.


vfreex (commented):
/lgtm
/approve

PR #697, opened 2024-06-14
prepare-release: Explicitly add newly created advisory to batch

vfreex:
Due to an issue in Errata Tool, if batch is enabled for an ET Release,
newly created advisories will be implicitly added to an open batch.

This has caused chaos because our ET release (RHOSE ASYNC - AUTO) is
shared among multiple teams (ART, CPaaS, etc) and advisories for
different OCP releases are prepared in parallel.

I would like to propose a new CLI option `--batch-id` for the `elliott
create` command, so that we can explicitly specify the batch ID when
creating an advisory. To achieve this, raw Errata REST API is used since
the errata_tool library doesn't expose that field.

Also updated prepare_release job to use the new CLI option.

### DIFF
--- a/elliott/elliottlib/cli/create_cli.py
+++ b/elliott/elliottlib/cli/create_cli.py
@@ -1,10 +1,12 @@
+from typing import Optional
 import click
-import datetime
+from datetime import datetime
 from artcommonlib import logutil
 from artcommonlib.assembly import AssemblyTypes
 from artcommonlib.exectools import cmd_assert
 from artcommonlib.format_util import green_prefix
-from elliottlib.cli.common import cli
+from elliott.elliottlib.errata_async import AsyncErrataAPI
+from elliottlib.cli.common import cli, click_coroutine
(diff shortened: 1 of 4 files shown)


vfreex:
This PR includes https://github.com/openshift-eng/art-tools/pull/692.

thegreyd:
@vfreex thanks for this! overall lots of good things, I previously had a wip at https://github.com/openshift-eng/art-tools/pull/600/files , overall i like the idea of encapsulating logic like this in elliott in some way. Then as release operators we can run these commands as needed on buildvm easily. wdyt?

thegreyd:
But we can also do this in another PR, don't want to block this getting in sooner to save effort

vfreex:
@thegreyd I am ok with having a separate set batch command if there is a need. This PR allows explicitly speicify `--batch-id` when creating advisories, but it doesn't cover the use case of "correcting incorrect batch ID".

Ximinhan:
let's merge it to see what happened

PR #700, opened 2024-06-17
ART-9934: Clean up FIPS installer archive paths

zaneb:
Currently the `openshift-install-rhel9-amd64.tar.gz` archive contains the file `artcd_working/to_mirror/openshift-v4/x86_64/clients/ocp/4.16.0-rc.4/openshift-baremetal-install`. Change this to the desired value, which is `openshift-install-fips`.


joepvd (approved):
/lgtm

Thanks!

PR #702, opened 2024-06-17
verify-attached-bugs: Use raw advisory fetch to be fast and less error prone

thegreyd:
Using Erratum(..) from errata_tool to initialize advisory is very slow and error prone for big GA advisories, 
It makes an http call for each build attached to the advisory (and more things) which 
we do not need and we hit timeout error in ET often which crashes us.
So instead use raw erratum api to fetch bugs for these advisories

## Test
`./elliott --assembly rc.4 -g openshift-4.16 verify-attached-bugs --verify-bug-status --skip-multiple-advisories-check
`


locriandev (commented):
/lgtm

PR #708, opened 2024-06-20
Revert "prepare-release: Explicitly add newly created advisory to batch"

ashwindasr:
Reverts openshift-eng/art-tools#697

prepare-release is failing, check job run for errors


ashwindasr:
/hold

seeing if https://github.com/openshift-eng/art-tools/pull/709 might fix it

PR #713, opened 2024-06-21
ART-9937: Publish openshift-install-fips for multiarch payloads

zaneb:


zaneb:
/cc @locriandev @joepvd 
I believe this is correct, but I have no way of testing it myself.

locriandev:
Hi @zaneb , thanks for looking into this. To test it out, we'll need to promote a release from this branch. AFAICS, all releases for this week have already been promoted so this means we'll have to defer testing (and merging) to next week

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -840,6 +840,8 @@ async def publish_multi_client(self, base_to_mirror_dir: str, pullspec: str, rel
             os.makedirs(client_mirror_dir)
             # extract release clients tools
             extract_release_client_tools(pullspec, f"--to={client_mirror_dir}", go_arch)
+            # extract baremetal installer binary
+            self.publish_baremetal_installer_binary(pullspec, client_mirror_dir, arch)
```
vfreex, line 844: Not tested, but seems `arch` should be `goarch`.

zaneb: I don't think so, because it's converted to `goarch` inside `publish_baremetal_installer()`: https://github.com/openshift-eng/art-tools/blob/1002eac91ecd6da518275a47fdf191ee7c6c7e2e/pyartcd/pyartcd/pipelines/promote.py#L790

Ximinhan:
tested with 4.15.20 https://mirror.openshift.com/pub/openshift-v4/multi/clients/ocp/4.15.20/amd64/

PR #714, opened 2024-06-21
Fix newly created advisory ID being set to 0

vfreex:
There is an issue in
https://github.com/openshift-eng/art-tools/pull/697, where `elliott create` command doesn't print newly created advisory ID to stdout. This breaks prepare-release job and causes 0 populated to releases.yml (e.g. https://github.com/openshift-eng/ocp-build-data/commit/52a2b2f8c5920cb734a30eb824862d657aa529f7)


locriandev (commented):
/lgtm

PR #716, opened 2024-06-24
ART-9844: Stop editing go.mod

joepvd:
Turns out, ART should not be in the business of editing go.mod. It is
strictly fine if the toolset version in go.mod is lower than the go
version that builds the project. Bumping the version in go.mod has
numerous side effects that the current naive implementation is not
taking care of. More importantly, most upstream projects get rebase
errors when getting their upstream in the release branches when this got
edited in between.


joepvd:
test run: https://github.com/openshift/csi-operator/pull/235

locriandev (commented):
/lgtm

PR #717, opened 2024-06-24
ART-9503 Handle multiple plashet host locations

locriandev:
`ocp-artifacts` is being migrated to a new place. To make this transition smoother, the plan is to:
- start producing plashets on both (old and new one) hosts
- let the world know we're migrating
- after a reasonable time, give a final notice and decommission the old host

This PR tackles the first point. This [ocp4 build](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4/3575/console) rebuilt 4.7 plashets, after having wiped them out manually on current ocp-artifacts VM. The result is  visible on both indexes:
- https://ocp-artifacts.hosts.prod.psi.rdu2.redhat.com/pub/RHOCP/plashets/4.7/
- https://ocp-artifacts-art--runtime-int.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/pub/RHOCP/plashets/4.7/

### DIFF
--- a/pyartcd/pyartcd/pipelines/rebuild.py
+++ b/pyartcd/pyartcd/pipelines/rebuild.py
@@ -1,7 +1,6 @@
 import asyncio
 import logging
 import os
-import re
 import shutil
 from collections import namedtuple
 from configparser import ConfigParser
@@ -37,7 +36,7 @@ PlashetBuildResult = namedtuple("PlashetBuildResult", ("repo_name", "local_dir",
 class RebuildPipeline:
     """ Rebuilds a component for an assembly """
 
-    def __init__(self, runtime: Runtime, group: str, assembly: str,
+    def __init__(self, runtime: Runtime, group: str, assembly: str, plashet_remote: dict,
                  type: RebuildType, dg_key: str, ocp_build_data_url: str, logger: Optional[logging.Logger] = None):
         if type in [RebuildType.RPM, RebuildType.IMAGE] and not dg_key:
             raise ValueError("'dg_key' is required.")
@@ -47,6 +46,7 @@ class RebuildPipeline:
         self.runtime = runtime
         self.group = group
         self.assembly = assembly
+        self.plashet_remote = plashet_remote
         self.type = type
         self.dg_key = dg_key
         self.logger = logger or runtime.logger
@@ -207,7 +207,7 @@ class RebuildPipeline:
         else:
             await exectools.cmd_assert_async(cmd, env=self._doozer_env_vars)
 
-        remote_url = constants.PLASHET_REMOTE_URL + f"/{major}.{minor}"
+        remote_url = self.plashet_remote['url'] + f"/{major}.{minor}"
         if el_version >= 8:
             remote_url += f"-el{el_version}"
         remote_url += f"/{self.assembly}/{directory_name}"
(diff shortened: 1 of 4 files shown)


pyartcd/pyartcd/pipelines/rebuild.py
```diff
@@ -207,7 +207,7 @@ async def _build_plashet_from_tags(self, name: str, directory_name: str, el_vers
         else:
             await exectools.cmd_assert_async(cmd, env=self._doozer_env_vars)
 
-        remote_url = constants.PLASHET_REMOTE_URL + f"/{major}.{minor}"
+        remote_url = constants.PLASHET_REMOTES[0]['url'] + f"/{major}.{minor}"
```
thegreyd, line 210: I'm thinking something like PLASHET_REMOTE = {..} and PLASHET_REMOTE_NEW = {..} would help readability and make it clear which one is being used.

thegreyd: also why isn't this using self.plashet_remote ?

pyartcd/pyartcd/pipelines/rebuild.py
```diff
@@ -265,7 +265,7 @@ async def _build_plashet_for_assembly(self, name: str, directory_name: str, el_v
             await exectools.cmd_assert_async(cmd, env=self._doozer_env_vars)
 
         major, minor = self._ocp_version
-        remote_url = constants.PLASHET_REMOTE_URL + f"/{major}.{minor}"
+        remote_url = constants.PLASHET_REMOTES[0]['url'] + f"/{major}.{minor}"
```
thegreyd, line 268: why isn't this self.plashet_remote ?

pyartcd/pyartcd/pipelines/rebuild.py
```diff
@@ -603,9 +606,12 @@ async def rebuild(runtime: Runtime, ocp_build_data_url: str, version: str, assem
             runtime.logger.warning('Env var BUILD_URL has not been defined: '
                                    'a random identifier will be used for the locks')
 
-        await locks.run_with_lock(
-            coro=pipeline.run(),
-            lock=lock,
-            lock_name=lock_name,
-            lock_id=lock_identifier
-        )
+        await asyncio.gather(*[
```
thegreyd, line 609: idk if it's worth it to have these changes in rebuild job right now.. we very rarely use rebuild job. If our scheduled plashet sync in ocp4 which runs regularly has these changes in - i.e. running for both hosts, then i think that is sufficient for testing and confirmation that everything works. Since we will be reverting back to running only on one host, so I'm thinking maybe not touch the rebuild job or very minimal changes like constant names etc

pyartcd/pyartcd/pipelines/rebuild.py
locriandev: Updated

pyartcd/pyartcd/pipelines/rebuild.py
locriandev: Updated by using `self.plashet_remote`

pyartcd/pyartcd/pipelines/rebuild.py
locriandev: `rebuild` hasn't run in a long time, but I was told this job was meant to be used with custom hotfixes. So it was suggested not to deprecate it for the moment.

This means `rebuild` also has to deal with a plurality of `PLASHET_REMOTES`. We could indeed get the first element in the list, but as you pointed out in a previous comment this could be cause of confusion with respect to which host we're rebuilding for.

In my opinion, `rebuild` can definitely handle multiple rebuilds in parallel; we're making use of asyncio tasks to optimize time usage, besides we'll rarely run this pipeline if ever. I would leave it like this.

joepvd (approved):
/lgtm

PR #721, opened 2024-06-25
fix plashet bug

ashwindasr:


locriandev:
Not a bug in the code, but an issue with SSH login for `jenkins` user

PR #725, opened 2024-06-26
Improve log handling

vfreex:
This PR improves log handling logic in following items:
- New helper function `setup_logging` is added to `artcommonlib` to set up logging for doozer and elliott in a common way.
- Logging level provided by command line and formatter strings are configured to the root logger, so that they are proprogated to child loggers obtained through `logging.getLogger`.
- No need to pass `runtime.logger` everywhere to log with configured options.  `logging.getLogger(__name__)` is able to obtain a logger from a different module without coupling Runtime.
- RedactingFilter is added to remove sensitive information like passwords, tokens, and keys from logs.

### DIFF
--- a/artcommon/artcommonlib/logutil.py
+++ b/artcommon/artcommonlib/logutil.py
@@ -1,9 +1,78 @@
 import logging
+import re
 
 
 class EntityLoggingAdapter(logging.LoggerAdapter):
     def process(self, msg, kwargs):
         return '[%s] %s' % (self.extra['entity'], msg), kwargs
+
+
+class RedactingFilter(logging.Filter):
+    """ A filter to remove sensitive information like passwords, tokens, and keys from logs
+    https://docs.python.org/3/library/logging.html#filter-objects
+    """
+
+    PATTERNS = {
+        'requests_gssapi.gssapi_': [
+            (re.compile(r'Authorization header:(.+)', re.IGNORECASE), 'Authorization header: <redacted>'),
+        ]
+    }
+
+    def filter(self, record: logging.LogRecord):
+        if record.name in self.PATTERNS.keys():
+            self._redact(record)
+        return True
+
+    def _redact(self, record: logging.LogRecord):
+        msg = isinstance(record.msg, str) and record.msg or str(record.msg)
+        patterns = self.PATTERNS[record.name]
+        for pattern, replacement in patterns:
+            m = pattern.search(msg)
+            if m:
+                record.msg = pattern.sub(replacement, msg)
+                break
+
+
+def setup_logging(log_level: int, debug_log_path: str):
+    """
+    Setup logging with common configurations for art-tools.
+
+    This function:
+    - configure the root logger to log messages with levels of DEBUG and higher to the specified debug.log file
+    - configure the root logger to log messages to the console with the specified log level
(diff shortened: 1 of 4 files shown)


artcommon/artcommonlib/logutil.py
```diff
@@ -1,11 +1,82 @@
...
+        msg = isinstance(record.msg, str) and record.msg or str(record.msg)
+        patterns = self.PATTERNS[record.name]
+        for pattern, replacement in patterns:
+            m = pattern.search(msg)
+            if m:
+                record.msg = pattern.sub(replacement, msg)
+                break
+
+
+def setup_logging(log_level: int, debug_log_path: str):
```
locriandev, line 36: This actually returns a value, so maybe it should be called `get_root_logger` or similar?

doozer/doozerlib/runtime.py
```diff
@@ -686,36 +696,8 @@ def initialize_logging(self):
...
-        main_stream_handler.setFormatter(default_log_formatter)
-        main_stream_handler.setLevel(log_level)
-        self.logger.addHandler(main_stream_handler)
-
-        debug_log_handler = logging.FileHandler(self.debug_log_path)
-        # Add thread information for debug log
-        debug_log_handler.setFormatter(logging.Formatter('%(asctime)s %(levelname)s (%(thread)d) %(message)s'))
-        debug_log_handler.setLevel(logging.DEBUG)
-        self.logger.addHandler(debug_log_handler)
+        logutil.init_logging(log_level, self.debug_log_path)
```
locriandev, line 699: I don't see `init_logging` defined anywhere, you meant `setup_logging`?

elliott/elliottlib/runtime.py
```diff
@@ -277,37 +286,8 @@ def initialize_logging(self):
...
-        main_stream_handler.setLevel(log_level)
-        self.logger.addHandler(main_stream_handler)
-
-        self.debug_log_path = os.path.join(self.working_dir, "debug.log")
-        debug_log_handler = logging.FileHandler(self.debug_log_path)
-        # Add thread information for debug log
-        debug_log_handler.setFormatter(logging.Formatter('%(asctime)s %(levelname)s (%(thread)d) %(message)s'))
-        debug_log_handler.setLevel(logging.DEBUG)
-        self.logger.addHandler(debug_log_handler)
+        logutil.init_logging(log_level, self.debug_log_path)
```
locriandev, line 289: same as commented above for doozer, where is `init_logging` defined?

doozer/doozerlib/runtime.py
vfreex: Yes, I renamed this function before committing but forgot to update the callers.
Now updated.

artcommon/artcommonlib/logutil.py
vfreex: I've removed the return value. The root logger is rarely used for printing logs. In case of the root logger is needed, we can always call `logging.getLogger()`.

elliott/elliottlib/runtime.py
vfreex: I renamed it to setup_logging.

vfreex:
@locriandev Thanks for the review. I've addressed the issues you raised.

locriandev (commented):
/lgtm

PR #728, opened 2024-06-27
removed early-kernel-ship-ok tag

mbiarnes:
removed leftovers of: early-kernel-ship-ok tag


doozer/tests/cli/test_gen_payload.py
```diff
@@ -727,7 +727,7 @@ def test_rpm_deliveries(self):
         )
         ai.check_installed_rpms_in_image.side_effect = lambda dg_key, bi: {
             "foo": [],
-            "bar": [Mock(AssemblyIssue, code=AssemblyIssueCode.MISSING_SHIP_OK_TAG, component="bar", msg="")]
+            "bar": [Mock(AssemblyIssue, code=AssemblyIssueCode.MISSING_RHCOS_CONTAINER, component="bar", msg="")]
```
thegreyd, line 730: Better to delete this test, since it doesn't make sense with this assemblyissue

mbiarnes: test removed

PR #729, opened 2024-06-27
Revert "Revert "ART-9503 Handle multiple plashet host locations""

locriandev:
Reverts openshift-eng/art-tools#720


joepvd:
/approve
yet:
/hold
As I am not sure if we should mix current operational issue with this possibility of new errors

ashwindasr:
Lets wait till Monday

PR #730, opened 2024-06-28
ART-9555 update alert message for untag event

Ximinhan:
for untag event use alert message type and add art-attention tag


pyartcd/pyartcd/pipelines/tag_rpms.py
```diff
@@ -67,8 +68,7 @@ async def run(self):
                     message += f"To revert, run `brew untag {tag} {' '.join(nvrs)}`.\n"
                 if tagged:
                     message += "If you untag a build manually, it will not be re-tagged by this job again.\n\n"
-            if untagged or tagged:  # Don't spam release-artists if nothing changed
-                await self.slack_client.say(f":white_check_mark: Hi @release-artists ,\n{message}")
+                    await self.slack_client.say(f":white_check_mark: Hi @release-artists ,\n{message}")
```
joepvd, line 71: Think this does not need the direct ping.

joepvd (commented):
Awesome improvement! A request to stop pinging in the good case, think a message is enough.

PR #735, opened 2024-07-02
Refactor Runtime

vfreex:
Moving certain functionalities like resolving/cloning sources, semaphore, record.log, etc, into their separate modules.

This may need further tests on production.

I already made some tests in my hackspace like [yuxzhu/job/aos-cd-builds/job/build%252Focp4_scan/30](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/hack/job/yuxzhu/job/aos-cd-builds/job/build%252Focp4_scan/30/console) and [yuxzhu/job/aos-cd-builds/job/build%252Focp4/6](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/hack/job/yuxzhu/job/aos-cd-builds/job/build%252Focp4/6/console), hopefully it is not too risky to merge.


doozer/doozerlib/git_helper.py
```diff
@@ -0,0 +1,66 @@
+import logging
```
locriandev, line 1: This module could be useful tool wide? Maybe it could be part of `artcommonlib`

doozer/doozerlib/lock.py
```diff
@@ -0,0 +1,30 @@
+from multiprocessing import Semaphore
```
locriandev, line 1: This could also be part of `artcommonlib`

doozer/doozerlib/git_helper.py
vfreex: This is a standard library. https://github.com/openshift-eng/art-tools/pull/725 makes it possible to log with standard logging library to print logs to both stderr and debug.log. I would prefer to directly use standard library until we find a use case that certain code pattern is reusable.

doozer/doozerlib/lock.py
vfreex: makes sense. I will move it to artcommonlib.

vfreex:
@locriandev I moved lock.py to artcommonlib

doozer/doozerlib/git_helper.py
locriandev: I was actually referring to the `git_helper` module. Since this seems to offer general purpose functionalities, we might benefit from it in other places (e.g. elliott, pyartcd)

locriandev (commented):
/lgtm

PR #736, opened 2024-07-02
Init elliott debug.log

thegreyd:
Follow up to https://github.com/openshift-eng/art-tools/pull/725
Without a valid debug.log path, we log debug messages to console.
So init this as how we do it for doozer.


locriandev (commented):
/lgtm

PR #745, opened 2024-07-05
ART-9443 update extract_opm to support multi rhel binaries

Ximinhan:
after https://issues.redhat.com/browse/ART-9441 complete, the operator-framework has been added into payload in 4.16+, using similar logic to extract binaries from it in promote job

### DIFF
--- a/pyartcd/pyartcd/pipelines/promote.py
+++ b/pyartcd/pyartcd/pipelines/promote.py
@@ -743,13 +743,7 @@ class PromotePipeline:
         self.create_symlink(client_mirror_dir, False, False)
 
         # extract opm binaries
-        if (major > 4 or minor >= 16):
-            # from 4.16 opm has multi rhel binaries
-            _, operator_framework_tools = get_release_image_pullspec(pullspec, "operator-framework-tools")
-            self.extract_framework_tools(client_mirror_dir, release_name, operator_framework_tools, build_arch)
-        else:
-            _, operator_registry = get_release_image_pullspec(pullspec, "operator-registry")
-            self.extract_opm(client_mirror_dir, release_name, operator_registry, build_arch)
+        self.extract_opm(client_mirror_dir, release_name, pullspec, build_arch)
 
         util.log_dir_tree(client_mirror_dir)  # print dir tree
         util.log_file_content(f"{client_mirror_dir}/sha256sum.txt")  # print sha256sum.txt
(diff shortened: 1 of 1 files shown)


pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -726,8 +726,13 @@ async def publish_client(self, base_to_mirror_dir: str, pullspec, release_name,
         self.create_symlink(client_mirror_dir, False, False)
 
         # extract opm binaries
-        _, operator_registry = get_release_image_pullspec(pullspec, "operator-registry")
-        self.extract_opm(client_mirror_dir, release_name, operator_registry, build_arch)
+        if (major > 4 or minor >= 16):
```
joepvd, line 729: Can we instead test for the existence of this image, and extract when it does?

Ximinhan: when we add a new image in the payload it will always exist until it is deprecated, so I think there is no need to add an extra check for existence in each release.

vfreex, line 729:
```suggestion
        if (major, minor) >= (4, 16):
```

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -829,6 +834,29 @@ def extract_opm(self, client_mirror_dir, release_name, operator_registry, arch):
             with open(f"{client_mirror_dir}/sha256sum.txt", 'a') as f:  # write shasum to sha256sum.txt
                 f.write(f"{shasum}  opm-{platform}-{release_name}.tar.gz\n")
 
+    def extract_framework_tools(self, client_mirror_dir, release_name, operator_registry, arch):
```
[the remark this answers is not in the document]
Ximinhan: used to think of `extract_operator_framework_tools` but I think that name is too long so omit the operator to current `extract_framework_tools`

pyartcd/pyartcd/pipelines/promote.py
Ximinhan: I think they are no difference

Ximinhan:
/hold
wait for confirmation from https://issues.redhat.com/browse/ART-9441

pyartcd/pyartcd/pipelines/promote.py
joepvd: Yes, but we may have it in 4.17 while we do not in 4.16. It may or may not get backported to 4.15. Small change to not have us to worry about when to deploy this.

joepvd:
/hold
please work on review comments

pyartcd/pyartcd/pipelines/promote.py
Ximinhan: we should have clear control of what to extract for what release, if in 4.16 the framework-tools pullspec is broken we should not fallback to operator-registry

ashwindasr:
rebased PR with new changes from art-tools, so that we can use this PR for 4.17 EC prep. No other changes added.

pyartcd/pyartcd/pipelines/promote.py
thegreyd: I see it was updated, resolving this

thegreyd:
Let's rebase the pr since https://github.com/openshift-eng/art-tools/pull/796 merged

PR #754, opened 2024-07-08
[ART-6797] Allow release fields in standard assembly names

ashwindasr:
ref https://issues.redhat.com/browse/ART-6797


joepvd:
Think this is in progress elsewhere
/close

PR #755, opened 2024-07-09
Fix _is_pub_ancestor_of_priv

vfreex:
There is a bug in `_is_pub_ancestor_of_priv` which doesn't interpret the exit code of `git merge-base` command correctly.

https://git-scm.com/docs/git-merge-base says

> Check if the first <commit> is an ancestor of the second <commit>, and exit with status 0 if true, or with status 1 if not. Errors are signaled by a non-zero status that is not 1.


locriandev (approved):
/lgtm

PR #764, opened 2024-07-11
Do not attempt to move microshift advisory to QE

thegreyd:
We try to move microshift advisory to QE in prep and promote,
at which time it doesn't have any builds and always fails,
which leads to noisy logs and complains. So skip that.

### DIFF
--- a/elliott/elliottlib/cli/find_bugs_sweep_cli.py
+++ b/elliott/elliottlib/cli/find_bugs_sweep_cli.py
@@ -353,7 +353,7 @@ def categorize_bugs_by_type(bugs: List[Bug], advisory_id_map: Dict[str, int],
 
         for bug in tracker_bugs:
             package_name = bug.whiteboard_component
-            if package_name == "microshift" and len(packages) == 0:
+            if kind == "microshift" and package_name == "microshift" and len(packages) == 0:
                 # microshift is special since it has a separate advisory, and it's build is attached
                 # after payload is promoted. So do not pre-emptively complain
                 logger.info(f"skip attach microshift bug {bug.id} to {advisory} because this advisory has no builds attached")
--- a/pyartcd/pyartcd/pipelines/promote.py
+++ b/pyartcd/pyartcd/pipelines/promote.py
@@ -204,7 +204,7 @@ class PromotePipeline:
 
             if assembly_type == AssemblyTypes.STANDARD:
                 # Attempt to move all advisories to QE
-                tasks = []
+                tasks_with_args = []
                 for impetus, advisory in impetus_advisories.items():
                     # microshift advisory is special, and it will not be ready at this time
                     if impetus == "microshift":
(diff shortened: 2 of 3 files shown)


elliott/elliottlib/cli/find_bugs_sweep_cli.py
```diff
@@ -353,7 +353,7 @@ def categorize_bugs_by_type(bugs: List[Bug], advisory_id_map: Dict[str, int],
 
         for bug in tracker_bugs:
             package_name = bug.whiteboard_component
-            if package_name == "microshift" and len(packages) == 0:
+            if kind == "microshift" and package_name == "microshift" and len(packages) == 0:
```
thegreyd, line 356: :lady_beetle:

Ximinhan:
I have some thoughts,  instead of excluding the microshift in if, we can improve it by updating the change state function to look at if there are builds attached to the advisory, if no build then changes state don't return error but a no-ops state( no build there, change state operation won't do), our current issue is fixing the false alarm about microshift advisory is empty and expected, this may also affect empty rpm advisory. and in build-microshift job, it should handle the microshift advisory changed to QE.

Ximinhan:
This PR also looks good if your direction is don't look at microshift advisory at all, microshift job will handle it.

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -206,18 +205,23 @@ async def run(self):
...
-                for impetus, advisory in impetus_advisories.items():
+                sorted_advisories = sorted(impetus_advisories.items())
+                for impetus, advisory in sorted_advisories:
+                    # microshift advisory is special, and it will not be ready at this time
+                    if impetus == "microshift":
+                        continue
                     if not advisory or advisory <= 0:
                         continue
                     logger.info("Moving advisory %s to QE...", advisory)
-                    if not self.runtime.dry_run:
```
locriandev, line 213: Are we moving advisories to QE even in case of a dry run now?

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -206,18 +205,23 @@ async def run(self):
             if assembly_type == AssemblyTypes.STANDARD:
                 # Attempt to move all advisories to QE
                 tasks = []
-                for impetus, advisory in impetus_advisories.items():
+                sorted_advisories = sorted(impetus_advisories.items())
```
locriandev, line 208: Is there a specific reason to sort them?

elliott/elliottlib/cli/find_bugs_sweep_cli.py
locriandev, line 356: if advisory is of kind `microshift`, will the package always `microshift`? Is there a redundancy here?

pyartcd/pyartcd/pipelines/promote.py
thegreyd: the dry run is added in the change_state function based on self.dry_run

pyartcd/pyartcd/pipelines/promote.py
thegreyd: yeah, it is to enable reporting when we get `results = asyncio.gather(*tasks, return_exceptions=True)`. The results are returned in order of the tasks given, so if you see the subsequent code, we log if there is an exception based on `sorted_advisories[i]`

vfreex (commented):
/lgtm

thegreyd:
yes @Ximinhan the idea is that the microshift job should handle on microshift advisory and report on errors, prep and promote should skip it.

pyartcd/pyartcd/pipelines/promote.py
locriandev: So IIUC, `tasks` will not contain the change microshift advisory state. You're then getting less results then sorted advisories. Doesn't this make indexing the advisories with the same `i` you're getting by looping over the results wrong?

thegreyd: ah you're right, good catch

locriandev (commented):
/lgtm

PR #765, opened 2024-07-11
set the major and minor value as class value

Ximinhan:
we invoke isolate_major_minor_in_group(self.group) many times to get the major and minor value, optimize it to call once and the value can be used globally.


vfreex (changes requested):
No. I would be against this change.

major and minor are calculated from group. it could be defined as a property, but shouldn't be a field. Otherwise if people update group, they will be likely to update major and minor then it will cause a bug.

Do not do micro optimization unless you have an evidence that it could significantly improve the performance.

PR #766, opened 2024-07-11
[ART-10181] Associate RHEL CVEs to RHCOS

vfreex:
ProdSec starts to file issues with `pscomponent:rhcos` to signal an RHEL CVE is present in the rhcos image.

The `rhcos` pscomponent is special in that Brew has per-architecture component for RHCOS, such as `rhcos-x86_64`, `rhcos-aarch64`, etc.

This requires special handling in Elliott to associate those CVEs to those per-architecture Brew builds.


joepvd (approved):
/lgtm

PR #772, opened 2024-07-12
Fix resolve dockerfile name

thegreyd:
Right now images:streams prs open breaks because of this.
We don't need the absolute path of Dockerfile but just the filename, 
so we do not need to pass in source_path.


ashwindasr:
/lgtm

test rebase succeeded as well, on local

PR #778, opened 2024-07-15
Fix fallback logic

ashwindasr:
Follows https://github.com/openshift-eng/art-tools/pull/772, where we accidentally included a regression to the fallback logic. We need to fallback when the file does not exist, not the field itself.


ashwindasr:
rebase succeeded

PR #784, opened 2024-07-16
Allow multiple comments on pr

lgarciaaco:
When one github repository builds more than one image, allow every build to post a comment on pr. Fix misleading comments

An example on how the notifier looks like can be found [here](https://github.com/lgarciaaco/ovn-kubernetes/pull/9)

### DIFF
--- a/doozer/doozerlib/comment_on_pr.py
+++ b/doozer/doozerlib/comment_on_pr.py
@@ -39,7 +39,7 @@ class CommentOnPr:
         """
         issue_comments = self.list_comments()
         for issue_comment in issue_comments:
-            if f"Distgit: {self.distgit_name}" in issue_comment["body"]:
+            if  "[ART PR BUILD NOTIFIER]" in issue_comment["body"] and f"Distgit: {self.distgit_name}" in issue_comment["body"]:
                 return True
         return False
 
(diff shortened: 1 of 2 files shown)


doozer/doozerlib/comment_on_pr.py
```diff
@@ -39,23 +39,23 @@ def check_if_comment_exist(self):
         """
         issue_comments = self.list_comments()
         for issue_comment in issue_comments:
-            if "[ART PR BUILD NOTIFIER]" in issue_comment["body"]:
```
thegreyd, line 42: I would still keep this as a check against matching some user's comment.

thegreyd:
Nice work! lgtm with a minor comment

doozer/doozerlib/comment_on_pr.py
```diff
@@ -39,23 +39,23 @@ def check_if_comment_exist(self):
...
+            if  "[ART PR BUILD NOTIFIER]" in issue_comment["body"] and f"Distgit: {self.distgit_name}" in issue_comment["body"]:
```
thegreyd, line 42:
```suggestion
            if "[ART PR BUILD NOTIFIER]" in issue_comment["body"] and f"Distgit: {self.distgit_name}" in issue_comment["body"]:
```

PR #787, opened 2024-07-17
scan-sources: Always search images by el_target

thegreyd:
This fixes the issue of not properly finding builds for ci golang builder images (because they share brew package for el8 and el9 builds happening in parallel). It was happening because we were only fetching latest build (for any el version), since https://github.com/openshift-eng/art-tools/pull/402  and then filtering out builds that did not match `branch_el_target` later, which resulted in no builds.

This would result in both images (el8 and el9) fetching builds by the same pattern (not including el version) and 
one would randomly get it's latest and other one would not.

Now since all our nvrs include ".el" suffix in them (it's been a while since we introduced them), use it by default. 

Before:
```
./doozer --assembly=stream --working-dir ../../ --group openshift-4.16 \
                                                         --images=ci-openshift-golang-builder-latest.rhel9 \
                                                         --rpms=openshift config:scan-sources

ci-openshift-golang-builder-latest.rhel9 is changed (reason: Component ci-openshift-golang-builder-latest-container
 has no latest build for assembly stream and target None)
```

^ if `ci-openshift-golang-builder-latest.rhel9` doesn't give that result, try `ci-openshift-golang-builder-latest.rhel8`

After:
```
Running a change assessment on ci-openshift-golang-builder-latest-container-
v4.16.0-202407172305.p0.g2ed769e.assembly.stream.el9 built at event 59120310
```

### DIFF
--- a/doozer/doozerlib/metadata.py
+++ b/doozer/doozerlib/metadata.py
@@ -445,6 +445,8 @@ class Metadata(object):
                     el_ver = isolate_el_version_in_brew_tag(el_target)
                 if not el_ver:
                     raise ValueError(f'Unable to determine rhel version from specified el_target: {el_target}')
+            elif self.meta_type == 'image':
+                el_ver = self.branch_el_target()
 
             if self.meta_type == 'image':
                 ver_prefix = 'v'  # openshift-enterprise-console-container-v4.7.0-202106032231.p0.git.d9f4379
@@ -476,11 +478,16 @@ class Metadata(object):
                 raise IOError(msg)
 
             def latest_build_list(pattern_suffix):
-                # Include * after pattern_suffix to tolerate other release components that might be introduced later.
-                # Also include a .el<version> suffix to match the new build pattern
+                # Include * after pattern_suffix to tolerate other release components that might be introduced later
+                # we do not want "**" in the pattern so do some additional checks
+                main_pattern = f"{pattern_prefix}{extra_pattern}{pattern_suffix}"
+                if not main_pattern.endswith("*"):
+                    main_pattern = f"{main_pattern}*"
 
+                # include a .el<version> suffix to match the new build pattern
                 el_suffix = f'.el{el_ver}*' if el_ver else ''
-                pattern = f'{pattern_prefix}{extra_pattern}{pattern_suffix}*{el_suffix}'
+                pattern = f'{main_pattern}{el_suffix}'
+                self.logger.info(f'Searching latest build for pattern={pattern}, state={build_state.name}')
                 builds = koji_api.listBuilds(packageID=package_id,
                                              state=None if build_state is None else build_state.value,
                                              pattern=pattern,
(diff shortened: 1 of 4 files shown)


doozer/tests/test_metadata.py
```diff
@@ -131,7 +133,7 @@ def _list_builds(self, builds, packageID=None, state=None, pattern=None, queryOp
             refined = [build for build in refined if build['state'] == state]
 
         refined.sort(key=lambda e: e['creation_ts'], reverse=True)
-        return refined
+        return refined[:1]
```
thegreyd, line 136: we did not update this when we changed our brew query to return only 1 build, this updates that.

doozer/tests/test_metadata.py
```diff
@@ -188,22 +190,22 @@ def list_builds(packageID=None, state=None, pattern=None, queryOpts=None, **kwar
         self.assertEqual(meta.get_latest_build(default=None), builds[3])
 
         # Make sure that just matching the prefix of an assembly is not sufficient.
```
thegreyd, line 192: I looked and this case basically is not supported with the code we have right now ("ignore a very similar looking assembly's build"). We fallback on finding stream build then which isn't too bad but not ideal. Would categorize this as a regression and discuss what we can do to support this. Maybe go back to fetching multiple builds.

doozer/doozerlib/metadata.py
```diff
@@ -487,24 +498,6 @@ def latest_build_list(pattern_suffix):
                 # "assembly.howdy'.
                 refined = [b for b in builds if b['nvr'].endswith(pattern_suffix) or f'{pattern_suffix}.' in b['nvr']]
 
-                # .el? was added to images well after assemblies were established. It allows us to build multiple
```
thegreyd, line 490: this doesn't apply now since we are specifically searching by elX

doozer/tests/test_metadata.py
thegreyd: One idea is to strictly look for ending "." in ".assembly.<name>.elX" since now we always have "elX"

elliott/elliottlib/metadata.py
```diff
@@ -258,11 +265,15 @@ def default_return():
                 raise IOError(msg)
 
             def latest_build_list(pattern_suffix):
-                # Include * after pattern_suffix to tolerate other release components that might be introduced later.
-                # Also include a .el<version> suffix to match the new build pattern
+                # Include * after pattern_suffix to tolerate other release components that might be introduced later
+                # we do not want "**" in the pattern so check first
+                main_pattern = f"{pattern_prefix}{extra_pattern}{pattern_suffix}"
+                if main_pattern[-1] != "*":
```
jupierce, line 271: `not main_pattern.endswith('*')` reads a bit cleaner to me.

elliott/elliottlib/metadata.py
```diff
@@ -224,9 +224,16 @@ def get_latest_build(self, default: Optional[Any] = -1, assembly: Optional[str]
 
             el_ver = None
             if el_target:
-                el_ver = isolate_el_version_in_brew_tag(el_target)
+                if isinstance(el_target, int):
```
jupierce, line 227: I'd pull these checks into a more abstract utility function for finding el. e.g. `isolate_el_version` which can take strings, ints, and tag names.

elliott/elliottlib/metadata.py
thegreyd: thanks, updated

thegreyd:
I pushed a change to be strict about assembly suffix: https://github.com/openshift-eng/art-tools/pull/787/commits/f30580837c43c88c43be2573f76367921ba5808e . Imo it works better when we are asking only 1 build from brew, to support the case of distinguishing between similar assembly names. This part of code that is tricky and I feel it would be better to be strict so we notice missing builds soon if something is wrong/changes and can address that in code appropriately.

doozer/doozerlib/metadata.py
jupierce: This means we can no  longer assemble & fix older releases that didn't follow this convention. We should be prepared to run with an older commit of doozer if this occurs.

doozer/tests/test_metadata.py
```diff
@@ -72,7 +72,7 @@ def test_is_rpm_exempt(self):
 
     def build_record(self, creation_dt: datetime.datetime, assembly, name='foo-container',
                      version='4.7.0', p='p0', epoch=None, git_commit='4c0ed6d',
-                     release_prefix=None, release_suffix='',
+                     release_prefix=None, el_target=8,
```
jupierce, line 75: Suggesting that el_target be a required parameter. The default make it harder to reason about each individual test result without looking at the implementation of `build_record`.

jupierce:
Just a couple nits, but looks good!

doozer/tests/test_metadata.py
thegreyd: using this now

doozer/tests/test_metadata.py
thegreyd: done

PR #792, opened 2024-07-19
Don't set publish date if batch is provided

thegreyd:
This is to resolve the issue of advisory publish date field being set and stuck in the past,
when we update the associated batch publish date to something else. In such a case errata-tool library 
cannot perform operations on it

```
ElliottFatalError("Error finding or attaching bugs: [ErrataException('Erratum 136272: publish_date_override: cannot be
 set before 2024-Jul-24 which is determined by batch date of the advisory or ship_date of release\\n')]. See logs for 
more information.")
```

[slack_ref](https://redhat-internal.slack.com/archives/C06625LHXQS/p1721412549172059?thread_ts=1721411461.605679&cid=C06625LHXQS)


joepvd (approved):
/lgtm
/approve

PR #800, opened 2024-07-23
Only alert when advisory cannot move from new_files to qe

thegreyd:
We try and move advisories to QE and alert in slack if we fail. But 
sometimes we run prep/promote jobs when advisories in REL_PREP, or 
even SHIPPED_LIVE to do something else. In those cases we don't want to 
error and be alerted when we cannot move advisory from those advanced
states to QE. So introduce a `--from` flag in elliott change-state which indicates 
that we should not operate if advisory is not in that state. Then we use 
`--from NEW_FILES` in our promote and prep pipelines.


vfreex (commented):
/approve
/lgtm

PR #801, opened 2024-07-23
Don't set publish date if batch is provided

thegreyd:
Reverts openshift-eng/art-tools#795
Original PR https://github.com/openshift-eng/art-tools/pull/792

Original PR failed for https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fprepare-release/226/console , so includes the fix.

It also 
- sets the Prerelease advisory date to +3 days from prep.
- tries to clear batch when creating advisory to remove automatic ErrataTool batch association

### DIFF
--- a/pyartcd/pyartcd/pipelines/prepare_release.py
+++ b/pyartcd/pyartcd/pipelines/prepare_release.py
@@ -471,10 +471,11 @@ class PrepareReleasePipeline:
             f"--assigned-to={self.runtime.config['advisory']['assigned_to']}",
             f"--manager={self.runtime.config['advisory']['manager']}",
             f"--package-owner={self.package_owner}",
-            f"--date={release_date}",
         ]
         if batch_id:
             create_cmd.append(f"--batch-id={batch_id}")
+        else:
+            create_cmd.append(f"--date={release_date}")
         if not self.dry_run:
             create_cmd.append("--yes")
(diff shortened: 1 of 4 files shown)


thegreyd:
/hold
need to investigate and fix the error

pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -471,10 +483,11 @@ def create_advisory(self, advisory_type: str, art_advisory_key: str, release_dat
             f"--assigned-to={self.runtime.config['advisory']['assigned_to']}",
             f"--manager={self.runtime.config['advisory']['manager']}",
             f"--package-owner={self.package_owner}",
-            f"--date={release_date}",
         ]
         if batch_id:
             create_cmd.append(f"--batch-id={batch_id}")
+        else:
+            create_cmd.append(f"--date={release_date}")
```
Ximinhan, line 490: what if we want to overwrite the date on batch?

thegreyd: That's handled on line 178 https://github.com/openshift-eng/art-tools/blob/55a21647aabd19ae611dc95f13ec8880be6d9a2b/pyartcd/pyartcd/pipelines/prepare_release.py#L178 this command is not used for it

PR #806, opened 2024-07-24
Fix olm bundle command init

thegreyd:
When invoking the command "olm-bundle:rebase-and-build" to build one nvr 
I noticed that the runtime initialization was taking a really long time (>10 minutes). This 
was before it even got to the bundle building part. This is because we were cloning all 
image distgits during init regardless of which NVRs were passed for building. 
This behavior isn't correct and wastes a lot of time when running this command locally 
so fix that. In case nvrs are given, fetch image names from nvrs and then init runtime 
with only those images.

### DIFF
--- a/doozer/doozerlib/cli/olm_bundle.py
+++ b/doozer/doozerlib/cli/olm_bundle.py
@@ -1,13 +1,18 @@
 import click
 import sys
 import traceback
+import koji
+from urllib.parse import urlparse
 
-from artcommonlib import exectools
+from artcommonlib import exectools, logutil
 from artcommonlib.model import Missing
-from doozerlib import Runtime
+from artcommonlib.constants import BREW_HUB
+from doozerlib import Runtime, brew
 from doozerlib.cli import cli, pass_runtime
 from typing import Optional, Tuple
 from doozerlib.olm.bundle import OLMBundle
+
+LOGGER = logutil.get_logger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


doozer/doozerlib/cli/olm_bundle.py
```diff
@@ -140,18 +145,37 @@ def rebase_and_build_olm_bundle(runtime: Runtime, operator_nvrs: Tuple[str, ...]
...
+    operator_builds = []
+    if operator_nvrs:
+        LOGGER.info("Fetching given nvrs from brew...")
+        brew_session = koji.ClientSession(BREW_HUB)
+        images = []
+        for index, build in enumerate(brew.get_build_objects(operator_nvrs, brew_session)):
+            if not build:
+                raise IOError(f"Build {operator_nvrs[index]} doesn't exist in Brew.")
+            source_url = urlparse(build['source'])
+            image_name = source_url.path.rstrip('/').rsplit('/')[-1]
```
vfreex, line 160: This is a little bit concerning... But it is fine unless the assumption breaks.

vfreex (commented):
/lgtm

doozer/doozerlib/cli/olm_bundle.py
thegreyd: I agree, it is not perfect. I copied that part from how we are determining it in OLMBundle class, so I think it has been working without breaking for all operator images. I will look if I can replace it with something more robust in another PR.

PR #756, opened 2024-07-09
Re-apply Runtime refactoring with fixes for cloning public upstream repo

vfreex:
The 4th commit in this PR fixes an issue that the public upstream repo is not cloned if `get_public_upstream` returns a None public branch.

PR #811, opened 2024-07-30
[golang builder] rename yaml safe_load to load

ashwindasr:
- Use `yaml.load` since its from the `YAML` package
- Set default value for `has_public_upstream` (follow up from https://github.com/openshift-eng/art-tools/pull/756)

### DIFF
--- a/doozer/doozerlib/source_resolver.py
+++ b/doozer/doozerlib/source_resolver.py
@@ -160,6 +160,7 @@ class SourceResolver:
             clone_branch, _ = self.detect_remote_source_branch(source_details, self.stage, use_source_fallback_branch)
 
             url = str(source_details["url"])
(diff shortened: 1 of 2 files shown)


doozer/doozerlib/source_resolver.py
```diff
@@ -160,6 +160,7 @@ def resolve_source(self, meta: 'Metadata') -> SourceResolution:
             clone_branch, _ = self.detect_remote_source_branch(source_details, self.stage, use_source_fallback_branch)
 
             url = str(source_details["url"])
+            has_public_upstream = False
             if self._group_config.public_upstreams:
```
ashwindasr, line 164: Edge case, since `self._group_config.public_upstreams` wasn't true, `has_public_upstream` wasn't being set.

PR #812, opened 2024-07-31
[ART-9920] add new cve type field

Ximinhan:
ref [ART-9920](https://issues.redhat.com/browse/ART-9920)
Prodsec will start creating new format of CVE bug with type `Vulnerability`, and `Vulnerability` type of bug has some useful customfield that we can use to simplify our check

the example bug has these fields is [OCPBUGS-36952](https://issues.redhat.com/browse/OCPBUGS-36952)

* customfield_12324749 --> `CVE ID` field, we can use this field to get cve id, no need to do string match
* customfield_12324751 --> `Upstream Affected Component` field, we can use this field to get pscomponent
* customfield_12324750 --> `Embargo Status` field, we can use this field as flag to check embargo status

### DIFF
--- a/elliott/elliottlib/bzutil.py
+++ b/elliott/elliottlib/bzutil.py
@@ -87,9 +87,6 @@ class Bug:
     def is_ocp_bug(self):
         raise NotImplementedError
 
-    def is_vnlnerability_cve(self):
-        raise NotImplementedError
-
     @property
     def component(self):
         raise NotImplementedError
@@ -220,9 +217,6 @@ class BugzillaBug(Bug):
     def is_ocp_bug(self):
         return self.product == constants.BUGZILLA_PRODUCT_OCP
 
-    def is_vnlnerability_cve(self):
-        return False
-
     def creation_time_parsed(self):
         return datetime.strptime(str(self.bug.creation_time), '%Y%m%dT%H:%M:%S').replace(tzinfo=timezone.utc)
 
@@ -265,7 +259,7 @@ class JIRABug(Bug):
             return None
 
     def is_tracker_bug(self):
-        if self.is_vnlnerability_cve():
+        if self.is_type_vulnerability():
             return True
         has_keywords = set(constants.TRACKER_BUG_KEYWORDS).issubset(set(self.keywords))
         has_whiteboard_component = bool(self.whiteboard_component)
@@ -366,7 +360,7 @@ class JIRABug(Bug):
 
     @property
     def cve_id(self):
-        if self.is_vnlnerability_cve():
+        if self.is_type_vulnerability():
             return getattr(self.bug.fields, JIRABugTracker.field_cve_id)
         if not (self.is_tracker_bug() or self.is_flaw_bug()):
             return None
@@ -393,7 +387,7 @@ class JIRABug(Bug):
         :returns: a string if a value is found, otherwise None
         """
         if self.is_vnlnerability_cve():
-            return getattr(self.bug.fields, JIRABugTracker.field_cve_upstream_component)
+            return getattr(self.bug.fields, JIRABugTracker.field_cve_component)
         markers = [r'^art:pscomponent:\s*(\S+)', r'^pscomponent:\s*(\S+)']
         for label in self.bug.fields.labels:
             for marker in markers:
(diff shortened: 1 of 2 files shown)


elliott/elliottlib/bzutil.py
```diff
@@ -97,6 +97,8 @@ def product(self):
 
     @property
     def cve_id(self):
+        if self.bug.fields.issuetype.name == "Vulnerability":
```
thegreyd, line 100: Would like to have a method for this check, something like self.is_type_vulnerability()

elliott/elliottlib/bzutil.py
```diff
@@ -259,6 +261,8 @@ def security_level(self):
             return None
 
     def is_tracker_bug(self):
+        if self.bug.fields.issuetype.name == "Vulnerability":
+            return True
```
thegreyd, line 265: Let's not do this.. we have learned from experience that bugs can be created with missing metadata, so let's verify all metadata that we need in this new type and return True only when all of them are present

elliott/elliottlib/cli/find_bugs_golang_cli.py
```diff
@@ -317,6 +317,8 @@ def is_valid(b: JIRABug):
             # Do not touch embargoed bugs
             if b.bug.fields.security.name == "Embargoed Security Issue":
                 return False
+            if self.bug.fields.issuetype.name == "Vulnerability" and getattr(self.bug.fields, JIRABugTracker.field_cve_is_embargo).value == "True":
+                return False
```
thegreyd, line 321: It would be nice to move this into a method like `bug.is_embargoed()`

elliott/elliottlib/bzutil.py
Ximinhan: we don't need a special function as this is already straightforward

elliott/elliottlib/bzutil.py
Ximinhan: the new "Vulnerability" is designed to can't be created or modified except ProdSec, so when we see such type we can trust it.

elliott/elliottlib/cli/find_bugs_golang_cli.py
Ximinhan: I was trying to check if there such is_embargoed() function but can't find, I am not sure our embargo manage process, right now I only find here, we can abstract to the function if we see other use places

elliott/elliottlib/bzutil.py
Ximinhan: maybe I misunderstand, if you are saying self.is_type_vulnerability() is to get true/false of self.bug.fields.issuetype.name == "Vulnerability" then it possible, but this place is already in the original Bug class, the function has to be in jirabug class

elliott/elliottlib/cli/find_bugs_golang_cli.py
Ximinhan: seems we don't need to change here because if new CVE bug is embargoed he security level will also be set to embargo level

elliott/elliottlib/bzutil.py
```diff
@@ -417,6 +438,9 @@ def creation_time_parsed(self):
     def is_ocp_bug(self):
         return self.bug.fields.project.key == "OCPBUGS" and not self.is_placeholder_bug()
 
+    def is_vnlnerability_cve(self):
```
thegreyd, line 441: would really like this to be `is_type_vulnerability`

elliott/elliottlib/bzutil.py
```diff
@@ -620,6 +644,9 @@ class JIRABugTracker(BugTracker):
     field_release_blocker = 'customfield_12319743'  # "Release Blocker"
     field_blocked_reason = 'customfield_12316544'  # "Blocked Reason"
     field_severity = 'customfield_12316142'  # "Severity"
+    field_cve_id = 'customfield_12324749'  # "CVE ID"
+    field_cve_upstream_component = 'customfield_12324751'  # "Upstream Affected Component"
```
thegreyd, line 648: should this be "downstream affected component" field instead? see https://issues.redhat.com/browse/OCPBUGS-37655 upstream is `ncurses`, downstream is `rhcos`

elliott/elliottlib/bzutil.py
```diff
@@ -87,6 +87,9 @@ def is_flaw_bug(self):
     def is_ocp_bug(self):
         raise NotImplementedError
 
+    def is_vnlnerability_cve(self):
```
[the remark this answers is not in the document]
Ximinhan: updated

elliott/elliottlib/bzutil.py
Ximinhan: updated

PR #814, opened 2024-08-01
[ART-10300] remove failed build's dependents from bundle build list

Ximinhan:
ref: [ART-10300](https://issues.redhat.com/browse/ART-10300)
This pr is to fix the issue when ocp4 job success build openshift-kubernetes-nmstate-operator but failed to build nmstate-console-plugin we should not trigger olm_bundle job for openshift-kubernetes-nmstate-operator because its referenced image failed to build.

Add image's dependents metadata into build record.
When generate the list of builds for olm_bumdle job, check if there are failed build exist in record, if true then don't trigger related olm_bundle build by removing its operator from build list to avoid build errors.

### DIFF
--- a/doozer/doozerlib/distgit.py
+++ b/doozer/doozerlib/distgit.py
@@ -1022,6 +1022,7 @@ class ImageDistGitRepo(DistGitRepo):
         owners = list(self.config.owners) if self.config.owners and self.config.owners is not Missing else []
         action = "build"
         release = self.org_release if self.org_release is not None else '?'
+        dependents = list(self.config.dependents) if self.config.dependents is not Missing else []
         record = {
             "dir": self.distgit_dir,
             "dockerfile": "%s/Dockerfile" % self.distgit_dir,
@@ -1036,6 +1037,7 @@ class ImageDistGitRepo(DistGitRepo):
             "status": -1,
             "push_status": -1,
             "has_olm_bundle": 1 if self.config['update-csv'] is not Missing else 0,
+            "dependents": ",".join(dependents),
             # Status defaults to failure until explicitly set by success. This handles raised exceptions.
         }
 
--- a/pyartcd/pyartcd/pipelines/ocp4.py
+++ b/pyartcd/pyartcd/pipelines/ocp4.py
@@ -784,11 +784,10 @@ class Ocp4Pipeline:
         for record in records:
             # If failed build has dependents, remove it operator bundle from list because olm_bundle job won't succeed
             if record['status'] != '0' and record.get('nvrs', None) and record.get('dependents', None):
-                distgit_nvrs = record['dependents'].split(', ')
-                for dep in distgit_nvrs:
+                for dep in record['dependents'].split(','):
                     for record in records:
                         if record['distgit'] == dep and record.get('nvrs', None):
-                            operator_nvrs.remove(record['nvrs'].split(', ')[0])
+                            operator_nvrs.remove(record['nvrs'].split(',')[0])
 
         await util.sync_images(
             version=self.version.stream,
--- a/pyartcd/pyartcd/pipelines/ocp4.py
+++ b/pyartcd/pyartcd/pipelines/ocp4.py
@@ -781,6 +781,14 @@ class Ocp4Pipeline:
         for record in records:
             if record['has_olm_bundle'] == '1' and record['status'] == '0' and record.get('nvrs', None):
                 operator_nvrs.append(record['nvrs'].split(',')[0])
+        for record in records:
+            # If failed build has dependents, remove it operator bundle from list because olm_bundle job won't succeed
+            if record['status'] != '0' and record.get('nvrs', None) and record.get('dependents', None):
+                distgit_nvrs = record['dependents'].split(', ')
+                for dep in distgit_nvrs:
+                    for record in records:
+                        if record['distgit'] == dep and record.get('nvrs', None):
+                            operator_nvrs.remove(record['nvrs'].split(', ')[0])
 
         await util.sync_images(
             version=self.version.stream,
--- a/pyartcd/pyartcd/pipelines/ocp4.py
+++ b/pyartcd/pyartcd/pipelines/ocp4.py
@@ -777,17 +777,17 @@ class Ocp4Pipeline:
 
         records = record_log.get('build', [])
         operator_nvrs = []
-
+        operator_nvrs_map = {}
+        operator_deps = []
         for record in records:
             if record['has_olm_bundle'] == '1' and record['status'] == '0' and record.get('nvrs', None):
                 operator_nvrs.append(record['nvrs'].split(',')[0])
-        for record in records:
-            # If failed build has dependents, remove it operator bundle from list because olm_bundle job won't succeed
             if record['status'] != '0' and record.get('nvrs', None) and record.get('dependents', None):
-                for dep in record['dependents'].split(','):
-                    for record in records:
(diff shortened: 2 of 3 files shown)


doozer/doozerlib/distgit.py
```diff
@@ -1022,6 +1022,7 @@ def build_container(
         owners = list(self.config.owners) if self.config.owners and self.config.owners is not Missing else []
         action = "build"
         release = self.org_release if self.org_release is not None else '?'
+        dependents = list(self.config.sependents) if self.config.dependents is not Missing else []
```
thegreyd, line 1025:
```suggestion
        dependents = list(self.config.dependents) if self.config.dependents is not Missing else []
```

pyartcd/pyartcd/pipelines/ocp4.py
```diff
@@ -781,6 +781,14 @@ async def _sync_images(self):
         for record in records:
             if record['has_olm_bundle'] == '1' and record['status'] == '0' and record.get('nvrs', None):
                 operator_nvrs.append(record['nvrs'].split(',')[0])
+        for record in records:
```
thegreyd, line 784: why a new loop, why not add this logic in the above loop?

Ximinhan: because we need first to get operator_nvrs list, then remove failed build, we don't know the order so it could be a successful child behind the failed parent so when do operator_nvrs.remove the nvr is not in the list yet

doozer/doozerlib/distgit.py
Ximinhan: fixed

pyartcd/pyartcd/pipelines/ocp4.py
```diff
@@ -781,6 +781,14 @@ async def _sync_images(self):
...
+            # If failed build has dependents, remove it operator bundle from list because olm_bundle job won't success
```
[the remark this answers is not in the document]
Ximinhan: fixed

pyartcd/pyartcd/pipelines/ocp4.py
```diff
@@ -781,6 +781,14 @@ async def _sync_images(self):
         for record in records:
             if record['has_olm_bundle'] == '1' and record['status'] == '0' and record.get('nvrs', None):
                 operator_nvrs.append(record['nvrs'].split(',')[0])
+        for record in records:
+            # If failed build has dependents, remove it operator bundle from list because olm_bundle job won't succeed
+            if record['status'] != '0' and record.get('nvrs', None) and record.get('dependents', None):
+                distgit_nvrs = record['dependents'].split(', ')
```
vfreex, line 787: This var contains a list of distgit keys instead of nvrs. Better to use a different name.

Ximinhan: updated

pyartcd/pyartcd/pipelines/ocp4.py
joepvd: Think your comment is about the new loop in line 788. The loop in line 784 seems not needed?

Ximinhan: updated in https://github.com/openshift-eng/art-tools/pull/814/commits/37f4ee437aab25852342f21f79c978916b784352

joepvd:
This just moves the problem to later. 

At prepare release time, we will find the operator build, and then trigger a bundle build to match the operator. This means waiting time of osbs is congested, only to conclude "it is failing" but later and at a time that we do not want to wait for things. 

So the "bundle build failed" messages are a better earlier notification for when this is broken than "prepare release failed". 

The bigger task ahead of us in operator metadata builds is the move to Konflux/FBC. Not sure how invasive these changes really are, and if changes in the current bundle process will survive the new logic. @vfreex, could you comment?

pyartcd/pyartcd/pipelines/ocp4.py
```diff
@@ -777,10 +777,17 @@ async def _sync_images(self):
...
         records = record_log.get('build', [])
         operator_nvrs = []
-
+        operator_nvrs_map = {}
+        operator_deps = []
         for record in records:
             if record['has_olm_bundle'] == '1' and record['status'] == '0' and record.get('nvrs', None):
                 operator_nvrs.append(record['nvrs'].split(',')[0])
+            if record['status'] != '0' and record.get('nvrs', None) and record.get('dependents', None):
+                operator_deps.extend([dep for dep in record['dependents'].split(',') if dep not in operator_deps])
```
thegreyd, line 786: use a set, it will deduplicate:
`depenents.update(record['dependents'].split(','))`

pyartcd/pyartcd/pipelines/ocp4.py
```diff
@@ -777,10 +777,17 @@ async def _sync_images(self):
...
+            operator_nvrs_map[record['distgit']] = record.get('nvrs')
```
thegreyd, line 787: why is this called operator_nvrs_map, since it is storing every image's (operator or not) distgit -> nvrs ? 
Something like `distgit_to_nvr_map` makes it obvious ?

pyartcd/pyartcd/pipelines/ocp4.py
thegreyd: also we are storing failed dependents, so the name should reflect that `failed_dependents`

pyartcd/pyartcd/pipelines/ocp4.py
```diff
@@ -777,10 +777,17 @@ async def _sync_images(self):
 
         records = record_log.get('build', [])
         operator_nvrs = []
-
+        operator_nvrs_map = {}
+        operator_deps = []
         for record in records:
             if record['has_olm_bundle'] == '1' and record['status'] == '0' and record.get('nvrs', None):
                 operator_nvrs.append(record['nvrs'].split(',')[0])
+            if record['status'] != '0' and record.get('nvrs', None) and record.get('dependents', None):
```
thegreyd, line 785: If an image fails building, does it have nvrs defined?

pyartcd/pyartcd/pipelines/ocp4.py
thegreyd: We are storing failed_images_dependents so the name should reflect that ..

pyartcd/pyartcd/pipelines/ocp4.py
```diff
@@ -777,10 +777,17 @@ async def _sync_images(self):
...
+                operator_deps.extend([dep for dep in record['dependents'].split(',') if dep not in operator_deps])
+            operator_nvrs_map[record['distgit']] = record.get('nvrs')
+        for dep in operator_deps:
+            if dep in operator_nvrs_map and operator_nvrs_map[dep] in operator_nvrs:
```
thegreyd, line 789: This code is very hard to read. I understand the intent but would suggest to make it clear as to what we're doing. 

I would suggest to first go through records to collect all `failed_dependent_images` , and then have a second loop through records to collect `operator_nvrs` based on the above. This way we do not have to `.remove()`

thegreyd:
@joepvd I will throw my 2 cents here. I have been confused before as to why bundle builds would fail. The problem is we would trigger brew tasks for a bundle which would have a reference of an image which does not exist (based on the uuid). This is quite confusing, it's buried deep within logs, it's not apparent without looking at logs if the bundle failed because of missing operator build or it's a genuine failure. So I like this being catched earlier. I agree with your prep time concern, so we can raise some more alarm bells, like art-attention emoji etc.

PR #816, opened 2024-08-01
increase timeout for microshift build

ashwindasr:
Malware scanning takes a random amount of time to complete, sometimes it finishes quickly, other times it may take a couple of hours


ashwindasr:
/close

Might be better to automate the malware scanning refetch

PR #819, opened 2024-08-02
[fips scan] [ART-10376] periodically scan all latest builds

ashwindasr:
- Add the `--all-images` flag to the doozer `images:scan-fips` command
- Update artcd to to run for all active ocp4 versions
- Remove trigger from ocp4

Follow up PR will be raised after this has been deployed on cluster, to setup the scheduled job, which in turn will trigger the pipeline. Right now, these changes are only add the "scan all latest images" part and remove the code from ocp4 which calls the pipeline after every successful run.

### DIFF
--- a/doozer/doozerlib/cli/scan_fips.py
+++ b/doozer/doozerlib/cli/scan_fips.py
@@ -4,6 +4,7 @@ For this command to work, https://github.com/openshift/check-payload binary has 
 import asyncio
 import json
 import os
+import re
 from typing import Optional
 
 import click
@@ -15,9 +16,10 @@ from artcommonlib.exectools import cmd_gather_async, limit_concurrency, cmd_gath
 
 
 class ScanFipsCli:
-    def __init__(self, runtime: Runtime, nvrs: Optional[list], clean: Optional[bool]):
+    def __init__(self, runtime: Runtime, nvrs: Optional[list], all_images: Optional[bool], clean: Optional[bool]):
         self.runtime = runtime
-        self.nvrs = nvrs
+        self.nvrs = nvrs if nvrs else []
+        self.all_images = all_images
         self.clean = clean
         self.could_not_clean = []
         self.is_root = False
@@ -26,11 +28,20 @@ class ScanFipsCli:
         self.runtime.initialize(clone_distgits=False)
         self.koji_session = koji.ClientSession(self.runtime.group_config.urls.brewhub)
 
+        if self.nvrs and self.all_images:
+            raise Exception("Cannot specify both --nvrs and --all-images")
+
     async def execute_and_handle_cmd(self, cmd, nvr):
         rc, out, _ = cmd_gather(cmd)
         if rc != 0:
             self.could_not_clean.append(nvr)
         return rc, out
+
+    @staticmethod
+    async def clean_all_images():
+        rc, _, _ = cmd_gather("podman image prune --all --force")
+        if rc != 0:
+            raise Exception("Could not clean all images")
 
     async def clean_image(self, nvr, pull_spec):
         if not self.clean:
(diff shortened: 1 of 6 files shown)


doozer/doozerlib/cli/scan_fips.py
```diff
@@ -26,12 +28,21 @@ def __init__(self, runtime: Runtime, nvrs: Optional[list], clean: Optional[bool]
...
         rc, out, _ = cmd_gather(cmd)
         if rc != 0:
             self.could_not_clean.append(nvr)
         return rc, out
 
+    @staticmethod
+    async def clean_all_images():
+        rc, _, _ = cmd_gather("podman image prune --all --force")
+        if rc != 0:
+            raise Exception("Could not clean all images")
```
vfreex, line 44: If `cmd_assert_*` is used, there will be no need to check the return value.
Also this exception message doesn't include the original error message from the executed command, which could be hard to debug in case something goes wrong.

doozer/doozerlib/cli/scan_fips.py
```diff
@@ -72,7 +83,53 @@ async def am_i_root(self):
...
 
+    def get_tagged_latest(self, tag):
+        """
+        Returns the latest RPMs and builds, triggered by the automation, tagged in to the brew tag received as input
+        """
+        latest_tagged = self.koji_session.listTagged(tag=tag, latest=True, owner="exd-ocp-buildvm-bot-prod")
+        if latest_tagged:
+            return latest_tagged
+        else:
+            return []
```
vfreex, line 94:
```suggestion
        return latest_tagged or []
```

pyartcd/pyartcd/pipelines/scan_fips.py
```diff
@@ -13,42 +13,45 @@
 
 
 class ScanFips:
-    def __init__(self, runtime: Runtime, version: str, nvrs: Optional[list]):
+    def __init__(self, runtime: Runtime, nvrs: Optional[list]):
         self.runtime = runtime
-        self.version = version
         self.nvrs = nvrs
+        self.active_ocp_versions = [4.12, 4.13, 4.14, 4.15, 4.16, 4.17]
```
vfreex, line 19: This could be hard to discover when we have a new minor version. Thinking this should be controlled in a more visible config option.

pyartcd/pyartcd/pipelines/scan_fips.py
```diff
@@ -13,42 +13,45 @@
...
+        results = {}
 
-        _, result, _ = await exectools.cmd_gather_async(cmd, stderr=True)
+        for version in self.active_ocp_versions:
+            cmd = [
+                "doozer",
+                "--group",
+                f"openshift-{version}",
+                "--data-path",
+                "https://github.com/openshift-eng/ocp-build-data",
```
vfreex, line 34: This shouldn't be hardcoded here. We either use the default value, or let the user to override through the job parameter.

vfreex (commented):
Thanks. This generally looks good. I add some review suggestions for some nits. But I don't think there are blockers for this PR.

doozer/doozerlib/cli/scan_fips.py
```diff
@@ -72,7 +83,53 @@ async def am_i_root(self):
...
+
+    def get_all_latest_builds(self):
+        brew_tags = []
+        # Setup brew tags
+        tags = self.runtime.get_errata_config()["brew_tag_product_version_mapping"].keys()
```
Ximinhan, line 99: you can use
`brew_tags = self.runtime.gitdata.load_data(key='erratatool', replace_vars=runtime.group_config.vars).data.get('brew_tag_product_version_mapping', {}).keys()`

doozer/doozerlib/cli/scan_fips.py
```diff
@@ -72,7 +83,53 @@ async def am_i_root(self):
...
+        for tag in tags:
+            major, minor = self.runtime.get_major_minor_fields()
+            brew_tags.append(tag.format(MAJOR=major, MINOR=minor))
+        self.runtime.logger.info(f"Retrieved candidate tags: {brew_tags}")
+
+        tag_rhel_mapping = {}
```
Ximinhan, line 105: This `tag_rhel_mapping` is not used?

pyartcd/pyartcd/pipelines/scan_fips.py
ashwindasr: Yeah I'm not sure where I should keep this. Will move to artcommonlib for now. If we need to change it somewhere else, we can do so later

pyartcd/pyartcd/pipelines/scan_fips.py
ashwindasr: well we don't really need a param for this, but I guess I can set a default in the pipeline run / pipeline task

doozer/doozerlib/cli/scan_fips.py
ashwindasr: Can I know what the difference is? Is the value is already in `get_errata_config` then we can use it directly right?

ashwindasr: Nice, thanks for pointing that out!

doozer/doozerlib/cli/scan_fips.py
ashwindasr: yeah it was a remnant from the SAST pipeline, forgot to remove. Thank you

PR #838, opened 2024-08-13
Add importPolicy PreserveOriginal for image tags

thegreyd:
Multi nightlies on release controller aren't preserving the original manifest list, just the x86 component.

### DIFF
--- a/doozer/doozerlib/cli/release_gen_payload.py
+++ b/doozer/doozerlib/cli/release_gen_payload.py
@@ -991,7 +991,7 @@ class GenPayloadCli:
         The release payload images referenced in the multi-release payload manifest list are
         themselves somewhat standard release payloads (i.e. they are based on CVO images for their
         arch) BUT, each component image they reference is a manifest list. For example, the `cli`
-        image in the s390x release payload will point to a a manifest list composed of cli image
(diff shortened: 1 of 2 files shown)


doozer/doozerlib/cli/release_gen_payload.py
```diff
@@ -1655,6 +1658,9 @@ def build_payload_istag(payload_tag_name: str, payload_entry: PayloadEntry) -> D
             "from": {
                 "kind": "DockerImage",
                 "name": payload_entry.dest_pullspec,
+            },
+            "importPolicy": {
+                "importMode": "PreserveOriginal"
```
thegreyd, line 1663: I'm assuming this would not have sideffect when the image is not a manifest list

doozer/doozerlib/cli/release_gen_payload.py
```diff
@@ -1309,6 +1309,9 @@ def is_accepted(tag):
                 "referencePolicy": {
                     "type": "Source"
                 },
+                "importPolicy": {
+                    "importMode": "PreserveOriginal"
```
thegreyd, line 1313: Also not sure if this is needed / would have any effect

PR #842, opened 2024-08-15
[ART-10376] Fix latest builds retrieval in FIPS scan

adobes1:
Previously, when retrieving the latest builds in a version for FIPS scan, doozer retrieved latest build per component per tag. This PR introduces logic to retrieve only the latest build for each component across all tags.

### DIFF
--- a/doozer/doozerlib/cli/scan_fips.py
+++ b/doozer/doozerlib/cli/scan_fips.py
@@ -5,6 +5,7 @@ import asyncio
 import json
 import os
 from typing import Optional
+from datetime import datetime
 
 import click
 import koji
@@ -96,9 +97,18 @@ class ScanFipsCli:
         for tag in brew_tags:
             builds += self.get_tagged_latest(tag=tag)
 
-        nvrs = []
+        # Find the latest build nvr per image across all tags
+        latest_build_map = {}
         for build in builds:
-            nvrs.append(build["nvr"])
+            image_name = build["package_name"]
+            completion_time = datetime.strptime(build["completion_time"], "%Y-%m-%d %H:%M:%S.%f")
+
+            if image_name not in latest_build_map or completion_time > latest_build_map[image_name]["time"]:
+                latest_build_map[image_name] = {"nvr": build["nvr"], "time": completion_time}
+
+        nvrs = []
+        for latest_build in latest_build_map.values():
+            nvrs.append(latest_build["nvr"])
 
         return nvrs
 


doozer/doozerlib/cli/scan_fips.py
```diff
@@ -96,9 +97,18 @@ def get_all_latest_builds(self):
         for tag in brew_tags:
             builds += self.get_tagged_latest(tag=tag)
 
-        nvrs = []
+        # Find the latest build nvr per image across all tags
+        latest_build_map = {}
         for build in builds:
-            nvrs.append(build["nvr"])
+            image_name = build["name"]
```
ashwindasr, line 103: I think `build["package_name"]` could be a better key

doozer/doozerlib/cli/scan_fips.py
```diff
@@ -96,9 +97,18 @@ def get_all_latest_builds(self):
...
+            completion_time = datetime.strptime(build["completion_time"], "%Y-%m-%d %H:%M:%S.%f")
+
+            if image_name not in latest_build_map or completion_time > latest_build_map[image_name]["time"]:
+                latest_build_map[image_name] = {"nvr": build["nvr"], "time": completion_time}
```
ashwindasr, line 107: Since we are not using or printing out the `time`field, we can exclude them from being stored in the dict itself right

doozer/doozerlib/cli/scan_fips.py
```diff
@@ -96,9 +97,18 @@ def get_all_latest_builds(self):
...
+
+        nvrs = []
+        for latest_build in latest_build_map.values():
+            nvrs.append(latest_build["nvr"])
```
ashwindasr, line 111: So if we don't store the time, we can reduce these lines to just

> return list(latest_build_map.values())

doozer/doozerlib/cli/scan_fips.py
adobes1: It's used to compare the `completion_time` of the current build and the previously stored newest build. Or am I misunderstanding what you mean?

ashwindasr: yes - but after its being compared, its being stored int he dict right?

`latest_build_map[image_name] = {"nvr": build["nvr"], "time": completion_time}`

since the `time` field is not used after the comparison, do we need to store it unless we're using it in some way?

ashwindasr: ohh my bad - I missed the `latest_build_map[image_name]["time"]` part

doozer/doozerlib/cli/scan_fips.py
adobes1: Thank you, should be fixed now

doozer/doozerlib/cli/scan_fips.py
```diff
@@ -96,9 +97,18 @@ def get_all_latest_builds(self):
...
-            nvrs.append(build["nvr"])
+            image_name = build["package_name"]
+            completion_time = datetime.strptime(build["completion_time"], "%Y-%m-%d %H:%M:%S.%f")
+
+            if image_name not in latest_build_map or completion_time > latest_build_map[image_name]["time"]:
+                latest_build_map[image_name] = {"nvr": build["nvr"], "time": completion_time}
+
+        nvrs = []
+        for latest_build in latest_build_map.values():
+            nvrs.append(latest_build["nvr"])
```
ashwindasr, line 111:
```suggestion
        return [latest_build["nvr"] for latest_build in latest_build_map.values()]
```

ashwindasr (approved):
/hold

Feel free to unhold if you prefer to merge it as is.

PR #843, opened 2024-08-15
Enable promote with upgrades_next edges

thegreyd:
Enable `oc adm release new --next [...]`

We do not in our normal process require populating a payload's "next" edges.
In normal flow, each release's "next" edges are the following release's "previous" edges
But in case we miss adding an edge in "previous" list, we can add it in a "next" list
Example: 4.13.a and 4.14.b are shipping together in a week. 4.14.b has 4.13.a in its "previous" list.
Due to new requirements we re-promote 4.13 which becomes 4.13.(a+1) 
4.14.b does not have 4.13.(a+1) in its "previous" list. 
So we need to add 4.14.b in 4.13.(a+1)'s "next" list, 
to make that upgrade path available to customers who will be on 4.13.(a+1),
without repromoting 4.14.

See: https://github.com/openshift-eng/ocp-build-data/pull/5260
https://redhat-internal.slack.com/archives/CJARLA942/p1723736306214179?thread_ts=1723710327.987479&cid=CJARLA942

https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fpromote-assembly/275/console

### DIFF
--- a/ocp-build-data-validator/validator/json_schemas/assembly_group_config.schema.json
+++ b/ocp-build-data-validator/validator/json_schemas/assembly_group_config.schema.json
@@ -151,6 +151,16 @@
       "$ref": "#/properties/upgrades"
     },
     "upgrades-": {},
+    "upgrades_next": {
+      "type": "string"
+    },
+    "upgrades_next!": {
+      "$ref": "#/properties/upgrades_next"
+    },
+    "upgrades_next?": {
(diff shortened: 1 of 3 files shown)


thegreyd:
/hold
Don't want to merge before team has looked at this.

vfreex (commented):
/lgtm
/approve

PR #853, opened 2024-08-19
[ART-10504] Doozer: Move some fields from ImageDistGitRepo to ImageMetadata

vfreex:
This PR intends to move some fields that are useful for both Distgit-Brew build and Konflux build, so that the rebase status can be shared regardless of the rebase method.


locriandev (approved):
/lgtm

PR #854, opened 2024-08-19
[ART-10504] Remove k_ocp4 job and doozer k:images:rebase verb

vfreex:
Those commands are for prototyping and demonstration. They will be replaced with a freash new implementation coming later.


locriandev (approved):
/lgtm

PR #857, opened 2024-08-19
[ART-10504]  Update cmd_assert_async and cmd_assert_async in artcommonlib

vfreex:
We have 2 different implementations of cmd_assert_async/cmd_gather_async in artcommonlib (previously moved from doozer) and pyartcd. Those 2 implementations have slight differences in their parameter names and log handling and create confusions.

This PR updates cmd_assert_async and cmd_assert_async in artcommonlib to use the pyartcd's implementation. Pyartcd's implementation is much clean, and its parameter names are almost identical to the wrapped `asyncio.subprocess.create_subprocess_exec` function. Furthermore, it also handles OpenTelemetry trace propagation.

Note 2 arguments are removed: retries, strip. Those are not commonly used and we can use tenacity for retries.

### DIFF
--- a/artcommon/artcommonlib/exectools.py
+++ b/artcommon/artcommonlib/exectools.py
@@ -1,4 +1,5 @@
 import asyncio
+import contextvars
 import errno
 import functools
 import os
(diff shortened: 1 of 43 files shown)


locriandev (approved):
/lgtm

PR #858, opened 2024-08-19
[ART-10504] Add helper function SourceResolver.get_source_dir

vfreex:
This function is to get the upstream source directory for a component.


locriandev (approved):
/lgtm

PR #820, opened 2024-08-06
[ART-10504] Doozer: Add command for rebasing images with Konflux

vfreex:
This PR adds Konflux rebase feature to Doozer. To rebase with Konflux, run `doozer -g openshift-x.y beta:images:konflux:rebase`.  This PR also contains a lot of refactoring code. We have to get this well tested before merging it to main.

The major part of this feature is in doozerlib/backend/rebaser.py. Most of the code there is taken from distgit.py with refactor, aiming to have a backend-independent implementation for the rebase logic.

/hold
This PR requires 
https://github.com/openshift-eng/art-tools/pull/854
https://github.com/openshift-eng/art-tools/pull/853
https://github.com/openshift-eng/art-tools/pull/855
https://github.com/openshift-eng/art-tools/pull/856
https://github.com/openshift-eng/art-tools/pull/857
https://github.com/openshift-eng/art-tools/pull/858

Note: Note: Build info database for Konflux builds haven't been not implemented at this moment. So currently it can only build base images and child images inheriting from parent images on Brew/OSBS.

PR #859, opened 2024-08-19
WIP: Add Doozer verb to build images with Konflux

vfreex:


doozer/doozerlib/backend/konflux_image_builder.py
```diff
@@ -0,0 +1,365 @@
...
+    def _new_component(name: str, application: str, component_name: str,
+                       image_repo: Optional[str], source_url: Optional[str], revision: Optional[str]) -> dict:
+        obj = {
+            "apiVersion": "appstudio.redhat.com/v1alpha1",
+            "kind": "Component",
+            "metadata": {
+                "name": name,
+                "annotations": {
+                    "build.appstudio.openshift.io/pipeline": '{"name":"docker-build","bundle":"latest"}',
+                }
```
ashwindasr, line 184:
```
mintmaker.appstudio.redhat.com/disabled=true
```
Need to add this annotation in, just to be safe. ref [thread](https://redhat-internal.slack.com/archives/GDBRP5YJH/p1724260588819889?thread_ts=1723123622.362349&cid=GDBRP5YJH)

ashwindasr (changes requested):

doozer/doozerlib/backend/konflux_image_builder.py
```diff
@@ -0,0 +1,365 @@
...
+
+
+class KonfluxImageBuildError(Exception):
+    def __init__(self, message: str, pipelinerun_name: str, pipelinerun: Optional[resource.ResourceInstance]) -> None:
+        super().__init__(message)
+        self.pipelinerun_name = pipelinerun_name
+        self.pipelinerun = pipelinerun
+
+
+KONFLUX_BUNDLE_PULLSPEC = "quay.io/konflux-ci/tekton-catalog/pipeline-docker-build:f3ac40bbc0230eccb8d98a4d54dabd55a4943c5d"
```
ashwindasr, line 29: Looks like that SHA is pretty outdated. We probably need to figure out a process, but for now we can probably use the latest build always. But I see that there is `devel` so can we try something like

`quay.io/konflux-ci/tekton-catalog/pipeline-docker-build:devel`

doozer/doozerlib/backend/konflux_image_builder.py
```diff
@@ -0,0 +1,319 @@
...
+        app_name = self._config.group_name.replace(".", "-")
+        app_manifest = self._new_application(app_name, app_name)
+        app = await self._create_or_patch(dyn_client, app_manifest)
+        self._logger.info(f"Created application: {app['metadata']['name']}")
+
+        # Ensure the component resource exists
+        component_name = f"{app_name}-{metadata.distgit_key}"
+        dest_image_repo = self._config.output_repo
+        # dest_image_tag = f"{metadata.distgit_key}_{self._uuid_tag}"
+        dest_image_tag = f"{version}-{release}"
```
thegreyd, line 110: why not name-version-release?

doozer/doozerlib/backend/konflux_image_builder.py
```diff
@@ -0,0 +1,319 @@
...
+            "source_url": https_url,
+            "revision": commit_sha,
+            "target_branch": target_branch,
+            "git_auth_secret": git_auth_secret,
+
+        })
+        obj = yaml.load(rendered)
+        del obj["metadata"]["name"]
+        del obj["metadata"]["namespace"]
+        del obj["metadata"]["annotations"]["pipelinesascode.tekton.dev/on-cel-expression"]
```
thegreyd, line 234: curious what's the reason behind this / or why do we have this in yaml in the first place if we don't want them?

doozer/doozerlib/backend/konflux_image_builder.py
```diff
@@ -0,0 +1,319 @@
...
+            await exectools.to_thread(api.get, name=name, namespace=namespace)
+        except exceptions.NotFoundError:
+            found = False
+        if self._config.dry_run:
+            if found:
+                self._logger.warning(f"[DRY RUN] Would have patched {api_version}/{kind} {namespace}/{name}")
+            else:
+                self._logger.warning(f"[DRY RUN] Would have created {api_version}/{kind} {namespace}/{name}")
+            return resource.ResourceInstance(dyn_client, manifest)
+        if found:
```
thegreyd, line 267:
```suggestion
        if found:
            self._logger.info(f"Patching {api_version}/{kind} {namespace}/{name}")
```

doozer/doozerlib/backend/konflux_image_builder.py
```diff
@@ -0,0 +1,319 @@
...
+            new = await exectools.to_thread(
+                api.patch,
+                body=manifest,
+                namespace=namespace,
+                content_type="application/merge-patch+json",
+            )
+        else:
```
thegreyd, line 274:
```suggestion
        else:
            self._logger.info(f"Creating {api_version}/{kind} {namespace}/{name}")
```

doozer/doozerlib/backend/rebaser.py
```diff
@@ -0,0 +1,1549 @@
...
+        if self._runtime.assembly:
+            sb.write(".assembly.")
+            sb.write(self._runtime.assembly)
+
+        el_ver = 0
+        try:
+            el_ver = metadata.branch_el_target()
+        except ValueError:
+            pass
+        if el_ver:
```
thegreyd, line 545: We've grown to depend on `.elX` suffix in our nvr names, so i think we should error out if it is not found instead of passing, so we can investigate

doozer/doozerlib/backend/rebaser.py
```diff
@@ -0,0 +1,1549 @@
...
+        self._add_build_repos(dfp)
+
+        self._reflow_labels(df_path)
+
+    def _add_build_repos(self, dfp: DockerfileParser):
+        # Populating the repo file needs to happen after every FROM before the original Dockerfile can invoke yum/dnf.
+        dfp.add_lines(
+            "\n# Start Konflux-specific steps",
+            "RUN mkdir -p /tmp/yum_temp; mv /etc/yum.repos.d/*.repo /tmp/yum_temp/ || true",
+            "COPY .oit/signed.repo /etc/yum.repos.d/",
```
thegreyd, line 718: I think we want to change this to be signed/unsigned depending on the version 

https://github.com/openshift-eng/art-tools/blob/24fd4652d8ef4b803b28c21c3d0d1f9aa6d21340/doozer/doozerlib/osbs2_builder.py#L161

doozer/doozerlib/backend/rebaser.py
thegreyd: but can also do in a separate PR, so upto you if you wanna make this change here

doozer/doozerlib/constants.py
```diff
@@ -29,6 +29,12 @@
 KONFLUX_REPO_CA_BUNDLE_TMP_PATH = "/tmp"
 KONFLUX_REPO_CA_BUNDLE_FILENAME = "Current-IT-Root-CAs.pem"
 KONFLUX_REPO_CA_BUNDLE_HOST = "https://certs.corp.redhat.com/certs"
+WORKING_SUBDIR_KONFLUX_BUILD_SOURCES = "konflux_build_sources"
```
ashwindasr, line 32: Noticed that these constants aren't being used? Or not sure if I missed something or it'll be in an another PR.

doozer/doozerlib/backend/konflux_image_builder.py
vfreex: You are right. Initially by default config each image gets an separate repo. Since we are going to use a mono repo for all images, we need to include the image name.

doozer/doozerlib/backend/konflux_image_builder.py
vfreex: In the future the PipelineRun template file (https://github.com/openshift-eng/art-tools/pull/859/files#diff-f1894aed6cfd6dbe45cd9701265dbf035239886d026b4dca0c40865f5e8511fe) will be loaded from an external repo, which includes some fields that are specific to a single image. We'd better to remove them.

vfreex: I'll add a comment here

doozer/doozerlib/backend/rebaser.py
vfreex: Right. This part was taken from existing code and I didn't change it. Let's make a change in future PRs.

doozer/doozerlib/backend/rebaser.py
vfreex: This part was taken from original Ashwin's PR. Let's address this in a separate PR.

doozer/doozerlib/constants.py
vfreex: right, it is used in https://github.com/openshift-eng/art-tools/pull/820

vfreex:
@thegreyd @ashwindasr Thanks for the review. I've addressed your review suggestions. Please have another look.
Note previously this PR includes some commits from the rebase PR https://github.com/openshift-eng/art-tools/pull/820, so it was a bit confusing. For those review suggestions, I will open another PR to address them.

PR #874, opened 2024-08-20
Fix _thread_does_image_need_change async call

locriandev:
Hoping to fix `RuntimeError: await wasn't used with future` as seen [here](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4_scan/65059/console)

### DIFF
--- a/doozer/doozerlib/cli/scan_sources.py
+++ b/doozer/doozerlib/cli/scan_sources.py
@@ -434,7 +434,7 @@ class ConfigScanSources:
                 loop = asyncio.new_event_loop()
                 asyncio.set_event_loop(loop)
                 try:
-                    return asyncio.run(image_meta.does_image_need_change(
+                    return loop.run_until_complete(image_meta.does_image_need_change(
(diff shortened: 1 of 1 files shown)


vfreex:
It doesn't seem related. According to the stack trace, loop.run_until_complete is internally used in asyncio.run:

```python
  File "/mnt/jenkins-workspace/aos-cd-builds_build_ocp4_scan@7/art-tools/doozer/doozerlib/cli/scan_sources.py", line 437, in _thread_does_image_need_change
    return asyncio.run(image_meta.does_image_need_change(
  File "/usr/lib64/python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
```

locriandev:
@vfreex yeah this might not be it. Reading around, find discussions about a similar issue in `aiohttp`. I didn't find a clear answer/solution, although many suggested it might be an issue with cpython itself. We're currently using python3.9 on buildvm, might be worth trying with 3.11?

PR #883, opened 2024-08-26
ART-10562 Run scheduled ocp4-scan in pyartcd

locriandev:


pyartcd/pyartcd/pipelines/ocp4_scan.py
```diff
@@ -26,18 +26,10 @@ def __init__(self, runtime: Runtime, version: str, data_path: Optional[str] = No
...
         self._doozer_working = self.runtime.working_dir / "doozer_working"
         self.locked = True  # True by default; if not locked, run() will set it to False
-        self.frozen = False
 
     async def run(self):
         # If we get here, lock could be acquired
         self.locked = False
-
-        # Check if automation is frozen for current group
-        if not await util.is_build_permitted(self.version, doozer_working=str(self._doozer_working)):
```
thegreyd, line 36: Why remove this?

pyartcd/pyartcd/pipelines/scheduled/schedule_ocp4_scan.py
```diff
@@ -0,0 +1,40 @@
...
+async def run_for(version: str, runtime: Runtime):
+    # Skip if locked
+    lock = Lock.BUILD
+    lock_name = lock.value.format(version=version)
+    lock_manager = LockManager.from_lock(lock)
+    try:
+        locked = await lock_manager.is_locked(lock_name)
+    finally:
+        await lock_manager.destroy()
+    if locked:
```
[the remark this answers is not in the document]
thegreyd: Ah my bad. I remember that this is scheduling ocp4_scan and it already skips if it is locked so we're just moving the logic

pyartcd/pyartcd/pipelines/scheduled/schedule_ocp4_scan.py
```diff
@@ -0,0 +1,40 @@
...
+
+from pyartcd import util, jenkins
+from pyartcd.cli import cli, click_coroutine, pass_runtime
+from pyartcd.locks import Lock, LockManager
+from pyartcd.runtime import Runtime
+
+
+async def run_for(version: str, runtime: Runtime):
+    # Skip if locked
+    lock = Lock.BUILD
```
thegreyd, line 13: I see we're locking on `"github-activity-lock-${cv}"` in https://github.com/openshift-eng/aos-cd-jobs/pull/4226/files , so this should be changed? If i understand the github-activity-lock is the least that's needed for scan-sources / rebase. Also ocp4_scan schedules rhcos build which is not dependent on build lock. So I think we should not ocp4_scan activities like priv rebase / rhcos builds on build lock

pyartcd/pyartcd/pipelines/scheduled/schedule_ocp4_scan.py
```diff
@@ -0,0 +1,40 @@
...
+        return
+
+    # Skip if frozen
+    if not await util.is_build_permitted(version, doozer_working=str(runtime.working_dir / "doozer_working-" / version)):
+        runtime.logger.info('[%s] Not permitted, skipping', version)
+        return
+
+    # Schedule scan
+    runtime.logger.info('[%s] Scheduling ocp4-scan', version)
+    jenkins.start_ocp4_scan(version=version)
```
thegreyd, line 31: There's a parent timeout of 60 mins https://github.com/openshift-eng/aos-cd-jobs/pull/4226/files . I'm thinking let's include that in the ocp4_scan job itself?

pyartcd/pyartcd/pipelines/scheduled/schedule_ocp4_scan.py
locriandev: IIRC, back when we had ocp4 in Groovy/aos-cd-jobs `github-activity-lock-{version}` was shared among ocp4 and ocp4-scan (and others). When ocp4 was moved to pyartcd, and with the adoption of Redis locks, we renamed its locks to `lock:build:{version}`. ocp4-scan is currently the only place where we still use the `github-activity-lock`s.

I agree that having RHCOS builds live under a distinct lock would make sense. Ideally, we might setup a `rhcos-scan` job that only deals with rhcos, and set it to share a common `lock:build:rhcos` lock with the `build/rhcos` pipeline.

Noting this as a follow-up, for the moment I'll just introduce a new `SCAN` Redis lock

pyartcd/pyartcd/pipelines/scheduled/schedule_ocp4_scan.py
locriandev: :+1:   Done [here](https://github.com/openshift-eng/aos-cd-jobs/pull/4226/commits/e38d0b10b6e2d3443d74731ecb8f8c675266e4da)

pyartcd/pyartcd/pipelines/scheduled/schedule_ocp4_scan.py
thegreyd: Hmm you're right. This might lead to some weird things, if I remember ocp4 job can also trigger scan sources which then can also rebase? If ocp4 job only relies on "lock:build" and ocp4_scan relies on "lock:scan" then we have a conflict

thegreyd: In my mind these are the distinct write operations we are doing that shouldn't conflict: scan-sources rebase, triggering rhcos build, ocp4 rebase&build

locriandev: > ocp4 job can also trigger scan-sources

`ocp4` does not trigger a ocp4-scan, it only triggers rhcos and sync-for-ci

locriandev: > In my mind these are the distinct write operations we are doing that shouldn't conflict: scan-sources rebase, triggering rhcos build, ocp4 rebase&build

scan-sources rebase deals with rebasing into -priv org, which has nothing to do with doozer rebase. The former only acts on GitHub, the latter pushes to distgit.

By changing the ocp4-scan lock to a new `SCAN` type, I simply purged the `github-activity-lock` which is no longer descriptive, and moved it to pyartcd with a new name. So far, no changes in locking behavior. Once we merge this, we can revisit the way we trigger rhcos (for example, by having a dedicated scan job) and see if we can optimize the locking logics

thegreyd: This is what I meant by ocp4 triggering scan-sources" https://github.com/openshift-eng/art-tools/blob/cb30548ded4d1dbf7f8711e203ae4bbbf0ce1db2/pyartcd/pyartcd/pipelines/ocp4.py#L946

thegreyd: but maybe i'm mistaken and ocp4 cannot do rebase (controlled by flag?)?

PR #884, opened 2024-08-26
elliott Bug sweep fixes

thegreyd:
If a tracker bug is included in assembly definition and it's build isn't found to be 
attached, it was erroneously being filtered out. Fix that.

### DIFF
--- a/elliott/elliottlib/bzutil.py
+++ b/elliott/elliottlib/bzutil.py
@@ -828,7 +828,7 @@ class JIRABugTracker(BugTracker):
 
     def filter_bugs_by_cutoff_event(self, bugs: Iterable, desired_statuses: Iterable[str],
                                     sweep_cutoff_timestamp: float, verbose=False) -> List:
-        dt = datetime.utcfromtimestamp(sweep_cutoff_timestamp).strftime("%Y/%m/%d %H:%M")
+        dt = datetime.fromtimestamp(sweep_cutoff_timestamp, timezone.utc).strftime("%Y/%m/%d %H:%M")
         val = ','.join(f'"{s}"' for s in desired_statuses)
         query = f"issue in ({','.join([b.id for b in bugs])}) " \
                 f"and status was in ({val}) " \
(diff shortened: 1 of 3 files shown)


elliott/elliottlib/bzutil.py
```diff
@@ -828,7 +828,7 @@ def attach_bugs(self, bugids: List, advisory_id: int = 0, advisory_obj: Erratum
 
     def filter_bugs_by_cutoff_event(self, bugs: Iterable, desired_statuses: Iterable[str],
                                     sweep_cutoff_timestamp: float, verbose=False) -> List:
-        dt = datetime.utcfromtimestamp(sweep_cutoff_timestamp).strftime("%Y/%m/%d %H:%M")
```
thegreyd, line 831: this is being deprecated

elliott/elliottlib/cli/find_bugs_sweep_cli.py
```diff
@@ -222,6 +224,9 @@ async def find_and_attach_bugs(runtime: Runtime, advisory_id, default_advisory_t
         green_prefix(f"Searching {bug_tracker.type} for bugs with status {statuses} and target releases: {tr}\n")
 
     bugs = await get_bugs_sweep(runtime, find_bugs_obj, brew_event, bug_tracker)
+    if not bugs:
```
thegreyd, line 227: return early so we skip unecessary ops and logs

elliott/elliottlib/cli/find_bugs_sweep_cli.py
```diff
@@ -376,25 +381,22 @@ def categorize_bugs_by_type(bugs: List[Bug], advisory_id_map: Dict[str, int],
                 logger.info(f"{kind} build found for #{bug.id}, {package_name} ")
                 found.add(bug)
                 bugs_by_type[kind].add(bug)
+            elif bug.id in permitted_bug_ids:
```
thegreyd, line 384: the actual fix

elliott/elliottlib/cli/find_bugs_sweep_cli.py
```diff
@@ -376,25 +381,22 @@ def categorize_bugs_by_type(bugs: List[Bug], advisory_id_map: Dict[str, int],
...
+                logger.info(f'Including tracker bug #{bug.id} since it is explicitly included in the assembly config')
+                found.add(bug)
+                bugs_by_type['image'].add(bug)
```
Ximinhan, line 387: if this bug is related to rpm should we add it to rpm group?

vfreex, line 387: Even if we add the tracker to the image advisory, later `elliott attach-cve-flaws` and `elliott verify-bugs` runs will also fail due to missing build. Maybe it is more reasonable to raise an error when this occurs?

PR #889, opened 2024-08-28
ocp4: never scan for changes, always build what's pinned

locriandev:


thegreyd:
/approve
/hold
should unhold when someone's around to keep an eye on pipeline

PR #905, opened 2024-09-04
ART-10624: added new fields for automatically created JIRAS

mbiarnes:


doozer/doozerlib/cli/images_streams.py
```diff
@@ -807,6 +807,8 @@ def search_issues(query):
             'labels': ['art:reconciliation', f'art:package:{image_meta.get_component_name()}'],
             'versions': [{'name': release_version}],  # Affects Version/s
             'customfield_12319940': [{'name': Model(runtime.gitdata.load_data(key='bug').data).target_release[-1]}],  # customfield_12319940 is Target Version in jira
+            'customfield_12320850': ['Release Note Not Required'],  # customfield_12320850 is Release Notes Type in JIRA
```
ashwindasr, line 810: `customfield_12320850` is a dict and not a list. How you can see this is by using the python JIRA API and dumping out the values of a field thats already set.

```
'customfield_12320850': {
    'self': 'https://issues.redhat.com/rest/api/2/customFieldOption/31862',
    'value': 'Release Note Not Required',
    'id': '31862',
    'disabled': False
  }
```
We do not need to pass in all the values, it should probably work with this
```suggestion
            'customfield_12320850': {'value': 'Release Note Not Required'},  # customfield_12320850 is Release Notes Type in JIRA
```

doozer/doozerlib/cli/images_streams.py
```diff
@@ -807,6 +807,8 @@ def search_issues(query):
...
+            'customfield_12317313': ['NA'],  # customfield_12317313 is Release Notes Text in JIRA
```
ashwindasr, line 811: I think this is the correct value, since the field type seems to be string
```suggestion
            'customfield_12317313': 'NA',  # customfield_12317313 is Release Notes Text in JIRA
```

ashwindasr (changes requested):

doozer/doozerlib/cli/images_streams.py
```diff
@@ -807,6 +807,8 @@ def search_issues(query):
             'labels': ['art:reconciliation', f'art:package:{image_meta.get_component_name()}'],
             'versions': [{'name': release_version}],  # Affects Version/s
             'customfield_12319940': [{'name': Model(runtime.gitdata.load_data(key='bug').data).target_release[-1]}],  # customfield_12319940 is Target Version in jira
+            'customfield_12320850': {'value': 'Release Note Not Required'},  # customfield_12320850 is Release Notes Type in JIRA
```
ashwindasr, line 810: /hold

have to make sure that this works, sometimes we need to add the id as well, along with the value

ashwindasr: tested, works

ashwindasr (approved):

ashwindasr:
lgtm

feel free to unhold when ready

mbiarnes:
unhold

PR #907, opened 2024-09-05
Add microshift plashet repo config

thegreyd:
To enable bootc image builds, we need microshift rpm available at build time in a plashet
See https://github.com/openshift-eng/ocp-build-data/pull/5345/files

Test run: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4/8011/console

https://ocp-artifacts.hosts.prod.psi.rdu2.redhat.com/pub/RHOCP/plashets/4.18/microshift/microshift-el9/latest/

### DIFF
--- a/pyartcd/pyartcd/plashets.py
+++ b/pyartcd/pyartcd/plashets.py
@@ -110,6 +110,22 @@ def plashet_config_for_major_minor(major, minor):
             "embargoed_tags": [f"rhaos-{major}.{minor}-rhel-7-embargoed"],
             "include_previous_packages": previous_packages,
         },
+        "rhel-8-server-microshift-rpms": {
+            "slug": "microshift-el8",
+            "tag": f"rhaos-{major}.{minor}-rhel-8-candidate",
+            "product_version": f"RHEL-8-OSE-{major}.{minor}",
+            "include_embargoed": False,
+            "embargoed_tags": [],
+            "include_previous_packages": [],
+        },
+        "rhel-9-server-microshift-rpms": {
+            "slug": "microshift-el9",
+            "tag": f"rhaos-{major}.{minor}-rhel-9-candidate",
+            "product_version": f"RHEL-9-OSE-{major}.{minor}",
+            "include_embargoed": False,
+            "embargoed_tags": [],
+            "include_previous_packages": [],
+        }
     }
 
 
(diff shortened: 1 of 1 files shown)


pyartcd/pyartcd/plashets.py
```diff
@@ -110,6 +110,22 @@ def plashet_config_for_major_minor(major, minor):
             "embargoed_tags": [f"rhaos-{major}.{minor}-rhel-7-embargoed"],
             "include_previous_packages": previous_packages,
         },
+        "rhel-8-server-microshift-rpms": {
```
ggiguash, line 113: We do not need microshift on rhel8, only on rhel9

pyartcd/pyartcd/plashets.py
```diff
@@ -110,6 +110,22 @@ def plashet_config_for_major_minor(major, minor):
...
+            "slug": "microshift-el8",
+            "tag": f"rhaos-{major}.{minor}-rhel-8-candidate",
+            "product_version": f"RHEL-8-OSE-{major}.{minor}",
+            "include_embargoed": False,
+            "embargoed_tags": [],
+            "include_previous_packages": [],
+        },
+        "rhel-9-server-microshift-rpms": {
+            "slug": "microshift-el9",
+            "tag": f"rhaos-{major}.{minor}-rhel-9-candidate",
```
vfreex, line 123: This will include all of our candidate builds, not only microshift.

thegreyd: Right! I noticed that, microshift rpm needs ocp dependencies so I only included those in the group.yml config https://github.com/openshift-eng/ocp-build-data/pull/5345/files#diff-005fe59037264b0a0c3005a1c55f7043bdfcfab6c740a26bf8d9cb9ea3750f88R524 . other rpms won't be picked up

joepvd: hmmm i think openshift-clients will come from non-stream assembly. 
I'd be good with just the microshift rpm in this repo. To actually install it, you should also enable the ose repo.

vfreex (commented):
/lgtm

pyartcd/pyartcd/plashets.py
thegreyd: To include microshift rpms for a named-assembly when we do `config:plashet --assembly named-assembly from-tags our-candidate-tag`, the microshift rpms + stream rpms are selected from the candidate tag. So basically this microshift plashet is a superset of ose plashet. That's why we're able to install microshift+ose-dependencies from this single plashet (also need fastdatapath repo for ovn).

thegreyd: The problem with enabling ose-repo was non-relevant packages like `kernel` were getting updated with `dnf update`

PR #911, opened 2024-09-06
ART-9803 Add Konflux db integration

locriandev:
- add `artcommonlib.konflux` package to model BigQuery records, and to provide a BigQuery client to write/fetch builds
- update runtimes to add a common `konflux_db` field in artcommonlib, so that both Doozer and Elliott can use it
- update Doozer to store successful Brew builds info in BigQuery
- define a stub for `elliott find-k-builds` verb

Test builds:

- https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/hack/job/dpaolell/job/test-konflux/10, that ran `elliott -g openshift-4.17 find-k-builds -k image`
- `doozer --assembly=test --group=openshift-4.16 --latest-parent-version --images=ironic images:build --repo-type=unsigned`, that stored an `ironic` build info in BigQuery. This can be verified on the GCP UI by executing this query:
```
SELECT * 
FROM `openshift-gce-devel.art_test.builds`
WHERE `build_id` = '3278295';
```

Depends on https://github.com/openshift-eng/aos-cd-jobs/pull/4244


artcommon/artcommonlib/runtime.py
```diff
@@ -1,13 +1,67 @@
...
+            log_level = logging.INFO
+
+        logutil.setup_logging(log_level, self.debug_log_path)
+        self._logger = logging.getLogger('artcommonlib')
+
+    def initialize_konxflux_db(self):
+        try:
+            self.konflux_db = KonfluxDb()
+        except Exception as err:
+            self._logger.warning('Cannot connect to the Konflux DB: %s\n%s', str(err), traceback.format_exc())
```
joepvd, line 52: We should not forget to not allow production builds to go unrecorded. But local experiments should be fine (like not record development work elsewhere).
Fine to not take care right now of this, but is this tracked somewhere?

locriandev: Filed [ART-10665](https://issues.redhat.com/browse/ART-10665) for when we're ready to enforce the BigQuery DB setup in Doozer

artcommon/artcommonlib/konflux/konflux_build_record.py
```diff
@@ -0,0 +1,155 @@
...
+    RPM = 'rpm'
+    IMAGE = 'image'
+
+
+class Engine(Enum):
+    KONFLUX = 'konflux'
+    BREW = 'brew'
+
+
+class KonfluxBuildRecord:
```
vfreex, line 31: Not required, but maybe it is more clear to use [dataclass](https://docs.python.org/3/library/dataclasses.html) for this type?

artcommon/artcommonlib/konflux/konflux_db.py
```diff
@@ -0,0 +1,155 @@
...
+        logging.getLogger('google.auth.transport.requests').setLevel(logging.WARNING)
+
+    def _check_env_vars(self):
+        """
+        Check if all required env vars are defined. Raise an exception otherwise
+        """
+
+        for var in self.REQUIRED_VARS:
+            if not os.getenv(var):
+                raise RuntimeError(f'Environment variable {var} must be defined')
```
vfreex, line 35: Here will be more convenient if we print all missing required env vars in one message.

artcommon/artcommonlib/runtime.py
```diff
@@ -1,13 +1,75 @@
...
+        self.initialized = False
+
+        self.konflux_db = None
+
+        for key, val in kwargs.items():
+            self.__dict__[key] = val
+
+    def initialize(self):
+        self.initialize_logging()
+        self.initialize_konxflux_db()
```
vfreex, line 33: Actually I didn't see a strong reason to make it a field in Runtime and initialize KonfluxDB for nearly all art-tools commands. I feel much better to initialize it when we need it, at least for now.

artcommon/artcommonlib/konflux/konflux_db.py
```diff
@@ -0,0 +1,155 @@
...
+            fields.append(f"`{param_name}`")
+
+        return f'({", ".join(fields)})'
+
+    def add_build(self, build: konflux_build_record.KonfluxBuildRecord):
+        """
+        Insert a build record into Konflux DB
+        """
+
+        query = f'INSERT INTO `{self.table_ref}` {self.column_names} VALUES '
```
vfreex, line 59: If possible, it is better to use parameterized queries to avoid SQL injection and make the code more readable. But I am ok with the current implementation for now.

https://cloud.google.com/bigquery/docs/parameterized-queries

vfreex (commented):
Thanks @locriandev . Generally LGTM. I added a few comments for some minor issues.

artcommon/artcommonlib/konflux/konflux_build_record.py
locriandev: This class would still need custom logic to assign `nvr`, `build_id` or `record_id` which can either:
- be left blank, and auto-generate them as needed
- provided in the constructor, as it would happen when building a BuildDataRecord object from a ResultRow (BigQuery result)

I'd rather leave it as-is not to make it overly complicated

artcommon/artcommonlib/runtime.py
locriandev: Doozer runtime has a `db` field [here](https://github.com/openshift-eng/art-tools/blob/main/doozer/doozerlib/runtime.py#L97). I followed the same pattern, by adding the new `konflux_db` field in the artcommon runtime. Initialization is still optional, as it happens only if we provide the required env vars (exactly the same way it currently happens for legacy DB in doozer). 

I don't have a strong opinions on this, but I generally prefer to have consistent patterns in the code.

artcommon/artcommonlib/konflux/konflux_db.py
locriandev: Done in 4ba453e0fb03aa8bb58c9b109d763db56dfcdde2

artcommon/artcommonlib/konflux/konflux_db.py
locriandev: I think it's doable, but needs time and care to implement in that `INSERT` statement. Doozer with ART's service account is the only actor who will be able to execute such statements, so I would not worry about SQL injection. Same applies for `SELECT` statements executed via the API (only Elliott with our service account will use that). We can still track this as a future enhancement

locriandev: I have created [ART-10667](https://issues.redhat.com/browse/ART-10667) as a investigation follow-up

artcommon/artcommonlib/logutil.py
```diff
@@ -55,6 +55,7 @@ def setup_logging(log_level: int, debug_log_path: str):
                         force=True)
 
     root_logger = logging.getLogger()
+    root_logger.setLevel(log_level)
```
vfreex, line 58: The log level is for stderr, which will be set at line 62.

locriandev: updated

doozer/doozerlib/distgit.py
```diff
@@ -1286,6 +1289,48 @@ def update_build_db(self, success_flag, task_id=None, scratch=False):
...
+        try:
+            source_repo, commitish = build_info['source'].split('#')
+
+            build_record = KonfluxBuildRecord(
+                name=self.metadata.name,
+                group=self.runtime.group,
+                version=build_info['version'],
+                release=build_info['release'],
+                assembly=self.runtime.assembly,
+                el_target=f'el{self.metadata.branch_el_target()}',
```
thegreyd, line 1310: I can think of cases where this is different from what's in the nvr, like in openshift rpm that builds for both el8 and el9. I think we're better off parsing it from the nvr.

thegreyd: ah I guess this is specific to an image

thegreyd: But imo it's better to rely on parsing it from nvr than branch_el_target, we don't want to run in an case where we build differently and our image config isn't updated or follows differently for konflux.

doozer/doozerlib/distgit.py
```diff
@@ -1286,6 +1289,48 @@ def update_build_db(self, success_flag, task_id=None, scratch=False):
...
+                source_repo=source_repo,
+                commitish=commitish,
+                embargoed=build_info['release'].split('.')[1] == 'p1',  # TODO
+                start_time=datetime.strptime(build_info['start_time'], '%Y-%m-%d %H:%M:%S.%f'),
+                end_time=datetime.strptime(build_info['completion_time'], '%Y-%m-%d %H:%M:%S.%f'),
+                artifact_type=ArtifactType.IMAGE,
+                engine=Engine.BREW,
+                image_tag=build_info['extra']['image']['index']['pull'][0],
+                outcome=success,
+                job_url=os.getenv('BUILD_URL', 'n/a'),
```
thegreyd, line 1323: maybe we store brew task url here? if we're planning to store the pipeline url for konflux, then it would be similar.

doozer/doozerlib/distgit.py
```diff
@@ -1286,6 +1289,48 @@ def update_build_db(self, success_flag, task_id=None, scratch=False):
...
+            build_record = KonfluxBuildRecord(
+                name=self.metadata.name,
+                group=self.runtime.group,
+                version=build_info['version'],
+                release=build_info['release'],
+                assembly=self.runtime.assembly,
+                el_target=f'el{self.metadata.branch_el_target()}',
+                arches=self.metadata.get_arches(),
+                rpms=[],  # TODO
+                parent_images=build_info['extra']['image']['parent_images'],
```
jupierce, line 1313: I think these should be NVR

doozer/doozerlib/distgit.py
```diff
@@ -1286,6 +1289,48 @@ def update_build_db(self, success_flag, task_id=None, scratch=False):
...
+
+            build_record = KonfluxBuildRecord(
+                name=self.metadata.name,
+                group=self.runtime.group,
+                version=build_info['version'],
+                release=build_info['release'],
+                assembly=self.runtime.assembly,
+                el_target=f'el{self.metadata.branch_el_target()}',
+                arches=self.metadata.get_arches(),
+                rpms=[],  # TODO
```
jupierce, line 1312: We will need this before we can implement ocp4_scan with these records. 
A better name would be `installed_packages` . `BrewBuildInspector.get_all_installed_package_build_dicts` can return these for a given image.

doozer/doozerlib/distgit.py
```diff
@@ -1286,6 +1289,48 @@ def update_build_db(self, success_flag, task_id=None, scratch=False):
...
+                parent_images=build_info['extra']['image']['parent_images'],
+                source_repo=source_repo,
+                commitish=commitish,
+                embargoed=build_info['release'].split('.')[1] == 'p1',  # TODO
```
jupierce, line 1316: `'p1' in build_info['release'].split('.')`   seems more forgiving.

doozer/doozerlib/distgit.py
locriandev: Done in 98db9d51324a513771f8f4ed06b44aca4c6d217d

doozer/doozerlib/distgit.py
locriandev: Done in 04f6e892b967742a3e7ca833ec6f017c1391fdb1

doozer/doozerlib/distgit.py
locriandev: Done in 70cd3fb9281aba8eb2936f36b7d6d228fc55daee

doozer/doozerlib/distgit.py
locriandev: Done in f7d71194840b46fd2c567f4c6fb9987becd90392

doozer/doozerlib/distgit.py
locriandev: Done in 405b372bfd82b0bc4f01e790787172ac29179731

doozer/doozerlib/distgit.py
```diff
@@ -1286,6 +1289,61 @@ def update_build_db(self, success_flag, task_id=None, scratch=False):
...
+                installed_packages=self.get_installed_packages(image_pullspec),
+                parent_images=[build['nvr'] for build in build_info['extra']['image']['parent_image_builds'].values()],
+                source_repo=source_repo,
+                commitish=commitish,
+                embargoed='p1' in build_info['release'].split('.'),
+                start_time=datetime.strptime(build_info['start_time'], '%Y-%m-%d %H:%M:%S.%f'),
+                end_time=datetime.strptime(build_info['completion_time'], '%Y-%m-%d %H:%M:%S.%f'),
+                artifact_type=ArtifactType.IMAGE,
+                engine=Engine.BREW,
+                image_tag=image_pullspec,
```
jupierce, line 1334: This is not being populated with a tag, but a full pullspec. 
I'm fine with storing that, but make an `image_pullspec` column. 
If you can image_tag to image_sha, you can then extract the `sha256:*` portion of the  pull spec.

doozer/doozerlib/distgit.py
```diff
@@ -1094,6 +1096,7 @@ def wait(n):
                         record["nvrs"] = build_info["nvr"]
                     if not dry_run:
                         self.update_build_db(True, task_id=task_id, scratch=scratch)
+                        self.update_konflux_db(build_info, KonfluxBuildOutcome.SUCCESS, job_url=task_url, scratch=scratch)
```
jupierce, line 1099: We should have two urls:
1. The ART job (Jenkins for the time being): `art_job_url`
2. The pipeline URL (brew today, konflux tomorrow): `build_pipeline_url`

doozer/doozerlib/distgit.py
```diff
@@ -1286,6 +1289,61 @@ def update_build_db(self, success_flag, task_id=None, scratch=False):
...
+
+        if not self.runtime.konflux_db:
+            self.logger.warning('Konflux DB connection is not initialized, not writing build record to the Konflux DB.')
+            return
+
+        image_pullspec = build_info['extra']['image']['index']['pull'][0]
+
+        self.logger.info('Storing build info in Konflux for Brew build %s', build_info['nvr'])
+        try:
+            source_repo, commitish = build_info['source'].split('#')
```
jupierce, line 1309: I think this needs to be the `openshift-priv` URL and commitish we rebased from. Distgit commitish doesn't inform any decisions as far as I know. `openshift-priv`'s commit, however, allows us to say "do we have an image build for github.com/openshift-priv/component-repo-name @ commit deadbeef" . This is a necessary insight for ocp4_scan.
We can also capture the information you are presently storing here, but I would call it `rebase_repo_url` and `rebase_commitish`.

doozer/doozerlib/distgit.py
locriandev: Done in 56aeef16c5c9442e469e0ca687f2a07b018f1539

doozer/doozerlib/distgit.py
locriandev: Done in 2df1af0cb9b2524338761c80edf43da13216fcd3

PR #924, opened 2024-09-12
add option to skip get_nightlies

ashwindasr:
- Add an option to skip `get_nightlies` i.e. skip nightly consistency check if necessary
- Small fix for slack dry-run option

Needs rebase if https://github.com/openshift-eng/art-tools/pull/925 merges first

### DIFF
--- a/pyartcd/pyartcd/pipelines/gen_assembly.py
+++ b/pyartcd/pyartcd/pipelines/gen_assembly.py
@@ -81,8 +81,6 @@ class GenAssemblyPipeline:
             raise ValueError(f"When using --skip-get-nightlies, nightlies for all given {len(self.arches)} arches must be specified")
 
     async def run(self):
-        if self.runtime.dry_run:
-            jenkins.update_title("[DRY_RUN]")
         self._slack_client.bind_channel(self.group)
         slack_response = await self._slack_client.say(
             f":construction: Generating assembly definition {self.assembly} :construction:")
--- a/pyartcd/pyartcd/pipelines/gen_assembly.py
+++ b/pyartcd/pyartcd/pipelines/gen_assembly.py
@@ -81,6 +81,8 @@ class GenAssemblyPipeline:
             raise ValueError(f"When using --skip-get-nightlies, nightlies for all given {len(self.arches)} arches must be specified")
 
     async def run(self):
+        if self.runtime.dry_run:
+            jenkins.update_title("[DRY_RUN]")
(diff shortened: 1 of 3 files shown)


pyartcd/pyartcd/pipelines/gen_assembly.py
```diff
@@ -76,6 +77,9 @@ def __init__(self, runtime: Runtime, group: str, assembly: str, data_path: str,
         self._doozer_env_vars["DOOZER_DATA_PATH"] = data_path if data_path else \
             self.runtime.config.get("build_config", {}).get("ocp_build_data_url")
 
+        if self.skip_get_nightlies and len(self.nightlies) != 4:
```
thegreyd, line 80:
```suggestion
        if self.skip_get_nightlies and len(self.nightlies) != len(self.arches):
```

pyartcd/pyartcd/pipelines/gen_assembly.py
```diff
@@ -76,6 +77,9 @@ def __init__(self, runtime: Runtime, group: str, assembly: str, data_path: str,
...
+            raise ValueError("When using --skip-get-nightlies, nightlies for all 4 arches must be specified")
```
thegreyd, line 81:
```suggestion
        if self.skip_get_nightlies and len(self.nightlies) != len(self.arches):
            raise ValueError(f"When using --skip-get-nightlies, nightlies for all given {len(self.arches)} arches must be specified")
```
we allow arches to be less than 4 so this would be accurate

pyartcd/pyartcd/pipelines/gen_assembly.py
```diff
@@ -88,8 +92,9 @@ async def run(self):
             if self.custom and (self.auto_previous or self.previous_list or self.in_flight):
                 raise ValueError("Specifying previous list for a custom release is not allowed.")
 
-            candidate_nightlies, latest_nightly = await asyncio.gather(
-                *[self._get_nightlies(), self._get_latest_accepted_nightly()])
+            if not self.skip_get_nightlies:
+                candidate_nightlies, latest_nightly = await asyncio.gather(
```
thegreyd, line 96: we would need to populate candidate nightlies for the else case?

thegreyd: probably just `candidate_nightlies = self.nightlies`

ashwindasr: thanks for catching that

PR #926, opened 2024-09-13
Konflux: Support building child images

vfreex:
Adds logic to build child images with Konflux. Note since DB integration hasn't been fully done, currently we have to build the parent image and the child image together.

Also changes the default image repo to
`quay.io/openshift-release-dev/ocp-v4.0-art-dev-test`, which can be customized using `--image-repo` parameter.

### DIFF
--- a/doozer/doozerlib/backend/konflux_image_builder.py
+++ b/doozer/doozerlib/backend/konflux_image_builder.py
@@ -37,7 +37,7 @@ class KonfluxImageBuilderConfig:
     kubeconfig: Optional[str] = None
     context: Optional[str] = None
     namespace: Optional[str] = None
-    output_repo: str = constants.KONFLUX_DEFAULT_DEST_IMAGE_REPO
+    image_repo: str = constants.KONFLUX_DEFAULT_IMAGE_REPO
     dry_run: bool = False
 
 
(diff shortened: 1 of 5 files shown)


doozer/doozerlib/backend/konflux_image_builder.py
```diff
@@ -105,8 +126,8 @@ async def _start_build(self, metadata: ImageMetadata, build_repo: BuildRepo, dyn
 
         # Ensure the component resource exists
         component_name = f"{app_name}-{metadata.distgit_key}"
-        dest_image_repo = self._config.output_repo
-        dest_image_tag = f"{metadata.get_component_name()}-{version}-{release}"
```
thegreyd, line 109: I'm curious about why we are changing this

vfreex: Some existing code (taken from distgit) uses the uuid tag to track images that are building together. I would like to keep that practice then consider to add additional tags for convenience (like encoding NVR in the tag) to have a minimal change for now.
Note the NVR tag is unknown until the .p0/.p1 flag is set.

thegreyd: hmm okay. But i think this is a temp change until we get get_latest_build() working, which might happen pretty soon? Or would you like to have this approach continue ?

PR #940, opened 2024-09-17
ART-10832: do not use reference nightlies for promotion

ashwindasr:
Use the generated image stream to promote and not the reference nightlies, if they are present in the assembly. The reference nighlties should only be used for validation.

i.e. we should not use the `--from-release` param in `oc adm release new`


thegreyd:
from what i remember this was for embargoed release. If we need this, let's bring it to art discuss or in team thread and discuss there.

ashwindasr:
@thegreyd we need this to change for our normal flow as well. Right now we promote the nightly directly, which is not desired. We should promote from the image streams from the CI cluster

vfreex (commented):
/lgtm
/approve

PR #941, opened 2024-09-17
Define konflux_db.get_latest_builds()

locriandev:
- create a separate abstraction layer to model BigQuery interactions in `bigquery.py`
- introduce `sqlalchemy` to let us easily construct higher level queries from Python objects
- define `get_latest_builds()`
- add unit tests for Konflux DB and BigQuery abstractions
- improve `elliott find-k-builds` stub

The `get_latest_builds()` implementation uses a scanning pattern that aims to limit as much as possible the amount of data that BigQuery needs to scan to satisfy our requests. The builds table has been partitioned by `start_time`, and the search function performs queries within 3-month wide time intervals of `start_time`. In the most frequent scenario, we would find a build for each component in the first window; for hotfix builds where we are dealing with old assembly definitions we would be moving backwards in time by shifting the search window.

Tested with:
```
elliott --group=openshift-4.18 find-k-builds --kind=image
```

### DIFF
--- a/artcommon/artcommonlib/bigquery.py
+++ b/artcommon/artcommonlib/bigquery.py
@@ -1,0 +1,112 @@
+import logging
+import os
+
+import typing
+
+from google.cloud import bigquery
+from google.cloud.bigquery.table import RowIterator
+from sqlalchemy import BinaryExpression, UnaryExpression
+from sqlalchemy.dialects import mysql
+
+
+class BigQueryClient:
+    REQUIRED_VARS = ['GOOGLE_CLOUD_PROJECT', 'GOOGLE_APPLICATION_CREDENTIALS', 'DATASET_ID', 'TABLE_ID']
+
+    def __init__(self):
+        self._check_env_vars()
+        self.client = bigquery.Client()
+        self.dataset_id = os.environ['DATASET_ID']
+        self._table_ref = f'{self.client.project}.{self.dataset_id}.{os.environ["TABLE_ID"]}'
+        self.logger = logging.getLogger(__name__)
+
+        # Make the gcp logger less noisy
+        logging.getLogger('google.auth.transport.requests').setLevel(logging.WARNING)
+
+    @property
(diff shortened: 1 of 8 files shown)


artcommon/artcommonlib/konflux/konflux_db.py
```diff
@@ -61,100 +48,134 @@ def add_build(self, build: konflux_build_record.KonfluxBuildRecord):
...
-
-        self.logger.info('Executing query: %s', query)
-        results = self.client.query(query).result()
-        self.logger.info('Found %s builds', results.total_rows)
-
-        for row in results:
-            yield konflux_build_record.from_result_row(row)
-
-    def search_builds_by_field(self, name, value):
+    def search_builds_by_fields(self, names: list, values: list, order_by: str = '', sorting: str = 'DESC',
```
jupierce, line 65: Wondering why you would would use a dict here?
```suggestion
    def search_builds_by_fields(self, where: Dict[str, Any] = None, order_by: str = '', sorting: str = 'DESC',
```
where `where is None` implies no constraint and `where["name"] is None` implies `name IS NULL` in SQL? Two lists creates the unnecessary possibilities of mismatched fields.

locriandev: Agreed, done in 1de6476e2a5f4604b26d8e293c5a465d7eed5dc1

artcommon/artcommonlib/bigquery.py
```diff
@@ -0,0 +1,112 @@
...
+    def query(self, query: str) -> RowIterator:
+        """
+        Execute a query in BigQuery and return a generator object with the results
+        """
+
+        self.logger.info('Executing query: %s', query)
+
+        try:
+            results = self.client.query(query).result()
+            self.logger.debug('Query returned %s result rows', results.total_rows)
```
joepvd, line 47: Would probably want to log the time it took for getting a result

PR #947, opened 2024-09-19
Doozer: Disallow extra 'v' in version string

vfreex:
This PR allows Doozer to raise an error if a version string like `--version=vv4.18` is passed through the command line.


joepvd (approved):
/lgtm

PR #948, opened 2024-09-19
ART-10831 Store failed Brew image builds in BigQuery

locriandev:
Storing failed builds in BigQuery will enable us to migrate images-health away from Brew API.

[Test build](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4/8851/console) stored:
```
SELECT * 
FROM `openshift-gce-devel.art_test.builds`
WHERE start_time > TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 24 HOUR)
AND `outcome` = 'failure'
```

### DIFF
--- a/artcommon/artcommonlib/konflux/konflux_db.py
+++ b/artcommon/artcommonlib/konflux/konflux_db.py
@@ -58,11 +58,21 @@ class KonfluxDb:
         Insert a build record into Konflux DB
         """
 
+        def value_or_null(value):
+            if value is None:
+                return 'NULL'
+
+            elif isinstance(value, str):
+                return f"'{value}'"
+
+            else:
+                return str(value)
+
         # Fill in missing record fields
         build.ingestion_time = datetime.now()
         build.schema_level = SCHEMA_LEVEL
         query = f'INSERT INTO `{self.table_ref}` {self.column_names} VALUES '
-        values = (f"'{value}'" if isinstance(value, str) else str(value) for value in build.to_dict().values())
+        values = (f"{value_or_null(value)}" for value in build.to_dict().values())
         query += '(' + ', '.join(values) + ')'
         self.logger.info('Executing query: %s', query)
 
--- a/artcommon/artcommonlib/konflux/konflux_db.py
+++ b/artcommon/artcommonlib/konflux/konflux_db.py
@@ -59,6 +59,12 @@ class KonfluxDb:
         """
 
         def value_or_null(value):
+            """
+            Return the value representation to be inserted into the query String.
+            Strings need to be quoted in '', other values (e.g. timestamps) are rendered without quotes.
+            None values translate into NULL
+            """
+
             if value is None:
                 return 'NULL'
 
--- a/doozer/doozerlib/distgit.py
+++ b/doozer/doozerlib/distgit.py
@@ -1314,72 +1314,59 @@ class ImageDistGitRepo(DistGitRepo):
 
         try:
             dfp = DockerfileParser(str(self.dg_path.joinpath('Dockerfile')))
-            source_repo = self.metadata.config.content.source.git.url
+            source_repo = convert_remote_git_to_https(self.metadata.config.content.source.git.url)
             commitish = dfp.labels['io.openshift.build.commit.url'].split('commit/')[-1]
 
             _, rebase_repo_url, _ = gather_git(['-C', self.distgit_dir, 'remote', 'get-url', 'origin'])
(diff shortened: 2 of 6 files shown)


artcommon/artcommonlib/konflux/konflux_db.py
```diff
@@ -58,11 +58,21 @@ def add_build(self, build: konflux_build_record.KonfluxBuildRecord):
         Insert a build record into Konflux DB
         """
 
+        def value_or_null(value):
+            if value is None:
+                return 'NULL'
+
+            elif isinstance(value, str):
+                return f"'{value}'"
```
thegreyd, line 66: we can just do `return value` here?

artcommon/artcommonlib/konflux/konflux_db.py
```diff
@@ -58,11 +58,21 @@ def add_build(self, build: konflux_build_record.KonfluxBuildRecord):
         Insert a build record into Konflux DB
         """
 
+        def value_or_null(value):
```
thegreyd, line 61: `stringify_val` or something like that might be a better name?

artcommon/artcommonlib/konflux/konflux_db.py
locriandev: The query string will need to enclose every value we want to store between `''`, if that value is a string. That's why this is needed.

artcommon/artcommonlib/konflux/konflux_db.py
locriandev: Here I am taking the value and casting to a proper value that BigQuery can digest. Strings need to be quoted in `''`, other values don't. For example, timestamps are rendered inside the query string without quotes. `None` values translate into NULL. `stringify_val` does not tell it all

thegreyd: ah okay! Then it makes sense. Can we include this in a comment? would make it clear why we're doing this

doozer/doozerlib/distgit.py
```diff
@@ -1306,50 +1312,74 @@ def update_konflux_db(self, build_info, outcome, build_pipeline_url='', scratch=
...
-                outcome=outcome,
-                art_job_url=os.getenv('BUILD_URL', 'n/a'),
-                build_pipeline_url=build_pipeline_url,
-                pipeline_commit='n/a',
-                nvr=build_info['nvr'],
-                build_id=str(build_info['id'])
-            )
+            if outcome == KonfluxBuildOutcome.FAILURE:
+                self.logger.info('Storing failed Brew build info for %s in Konflux DB', self.metadata.name)
+                build_record = KonfluxBuildRecord(
```
thegreyd, line 1331: Thinking it would be good to DRY this, so
```python
build_record_params = {name: self.metadata.name, ..  }  # all the common params
if outcome == success:
  build_record_params.update(..)
else:
  build_record_params.update(..)
build_record = KonfluxBuildRecord(**build_record_params)
```

doozer/doozerlib/distgit.py
```diff
@@ -1306,50 +1312,74 @@ def update_konflux_db(self, build_info, outcome, build_pipeline_url='', scratch=
...
-
-            image_pullspec = build_info['extra']['image']['index']['pull'][0]
+            _, rebase_repo_url, _ = gather_git(['-C', self.distgit_dir, 'remote', 'get-url', 'origin'])
+            _, rebase_commitish, _ = gather_git(['-C', self.distgit_dir, 'rev-parse', 'HEAD'])
 
             try:
                 release_target = self.org_release.split('.')[-1]
             except KeyError:
                 release_target = ''
             el_target = release_target if release_target else f'el{self.metadata.branch_el_target()}'
```
thegreyd, line 1327: This isn't part of the pr but would be nice to use `isolate_el_version_in_release` here

artcommon/artcommonlib/konflux/konflux_db.py
locriandev: Done in 9717adcbe18487dc4954490de5fbc4f6201be5bd

doozer/doozerlib/distgit.py
locriandev: Done in fcd7faf8481e280412233b8051c9d22ff251b722

doozer/doozerlib/distgit.py
locriandev: Done in fcd7faf8481e280412233b8051c9d22ff251b722

joepvd (approved):
/lgtm

PR #955, opened 2024-09-20
[konflux] skip-checks by default, replaces dots with dashes

ashwindasr:
Changes in this PR:
- Skip checks by default for now (since its failing the pipeline runs)
- Fix issue with successful / failed builds
- Replace dots with dashes, since openshift doesn't support dots

### DIFF
--- a/doozer/doozerlib/backend/konflux_image_build_pipelinerun.yaml
+++ b/doozer/doozerlib/backend/konflux_image_build_pipelinerun.yaml
@@ -25,6 +25,8 @@ spec:
     value: quay.io/redhat-user-workloads/crt-nshift-art-pipeli-tenant/ui-openshift-4-18/ui-ocp-4-18-base-rhel-9:{{revision}}
   - name: dockerfile
     value: Dockerfile
+  - name: skip-checks
+    value: "false"
   pipelineSpec:
     finally:
     - name: show-sbom
@@ -65,7 +67,7 @@ spec:
       description: Force rebuild image
       name: rebuild
       type: string
-    - default: "true"
+    - default: "false"
       description: Skip checks against built image
       name: skip-checks
       type: string
--- a/doozer/doozerlib/backend/konflux_image_build_pipelinerun.yaml
+++ b/doozer/doozerlib/backend/konflux_image_build_pipelinerun.yaml
@@ -65,7 +65,7 @@ spec:
       description: Force rebuild image
       name: rebuild
       type: string
-    - default: "false"
+    - default: "true"
       description: Skip checks against built image
       name: skip-checks
       type: string
--- a/doozer/doozerlib/backend/konflux_image_builder.py
+++ b/doozer/doozerlib/backend/konflux_image_builder.py
@@ -135,7 +135,8 @@ class KonfluxImageBuilder:
         self._logger.info(f"Using application: {app['metadata']['name']}")
 
         # Ensure the component resource exists
-        component_name = f"{app_name}-{metadata.distgit_key}"
+        # Openshift doesn't allow dots in any of its fields, so we replace them with dashes
+        component_name = f"{app_name}-{metadata.distgit_key}".replace(".", "-")
         dest_image_repo = self._config.image_repo
         dest_image_tag = df.envs["__doozer_uuid_tag"]
         default_revision = f"art-{self._config.group_name}-assembly-test-dgk-{metadata.distgit_key}"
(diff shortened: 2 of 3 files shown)


doozer/doozerlib/backend/konflux_image_builder.py
```diff
@@ -273,6 +274,8 @@ def _new_pipelinerun(generate_name: str, application_name: str, component_name:
         for param in obj["spec"]["params"]:
             if param["name"] == "output-image":
                 param["value"] = output_image
+            if param["name"] == "skip-checks":
+                param["value"] = "true"
```
ashwindasr, line 278: @vfreex this doesn't seem to be working, not sure if you have any idea?

doozer/doozerlib/backend/konflux_image_build_pipelinerun.yaml
```diff
@@ -65,7 +65,7 @@ spec:
       description: Force rebuild image
       name: rebuild
       type: string
-    - default: "false"
+    - default: "true"
```
ashwindasr, line 68: doing this worked, but ideally its better to not edit here

doozer/doozerlib/backend/rebaser.py
```diff
@@ -216,6 +216,14 @@ def _resolve_parents(self, metadata: ImageMetadata, dfp: DockerfileParser, image
         for builder in image_from.get("builder", []):
             parents.append(builder)
         parents.append(image_from)
+
+        # Remove duplicates
+        temp = []
+        for parent in parents:
+            if parent not in temp:
+                temp.append(parent)
+        parents = temp
```
ashwindasr, line 225: @vfreex Looks like one duplicate was getting into the list, so removed it, but if there is a better way we can refactor to that

ashwindasr: hmm actually this failed rebase for a couple of images (job [run](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-konflux/42/console))

doozer/doozerlib/backend/konflux_image_builder.py
```diff
@@ -220,6 +222,7 @@ def _new_component(name: str, application: str, component_name: str,
                 "name": name,
                 "annotations": {
                     "build.appstudio.openshift.io/pipeline": '{"name":"docker-build-multi-platform-oci-ta","bundle":"latest"}',
+                    "build.appstudio.openshift.io/status": '{"pac":{"state":"disabled"}}',
```
ashwindasr, line 225: Konflux by default seems to be enabling custom build pipelines by default, causing it to raise PRs (like [this](https://github.com/openshift-priv/oauth-proxy/pull/1)). Was looking at annotations, and this seem to be the difference between one that is enabled and one that is disabled.

doozer/doozerlib/backend/konflux_image_builder.py
vfreex: ship-checks parameter is not defined under `spec.params` (see line 19 of the yaml template). We can add it as in https://github.com/openshift-eng/art-tools/pull/955/files#diff-e45d70130344e3d45bd59cb091bd0a794121c675ce90f17c5b4e4702ded620ffR281

doozer/doozerlib/backend/rebaser.py
vfreex: @ashwindasr I checked the build log and looks like builders are added to the `parents` array twice. Removing line 216-217 should solve this.

vfreex (commented):
@ashwindasr Thanks for the PR.  After you address the issues I commented, feel free to merge.

doozer/doozerlib/backend/konflux_image_builder.py
ashwindasr: Thanks!

doozer/doozerlib/backend/konflux_image_build_pipelinerun.yaml
ashwindasr: Reverted

doozer/doozerlib/backend/rebaser.py
ashwindasr: Thanks!

ashwindasr:
[tested](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-konflux/45/console), works as expected

PR #967, opened 2024-09-24
deprecate upload_image

Ximinhan:
this function is not used now, it was added along with report-rhcos but that job is also deprecated now


joepvd (approved):
/lgtm

PR #968, opened 2024-09-24
Use the distgit key as primary image name when storing Konflux builds

locriandev:
Using `image_meta.name` is wrong, because we have multiple images with the same `name`, but different distgit key. For example, `ci-openshift-build-root-latest.rhel8` and `ci-openshift-build-root-latest.rhel9`. 

We can still distinguish builds for these components in BigQuery using the `el_target` column, but only among successful ones. Failed builds won't store the rhel target (as it's being deducted from the build NVR), so actors like `doozer images:health` treat failed builds for the rhel8/rhel9 images as belonging to the same item.

The distgit key is currently the right way to uniquely define an image component. We will eventually need to purge references to `distgit`, but that can be addressed when time comes.


joepvd (approved):
/lgtm

PR #972, opened 2024-09-24
[konflux] Fail builds fix

ashwindasr:
In images.py we collect failed builds as follows:

```
failed = [r for r in results if isinstance(r, Exception)]
        if failed:
            raise DoozerFatalError(f"Failed to build images: {failed}")
```

during refactoring, we removed the child function from raising a error if it fails, adding that back in


ashwindasr:
test [run](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-konflux/49/console) seems to work

added `Exception(metadata.distgit_key)` after the test run, so from next time, the distgit name of the failed images will also be displayed

PR #978, opened 2024-09-25
[Konflux] Fix resolving short images in streams.yml

vfreex:
We have a lot of stream parant images defined in streams.yml, where short image names are used, such as
`openshift/golang-builder:tag`.

In order to reference those images in Konflux, we need to prepend the Brew registry URL.


joepvd:
/hold
unsure, but suspect this might trigger mass rebuilds

ashwindasr (approved):
Since this is file / class is only for Konflux, we can merge it

joepvd:
/unhold
This is a konflux only change, my fear does not apply.

PR #982, opened 2024-09-25
[konflux] replace golang stream with brew.registry

ashwindasr:
The brew.registry url is the registry-proxy url, but with the domain replaced by the former and the repo is `rh-osbs`. Also the golang builder's pullspec is openshift-golang-builder


ashwindasr:
Test run [succeeded](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-konflux/62)

vfreex:
@ashwindasr Thanks. I pushed another commit to this PR so it can convert for all images on Brew besides the golang-builder.

/lgtm
/approve

PR #984, opened 2024-09-26
elliott: Filter out archived target versions

vfreex:
In order to process bugs like OCPBUGS-39183, which have multiple target versions but we couldn't remove the incorrect target version with archived flag set to True.


joepvd (approved):
/lgtm

PR #985, opened 2024-09-26
Revert "Fix sticky revert when annotation is not present"

joepvd:
Reverts openshift-eng/art-tools#980


joepvd:
/hold
unsure if revert is the path forward. [thread](https://redhat-internal.slack.com/archives/GDBRP5YJH/p1727349800546869)

PR #987, opened 2024-09-26
verify-bugs: Check for invalid bug status first

thegreyd:
Right now we check for invalid status after filtering bugs by desired target release.
Sometimes we want to get info about invalid bug status before invalid target release check.
Example, when we have invalid target release 4.17.z when the desired is 4.17.0, the bug might
still be valid to be attached. But not if it has invalid bug status.


joepvd (approved):
/lgtm

PR #989, opened 2024-09-27
konflux build: Fix a deadlock when waiting for failed parents

vfreex:
This issue was introduced from
https://github.com/openshift-eng/art-tools/pull/955/files and https://github.com/openshift-eng/art-tools/pull/972/files, where a build error raises an exception and skips the line to set build_event.

This PR also adds distgit key to the log message to make troubleshooting easier.


joepvd (approved):
/lgtm

PR #1000, opened 2024-10-02
Populate parent images + store image tag

thegreyd:


doozer/doozerlib/backend/konflux_image_builder.py
```diff
@@ -99,7 +99,7 @@ async def build(self, metadata: ImageMetadata):
 
             # Start the build
             self._logger.info("Starting Konflux image build for %s...", metadata.distgit_key)
-            retries = 5
+            retries = 3
```
ashwindasr, line 102:
```suggestion
            retries = 5
```
Lets keep the retry as 5?

thegreyd: 3 should be enough to rule out flakes, or do you think otherwise?

thegreyd: keeping at 5 means when there is a genuine failure,  we wait longer. That's the case I saw in the pipeline recently and i want to avoid

ashwindasr (approved):

PR #1002, opened 2024-10-02
remove pipeline timeout, but keep build

ashwindasr:
Remove pipeline timeout override. We need build time out to be 12 hrs so that during mass rebuilds, the PLR is not pruned while waiting to build, which has happened a couple of times


ashwindasr:
/close

caused by unrelated [issue](https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1727895264059049?thread_ts=1727889917.143019&cid=C04PZ7H0VA8) on konflux's side

PR #1004, opened 2024-10-02
Always lock to datetime.UTC

thegreyd:
Weird things happen when we don't lock datetime.now() to UTC and run from our own local systems.


locriandev (commented):
/lgtm

PR #1008, opened 2024-10-03
Log build and rebase exceptions

thegreyd:


doozer/doozerlib/cli/images_konflux.py
```diff
@@ -71,9 +72,15 @@ async def run(self):
...
-        failed = [r for r in results if isinstance(r, Exception)]
-        if failed:
-            raise DoozerFatalError(f"Failed to rebase images: {failed}")
+        failed_images = []
+        for index, result in enumerate(results):
+            if isinstance(result, Exception):
+                image_name = metas[index].distgit_key
+                failed_images.append(image_name)
+                LOGGER.error(f"Failed to rebase {image_name}")
+                traceback.print_exception(result)
```
locriandev, line 81: Is this really needed? Printing the stack trace to the logs makes them less readable, I think it should be done only if strictly needed.

locriandev: otherwise LGTM

ashwindasr:
Might be a good idea for @vfreex to review before we merge

doozer/doozerlib/cli/images_konflux.py
thegreyd: There's no debug.log we have here which we can refer to. Until we have that we have no way of knowing where exactly the error occured

thegreyd: reverted traceback, merging with minimal changes

thegreyd:
I merged with minimal changes to not be too disruptive, but we really need some kind of good logging

PR #1012, opened 2024-10-04
[konflux] bump pipeline run template

ashwindasr:
Right now, we are using an old template. 

Most of the params are overriden, but we need to double check, but this is the new template from https://github.com/openshift-priv/art-konflux-template/blob/main/.tekton/art-konflux-template-pulkl-request.yaml

The next step is to pull directly from that repo, instead of storing it in art-tools

### DIFF
--- a/doozer/doozerlib/backend/konflux_image_build_pipelinerun.yaml
+++ b/doozer/doozerlib/backend/konflux_image_build_pipelinerun.yaml
@@ -14,7 +14,7 @@ metadata:
     appstudio.openshift.io/component: ui-ocp-4-18-base-rhel-9
     pipelines.appstudio.openshift.io/type: build
   name: ui-ocp-4-18-base-rhel-9-on-push
-  namespace: crt-nshift-art-pipeli-tenant
+  namespace: ocp-art-tenant
 spec:
   timeouts:
     # See https://konflux-ci.dev/docs/how-tos/configuring/customizing-the-build/#configuring-timeouts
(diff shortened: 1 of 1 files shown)


ashwindasr:
/hold

Until we do a test build

ashwindasr:
/unhold

Test build [succeeded](https://konflux.apps.stone-prod-p02.hjvn.p1.openshiftapps.com/application-pipeline/workspaces/ocp-art/applications/openshift-4-18/pipelineruns/doozer-build-openshift-4-18-openshift-base-rhel9-9x5s4)

doozer/doozerlib/backend/konflux_image_build_pipelinerun.yaml
```diff
@@ -167,40 +185,38 @@ spec:
...
+        workspace: workspace
       - name: git-basic-auth
         workspace: git-auth
       - name: netrc
         workspace: netrc
-    - matrix:
-        params:
-        - name: PLATFORM
-          value:
-          - $(params.build-platforms)
```
thegreyd, line 202: Looking at the build that got created, built for only 1 arch, https://konflux.apps.stone-prod-p02.hjvn.p1.openshiftapps.com/application-pipeline/workspaces/ocp-art/applications/openshift-4-18/pipelineruns/doozer-build-openshift-4-18-openshift-base-rhel9-9x5s4/logs

thegreyd: is this the multi arch template?

ashwindasr:
/hold

Looks like we need to onboard a multi-arch component.

ashwindasr:
since this config was the correct one. Copied over to openshift-priv: https://github.com/openshift-priv/art-konflux-template/pull/7

will wait a couple of days to see if konflux auto updates that file

PR #1016, opened 2024-10-08
update file_upload_v2 parameters

Ximinhan:
fix error
```
UserWarning: Although the channels parameter is still supported for smooth migration from legacy files.upload, we recommend using the new channel parameter with a single str value instead for more clarity.
The filetype parameter is no longer supported. Please remove it from the arguments.
```
update channels parameter to channel
remove filetype parameter
file parameter should accept file path string
and file_upload_v2 will call files.completeUploadExternal which seems only accept channel_id so set channel with id instead of its name


joepvd (approved):
/lgtm

PR #1017, opened 2024-10-08
Add extra pattern search to get_latest_build in Konflux

locriandev:
Needed by scan sources to add select clauses like `AND commitish LIKE  '%b45ea65%'`

### DIFF
--- a/artcommon/artcommonlib/konflux/konflux_db.py
+++ b/artcommon/artcommonlib/konflux/konflux_db.py
@@ -170,7 +170,7 @@ class KonfluxDb:
 
     def get_latest_build(self, name: str, group: str, outcome: KonfluxBuildOutcome = KonfluxBuildOutcome.SUCCESS,
                          assembly: str = 'stream', el_target: str = None, artifact_type: ArtifactType = None,
-                         engine: Engine = None, completed_before: datetime = None)\
+                         engine: Engine = None, completed_before: datetime = None, extra_patterns: dict = {})\
             -> typing.Optional[konflux_build_record.KonfluxBuildRecord]:
         """
         Search for the latest Konflux build information in BigQuery.
@@ -183,6 +183,7 @@ class KonfluxDb:
         :param artifact_type: 'rpm' | 'image'
         :param engine: 'brew' | 'konflux'
         :param completed_before: cut off timestamp for builds completion time
+        :param extra_patterns: e.g. {'release': 'b45ea65'} will result in adding "AND release LIKE '%b45ea65%'" to the query
         """
 
         if not completed_before:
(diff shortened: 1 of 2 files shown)


artcommon/artcommonlib/konflux/konflux_db.py
```diff
@@ -170,7 +170,7 @@ async def get_latest_builds(self, names: typing.List[str], group: str,
 
     def get_latest_build(self, name: str, group: str, outcome: KonfluxBuildOutcome = KonfluxBuildOutcome.SUCCESS,
                          assembly: str = 'stream', el_target: str = None, artifact_type: ArtifactType = None,
-                         engine: Engine = None, completed_before: datetime = None)\
+                         engine: Engine = None, completed_before: datetime = None, extra_patterns: dict = {})\
```
thegreyd, line 173: While this works, I think it would be nice to add commit as a first class search param (something like `git_short_commit`). Then the fact that the short commit is part of the release field can remain the funcs internal implementation which the caller doesn't need to know.

thegreyd: Also this could be different for bundle builds since I think `version` is where the git commit resides..

thegreyd: Or we just implement generic support for nvr pattern, then the caller doesn't need to know which field, but just the intent of "search nvrs with this in nvr name"

vfreex: `extra_patterns` sounds too generic to me either. Maybe we need to rename it or make those fields for search as first-class parameters. otherwise it looks good to me.

locriandev: In the unit tests I have probably provided a misleading example. Commitish is being stored in its on field in the DB, and it's also part of the NVR (that we're already storing). 

I deliberately made this generic to give us extra flexibility, so `extra_patterns` was named after this idea. If you have a better name I can replace it, but I would still keep the generic thing. Then, I can give the `commitish` search a greater dignity, by making it explicit like Sid suggested. We'll be searching in the dedicated field anyway, since the DB was designed to state the upstream commitish explicitly for successful builds.

vfreex (commented):
/lgtm
/approve

vfreex (commented):
/hold

artcommon/artcommonlib/konflux/konflux_db.py
locriandev: @thegreyd we're already storing the NVR. To do stuff like "nvr contains this commit" we still need this PR in. Keeping it generic will give us the greatest flexibility. I renamed the param to `search_patterns` to make it clearer. Merging as-is, then I will open a follow up PR to make the commitish search a first class citizen as you suggested

PR #1018, opened 2024-10-08
scan-sources: skip unnecessary checks

locriandev:
scan-sources first checks for upstream changes (new commits that have not been built yet) and marks those images that need a fresh rebuild. 

Then, it looks for other reasons to rebuild (e.g. configuration changes, dependency reasons. etc.). These are time consuming at should be skipped completely if the image has already been tagged for rebuild.

I expect this change to reduce the execution time for scan-sources.


joepvd (approved):
/lgtm

PR #1024, opened 2024-10-10
Fix download_file_from_github(); add a retry

locriandev:
- Change auth headers from `{"Authorization": f"token {token}"}` to `{"Authorization": token}`
- Add a retry block (up to 3 retries, 10 seconds sleep)
- Add logs


vfreex (commented):
/lgtm
/approve

joepvd:
/unhold

was confused about lack of `Bearer`, but this is a request to a different end point.

PR #1035, opened 2024-10-11
[konflux] DB log print distgit name

ashwindasr:
```
2024-10-10 21:11:04,742 doozerlib.backend.konflux_image_builder INFO Konflux build info stored successfully
```
Since builds are running async, we don't know which distgit components build got stored correctly. Hence proposing to prepend the distgit name


vfreex (commented):
/lgtm
/approve

PR #1036, opened 2024-10-11
Slow down multi nightly creation

jupierce:
Multi nightlies are different than single arch nightlies. Each update causes the release controller to run CI jobs vs single arch imagestream updates, where the release controller prevents nightlies from being created too often.
Currently, nearly every run on gen_payload is going to create a new multi-nightly. Limit this rate.


joepvd (approved):
/lgtm

PR #1041, opened 2024-10-14
Konflux: make get_latest_build async

locriandev:
- make `bigquery.select()` async, to allow concurrent queries to the DB
- make `konflux_db.get_latest_build()` async to boost scan-sources for Konflux performance
- fix a small bug in `konflux_db.get_latest_build()` that was messing up log messages (fortunately, not the behavior of the function itself)


vfreex (commented):
/lgtm
/approve

PR #1044, opened 2024-10-14
Limit number of concurrent Konflux pipelines

locriandev:
Currently, Konflux has very low capacity and we must ensure we do not allocate more pipelines than the cluster can handle. According to previous discussions, we should limit concurrent builds to 2 for now. Setting an explicit constant to let us bump this number easily once we get a green flag.

### DIFF
--- a/doozer/doozerlib/backend/konflux_image_builder.py
+++ b/doozer/doozerlib/backend/konflux_image_builder.py
@@ -17,6 +17,7 @@ from kubernetes.client import Configuration
 from kubernetes.dynamic import DynamicClient, exceptions, resource
 from ruamel.yaml import YAML
 
+from artcommonlib.exectools import limit_concurrency
 from doozerlib import constants
 from doozerlib.backend.build_repo import BuildRepo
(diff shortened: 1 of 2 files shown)


ashwindasr:
Agreed with Yuxiang's comment above. We can increase the number of concurrent builds to 10-15 if that makes things easier for us

/hold
Daniele, feel free to unhold when ready

locriandev:
Retry is already in place. Will increase to 10, but will need to change a bit because this implementation might lead to deadlocks when waiting for parents. I will keep you posted

ashwindasr:
/hold

Until issue of retires not working properly is fixed

ashwindasr:
rebased PR to pull in new changes from prod/main

locriandev:
@vfreex I've checked how we start builds in Doozer, and it seems that we safely manage dependency relationships (details in [this thread](https://redhat-internal.slack.com/archives/GDBRP5YJH/p1728923699573349?thread_ts=1728898681.147159&cid=GDBRP5YJH))

I think we're good, unholding.

/unhold

PR #1050, opened 2024-10-15
errata: reverse advisory dependencies

joepvd:
Looks like the image advisory cannot ship when the rpm advisory is not PUSH_READY. Reverse the dependencies.


ashwindasr:
/hold

Do not think this is the issue [here](https://redhat-internal.slack.com/archives/CJARLA942/p1729013359415379)

PR #1053, opened 2024-10-15
[konflux] reduce character length PLR

ashwindasr:
Kubernetes resources has a maximum character length of [63 characters](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names). We seem to be generating names greater than that, so its getting truncated, for example:
```
doozer-build-openshift-4-18-ose-installer-terraform-providpkxhm
```

Proposing to change remove the `doozer-build-openshift-` prefix so that we can preserve as much as the distgit name as possible

### DIFF
--- a/doozer/doozerlib/backend/konflux_image_builder.py
+++ b/doozer/doozerlib/backend/konflux_image_builder.py
@@ -190,8 +190,12 @@ class KonfluxImageBuilder:
             raise ValueError(f"[{metadata.distgit_key}] Unsupported arches: {', '.join(unsupported_arches)}")
         build_platforms = [self.SUPPORTED_ARCHES[arch] for arch in arches]
         pipelineruns_api = await self._get_pipelinerun_api(dyn_client)
+
+        # Replace 'openshift-4-18-ose-installer-terraform' with '4-18-ose-installer-terraform'
+        formatted_pipelinerun_name = component_name.replace("openshift-", "", 1) if component_name.startswith("openshift-") else component_name
+
         pipelinerun_manifest = self._new_pipelinerun(
-            f"doozer-build-{component_name}-",
+            formatted_pipelinerun_name,
             app_name,
             component_name,
             git_url,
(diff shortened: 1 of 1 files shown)


doozer/doozerlib/backend/konflux_image_builder.py
```diff
@@ -190,8 +190,12 @@ async def _start_build(self, metadata: ImageMetadata, build_repo: BuildRepo, dyn
             raise ValueError(f"[{metadata.distgit_key}] Unsupported arches: {', '.join(unsupported_arches)}")
         build_platforms = [self.SUPPORTED_ARCHES[arch] for arch in arches]
         pipelineruns_api = await self._get_pipelinerun_api(dyn_client)
+
+        # Replace 'openshift-4-18-ose-installer-terraform' with '4-18-ose-installer-terraform'
+        formatted_pipelinerun_name = component_name.replace("openshift-", "", 1) if component_name.startswith("openshift-") else component_name
+
         pipelinerun_manifest = self._new_pipelinerun(
-            f"doozer-build-{component_name}-",
+            formatted_pipelinerun_name,
```
thegreyd, line 198: generate name should end with a trailing `-`

ashwindasr: Thanks for the catch! Updated.

doozer/doozerlib/backend/konflux_image_builder.py
```diff
@@ -190,8 +190,12 @@ async def _start_build(self, metadata: ImageMetadata, build_repo: BuildRepo, dyn
             raise ValueError(f"[{metadata.distgit_key}] Unsupported arches: {', '.join(unsupported_arches)}")
         build_platforms = [self.SUPPORTED_ARCHES[arch] for arch in arches]
         pipelineruns_api = await self._get_pipelinerun_api(dyn_client)
+
+        # Replace 'openshift-4-18-ose-installer-terraform' with '4-18-ose-installer-terraform'
+        formatted_pipelinerun_name = component_name.replace("openshift-", "", 1) if component_name.startswith("openshift-") else component_name
```
vfreex, line 195: Let's use `removeprefix` in case a component name has `openshift` in its name.
```suggestion
        formatted_pipelinerun_name = component_name.removeprefix("openshift-")
```

Or maybe just truncate the pipelinerun name if it is too long.

ashwindasr: >  in case a component name has openshift in its name.
yeah - my crude implementation was removing only the first occurrence.

> Let's use removeprefix
agreed - this is better

ashwindasr:
Merging this for now, please let me know if we need any follow ups

PR #1057, opened 2024-10-16
[konflux] setup art-db custom task

ashwindasr:
If the Jenkins pipeline fails due to some reason, after triggering konflux builds, the builds will still be running. Since the pipeline is not running anymore, we won't be able to store the status of the build, once it completes on Konflux.

Tekton has [git resolvers](https://tekton.dev/docs/pipelines/git-resolver/) that we can leverage for creating custom tasks. Task defined here: https://github.com/openshift-priv/art-konflux-template/blob/main/custom-tasks/art-store-to-db.yaml.


ashwindasr:
Tested

/unhold

PR #1059, opened 2024-10-17
Fix konflux_db.get_latest_build calls

vfreex:
This function has changed to be async.


locriandev:
/hold
I want to check if we can avoid running `asyncio.run()`, which will fail if there's another event loop running

locriandev:
/unhold

Turns out the refactoring effort to make everything async is too big at this moment

locriandev (approved):
/lgtm

PR #1066, opened 2024-10-18
ART-10995 fix false alarm for regression check

Ximinhan:
add two API function, 
`get_next_release_schedule` will get the next release date for a stream release 
the pp API `schedule-tasks/?flags_and__in=dev&fields=path,date_finish` return list of dates like
```
{
      "path": [
          "OpenShift Container Platform (OCP) 4.16.z GA Release Schedule",
          "4.16.z Release Cycle",
          "4.16.25"
      ],
      "date_finish": "2024-12-04"
  },
  {
      "path": [
          "OpenShift Container Platform (OCP) 4.16.z GA Release Schedule",
          "4.16.z Release Cycle",
          "4.16.26"
      ],
      "date_finish": "2024-12-11"
  },
```
`is_release_this_week` will check if the next release need to prepare this week
so when check regression bug issue, if we don't have next X+1 release need to prepare then blocker not attached to any advisory is expected.

### DIFF
--- a/artcommon/artcommonlib/util.py
+++ b/artcommon/artcommonlib/util.py
@@ -162,7 +162,7 @@ def is_release_this_week(release_date):
 
 def get_next_release_schedule(group):
     """
-    Get next release name based on current date
+    Get next release date based on current date
     """
     dev_schedules = requests.get(f'{RELEASE_SCHEDULES}/{group}.z/schedule-tasks/?flags_and__in=dev&fields=path,date_finish', headers={'Accept': 'application/json'})
     dev_schedules.raise_for_status()
--- a/artcommon/artcommonlib/util.py
+++ b/artcommon/artcommonlib/util.py
@@ -1,7 +1,7 @@
 import logging
 from time import sleep
 from typing import OrderedDict, Optional, Tuple, Iterable, List
-from datetime import datetime, timezone
+from datetime import datetime, timezone, timedelta
 import re
 import asyncio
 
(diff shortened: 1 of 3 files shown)


artcommon/artcommonlib/util.py
```diff
@@ -153,6 +153,25 @@ def get_assembly_release_date(assembly, group):
...
+    Check if release date is in the near week
+    """
+    return datetime.strptime(release_date, "%Y-%m-%d").date() <= date.today() + timedelta(days=7)
+
+
+def get_next_release_schedule(group):
+    """
+    Get next release date based on current date
+    """
+    dev_schedules = requests.get(f'{RELEASE_SCHEDULES}/{group}.z/schedule-tasks/?flags_and__in=dev&fields=path,date_finish', headers={'Accept': 'application/json'})
```
joepvd, line 167: Maybe a retry would be good?

elliott/elliottlib/cli/verify_attached_bugs_cli.py
```diff
@@ -474,6 +475,10 @@ def _verify_blocking_bugs(self, blocking_bugs_for, is_attached=False):
                 if is_attached and blocker.status in ['ON_QA', 'Verified', 'VERIFIED']:
                     blocker_advisories = blocker.all_advisory_ids()
                     if not blocker_advisories:
+                        # check if we need to release X+1 release this week
```
joepvd, line 478: Can this be checked to see if the blocking test should run at all, and not for every issue if it is not attached in the next version? 

Think the new release date check should decide if `blocker.all_advisory_ids()` should be run.

joepvd (commented):
Thanks for picking this up! Left some comments.

artcommon/artcommonlib/util.py
Ximinhan: from my experience, I think there is no need for a retry here

elliott/elliottlib/cli/verify_attached_bugs_cli.py
Ximinhan: good point, updated in https://github.com/openshift-eng/art-tools/pull/1066/commits/b63a6c14107cdf5a9bb6f716bb9258fc8ab2dc66

vfreex:
Sorry for reverting your PR. See conversation at https://redhat-internal.slack.com/archives/C03JYDGPX8F/p1730368623345419?thread_ts=1730365527.269409&cid=C03JYDGPX8F. After you get it improved feel free to open another PR to get it in.

PR #1083, opened 2024-10-22
ART-11111: versioning scheme decides the right param value for Jira t…

mbiarnes:
…emplate

rh-pre-commit.version: 2.3.1

### DIFF
--- a/pyartcd/pyartcd/pipelines/prepare_release.py
+++ b/pyartcd/pyartcd/pipelines/prepare_release.py
@@ -134,6 +134,9 @@ class PrepareReleasePipeline:
         release_config = releases_config.get("releases", {}).get(self.assembly, {})
         self.release_name = get_release_name_for_assembly(self.group_name, releases_config, self.assembly)
         self.release_version = semver.VersionInfo.parse(self.release_name).to_tuple()
+        # Convert string to an integer if necessary (like 4.20.0-0 will result in
+        # release_versions = (4, 20, 0, '0', None) )
+        self.release_version = tuple(int(i) if isinstance(i, str) and i.isdigit() else i for i in self.release_version)
         if not release_config:
             raise ValueError(f"Assembly {self.assembly} is not explicitly defined in releases.yml for group {self.group_name}.")
         group_config = assembly_group_config(Model(releases_config), self.assembly, Model(group_config)).primitive()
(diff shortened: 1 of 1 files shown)


pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -134,6 +134,9 @@ async def run(self):
         release_config = releases_config.get("releases", {}).get(self.assembly, {})
         self.release_name = get_release_name_for_assembly(self.group_name, releases_config, self.assembly)
         self.release_version = semver.VersionInfo.parse(self.release_name).to_tuple()
+        # Convert string to an integer if necessary (like 4.20.0-0 will result in
```
joepvd, line 137: Not sure why the line above does `to_tuple()`. the semver.version.Version instance has nice attributes:

```
>>> v = semver.VersionInfo.parse('4.5.6-1')
>>> v.to_dict()
{'major': 4, 'minor': 5, 'patch': 6, 'prerelease': '1', 'build': None}
>>> v.major
4
```

But this makes me realize that the proposed versioning scheme is not semver compliant. 

Most standard compliant solution would be to have this in the build info. 

```
>>> semver.VersionInfo.parse('4.5.6-1')
Version(major=4, minor=5, patch=6, prerelease='1', build=None)
>>> semver.VersionInfo.parse('4.5.6-rc.0-0')
Version(major=4, minor=5, patch=6, prerelease='rc.0-0', build=None)
```

meaning, for an RC, semver does not understand semantically understand what is going on. 

The scheme to denote our build after a `+` seems the standard way of doing it:
```
>>> semver.VersionInfo.parse('4.5.6-rc.0+0')
Version(major=4, minor=5, patch=6, prerelease='rc.0', build='0')
>>> semver.VersionInfo.parse('4.5.6-rc.0+1')
Version(major=4, minor=5, patch=6, prerelease='rc.0', build='1')
>>> semver.VersionInfo.parse('4.5.7+1')
Version(major=4, minor=5, patch=7, prerelease=None, build='1')
```

pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -216,11 +219,21 @@ async def run(self):
...
+        # Apply conditional logic if the new_version scheme is applied (like 4.20.0-0)
+        if self.release_version[3] is not None:
+            z = f"{self.release_version[2]}-{self.release_version[3]}"
+
         jira_template_vars = {
             "release_name": self.release_name,
-            "x": self.release_version[0],
-            "y": self.release_version[1],
-            "z": self.release_version[2],
+            "x": x,
```
joepvd, line 234: would rather use the attributes semver returns, 

```python
self.release_version = semver.VersionInfo.parse(self.release_name)

# so here it becomes
"x": self.release_version.major
```

joepvd (changes requested):
A comment, but maybe we should return to the drawing board.

pyartcd/pyartcd/pipelines/prepare_release.py
joepvd: The doubt about the versioning scheme is resolved, `+` is not a valid token for a k8s object. We need to deviate from semver.

mbiarnes:
the new-version - system was never implemented

PR #1084, opened 2024-10-24
Respect art:pscomponent: label for new-typed CVE trackers

vfreex:
We need this flexibility.


joepvd:
/hold
Doubt that this is the solution

PR #1085, opened 2024-10-24
Elliott: A new approach to check for shipped rpms with product version awareness

vfreex:
Currently, we consider an rpm shipped if it has `*-released` Brew tag.
Released rpms will be excluded from `find-builds` so they will not be
attached to Errata.

This has caused a few issues: Some external rpms (e.g. python-zipp) are
tagged into both rhaos-4.17-ironic-rhel-9-candidate and rhaos-4.16-ironic-rhel-9-candidate.
Once the rpm is shipped to 4.17, we will not be able to ship it to 4.16.

With the new approach, we read all candidate tags for an OCP minor
version from ocp-build-data erratatool.yml, they consider a rpm is
shipped if it has a tag without the -candidate suffix.

The feature is not enabled by default unless group config option
`feature_gates.product_version_aware_rpm_sweep` is enabled.

### DIFF
--- a/elliott/elliottlib/cli/find_builds_cli.py
+++ b/elliott/elliottlib/cli/find_builds_cli.py
@@ -4,7 +4,7 @@ import json
 import logging
 import re
 import sys
-from typing import Dict, List, Set, Union
+from typing import Dict, List, Optional, Set, Union
 
 import click
 import koji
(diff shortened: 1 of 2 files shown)


vfreex:
/hold cancel

Would like to use a group config option to control whether this new approach should be used.

PR #1086, opened 2024-10-24
[konflux] restrict arches to x86 temporarily

ashwindasr:
Restrict to just x86 so that we can mass build and test konflux ocp4 scan. x86 builds on cluster, to builds are significantly faster than on other arches. ([thread](https://redhat-internal.slack.com/archives/C02CTEB3MMF/p1729778261627059?thread_ts=1729778032.439479&cid=C02CTEB3MMF))


locriandev (commented):
/lgtm

ashwindasr:
Fixed flake test, reapplying lgtm

/approve

PR #1089, opened 2024-10-25
[konflux] fix bug arches

ashwindasr:
Changes in the PR:
-  cosign and DB were storing all the arches defined for an image, instead of storing the arches that we are actually building for. 
- Temporarily do not fail for unsupported arches until we are on the new cluster

Test run [succeeded](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-konflux/231/console)


locriandev (commented):
/lgtm

PR #1091, opened 2024-10-28
ART-11168 Add ocp4-scan for konflux artcd pipeline

locriandev:
Ref. [ART-11168](https://issues.redhat.com/browse/ART-11168)
Needs https://github.com/openshift-eng/aos-cd-jobs/pull/4275

### DIFF
--- a/doozer/doozerlib/cli/scan_sources_konflux.py
+++ b/doozer/doozerlib/cli/scan_sources_konflux.py
@@ -267,6 +267,9 @@ class ConfigScanSources:
             image_names, latest_image_builds)))
 
     async def scan_images(self, image_names: typing.List[str]):
+        # Do not scan images that have been disabled for Konflux operations
+        image_names = filter(lambda name: self.runtime.image_map[name].config.konflux.mode != 'disabled', image_names)
+
         # Do not scan images that have already been requested for rebuild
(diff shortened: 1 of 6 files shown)


doozer/doozerlib/cli/scan_sources_konflux.py
```diff
@@ -267,6 +267,9 @@ async def find_latest_image_builds(self, image_names: typing.List[str]):
             image_names, latest_image_builds)))
 
     async def scan_images(self, image_names: typing.List[str]):
+        # Do not scan images that have been disabled for Konflux operations
+        image_names = filter(lambda name: self.runtime.image_map[name].config.konflux.mode != 'disabled', image_names)
```
ashwindasr, line 271: In addition to this, we have to filter out images with `mode: disabled` as well. Are we already doing that?

locriandev: Yes, `mode: disabled` images are excluded by Doozer. If you try to explicitly request one, you'll get an error like this:
```
The following images or rpms were either missing or filtered out: microshift
```

ashwindasr: ack thanks

PR #1095, opened 2024-10-29
[konflux] Support for dockerfile override

ashwindasr:
Support a Dockerfile override so that we can specify a separate file path in upstream. This is required for cachito images, since cachi2 works differently on konflux.

Details in https://github.com/openshift/etcd/pull/294


joepvd (approved):
/lgtm

PR #1101, opened 2024-10-31
Revert "ART-10995 fix false alarm for regression check"

vfreex:
Reverts openshift-eng/art-tools#1066

This has caused an error when promoting 4.12.68: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fpromote-assembly/402/consoleFull

PR #1103, opened 2024-10-31
ART-10995 fix false alarm for regression check

Ximinhan:
relate to https://github.com/openshift-eng/art-tools/pull/1101
in https://github.com/openshift-eng/art-tools/pull/1066 we use development_cut_off date as a line to check is_release_next_week, but as 4.12 release is going to slow down the latest release of development_cut_off could be the day before release prepare day
so use the schedule of release in fast channel, if there is a date in next week then means we need to prepare it today.

### DIFF
--- a/artcommon/artcommonlib/util.py
+++ b/artcommon/artcommonlib/util.py
@@ -1,6 +1,6 @@
 import logging
 from typing import OrderedDict, Optional, Tuple, Iterable, List
-from datetime import datetime, timezone
+from datetime import datetime, timezone, timedelta, date
 import re
 import asyncio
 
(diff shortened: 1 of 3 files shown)


joepvd:
is the strptime issue also fixed? See e.g. https://redhat-internal.slack.com/archives/CJARLA942/p1730349250556049

Ximinhan:
> is the strptime issue also fixed? See e.g. https://redhat-internal.slack.com/archives/CJARLA942/p1730349250556049

yes, the error is because used to use get_next_release_schedule return None and pass it to is_release_this_week 
with this pr use another function https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fcheck-bugs/187/console

PR #1104, opened 2024-11-01
ART-11093 add ignore_non_x86_nightlies arg for gen-assembly

Ximinhan:
add ignore_non_x86_nightlies arg, if this arg is checked then it will only use latest accepted x86_64 nightly, default to False.
this arg honors the skip_get_nightlies arg, if skip_get_nightlies is true the input value of nighties set will be used.

------------
Tested with https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/hack/job/ximhan/job/build%252Fgen-assembly/5/consoleFull

and it created assembly pr https://github.com/openshift-eng/ocp-build-data/pull/5832
only have x86_64 nightly, but full arch of RHCOS and the brew event come from x86_64 nightly


vfreex (commented):
Shouldn't this be accomplished by LIMIT_ARCHES?

Ximinhan:
> Shouldn't this be accomplished by LIMIT_ARCHES?

they are a little bit different, the LIMIT_ARCH needs to be used with skip_get_nightlies and use the inputted nightly to get candidate_nightlies and rhcos is also limited by that arch config.
but ignore_non_x86_nightlies is trying to skip looking at non-x86_64 nightlies but still gen-assembly as we have full arches, rhcos and builds still use full arches.

pyartcd/pyartcd/pipelines/gen_assembly.py
```diff
@@ -97,11 +98,13 @@ async def run(self):
             if self.custom and (self.auto_previous or self.previous_list or self.in_flight):
                 raise ValueError("Specifying previous list for a custom release is not allowed.")
 
-            if not self.skip_get_nightlies:
+            if self.skip_get_nightlies:
+                candidate_nightlies = self.nightlies
+            elif self.ignore_non_x86_nightlies:
+                candidate_nightlies = await self._get_latest_accepted_nightly()
```
vfreex, line 104: Looks like setting this option will cause the pipeline completely ignore the `--nightly` parameter. Is there a way to fix that?

Ximinhan: this is the design, when ignore_non_x86_nightlies set only use latest accepted x86 nightly.
nightly parameter is used with skip_get_nightlies, if skip_get_nightlies is true the input value of nighties set will be used.

.github/workflows/unit-tests.yaml
```diff
@@ -22,7 +22,11 @@ jobs:
           fetch-depth: 0
           set-safe-directory: ${{ env.GITHUB_WORKSPACE }}
       - name: Install packages
-        run: dnf install -y gcc krb5-devel make glibc
+        run: dnf install -y gcc krb5-devel make glibc python3-rpm
```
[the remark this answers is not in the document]
Ximinhan: there used to be an error with `Failed to import system RPM module. Make sure RPM Python bindings are installed on your system.` on https://github.com/openshift-eng/art-tools/actions/runs/11624887515/job/32374119155 so import rpm in python rely on `python3-rpm` rpm instead of rpm lib in requirements.txt

pyartcd/pyartcd/pipelines/gen_assembly.py
thegreyd: why not use the x86 nightly if it is provided as an arg?

Ximinhan: this is what we want, if ignore_non_x86_nightlies is true we only use x86_64 nightly and build the assembly as we did with all arch nightly, we use x86_64 nightly get brew event and get full arch of rhcos from the rhcos version in x86_64 nightly

joepvd (approved):
/lgtm

PR #1105, opened 2024-11-01
[konflux] add assembly type to app name, component name

ashwindasr:
When we tried to trigger `stream` builds, openshift did not let us update the `Component` since it was tied to a `test` GH branch. Hence we want to separate `Application` and `Component` CRDs based on version and assembly.

So an application from now on will look like `openshift-4-18-stream` / `openshift-4-18-test` and a component will look like `stream-4-18-ose-etcd` / `test-4-18-ose-etcd`

Also since Kubernetes resources have a maximum character length of 63 characters, lets no onboard a component with very long distgit names.


doozer/doozerlib/backend/konflux_image_builder.py
```diff
@@ -169,18 +169,22 @@ async def _start_build(self, metadata: ImageMetadata, build_repo: BuildRepo, dyn
             raise ValueError(f"[{metadata.distgit_key}] Dockerfile must have a '__doozer_uuid_tag' environment variable; Did you forget to run 'doozer beta:images:konflux:rebase' first?")
 
         # Ensure the Application resource exists
-        app_name = self._config.group_name.replace(".", "-")
+        group_name = self._config.group_name.replace(".", "-")
+        app_name = group_name
+        app_name = f"{app_name}-{metadata.runtime.assembly}"
```
joepvd, line 174: use group_name directly

doozer/doozerlib/backend/konflux_image_builder.py
```diff
@@ -169,18 +169,22 @@ async def _start_build(self, metadata: ImageMetadata, build_repo: BuildRepo, dyn
...
         app_manifest = self._new_application(app_name, app_name)
         app = await self._create_or_patch(dyn_client, app_manifest)
         self._logger.info(f"[%s] Using application: {app['metadata']['name']}", metadata.distgit_key)
 
         # Ensure the component resource exists
         # Openshift doesn't allow dots or underscores in any of its fields, so we replace them with dashes
-        component_name = f"{app_name}-{metadata.distgit_key}".replace(".", "-").replace("_", "-")
+        # Since Component is a CRD, it has to be unique in the namespace. Also once a Component and Application is created, the branches that its linked to cannot be modified
+        # Hence prefixing the assembly and version to the component name to make it unique, eg: "stream-4-18-ose-etcd"
+        component_name = f"{metadata.runtime.assembly}-{group_name.removeprefix(f'openshift-')}-{metadata.distgit_key}".replace(".", "-").replace("_", "-")
```
joepvd, line 183: rather than computing something from distgit-key, can we have the konflux name entity hard coded rather than calculated through an attribute in ocp-build-data?

doozer/doozerlib/backend/konflux_image_builder.py
ashwindasr: I'm using the group name below: https://github.com/openshift-eng/art-tools/pull/1105/files#diff-e45d70130344e3d45bd59cb091bd0a794121c675ce90f17c5b4e4702ded620ffR183

So didn't want to replace the `.` with `-` again

vfreex:
We might need to ship a release coming from mixed assemblies (e.g. stream and 4.y.z). I am not sure if it will be an issue if we separate assemblies and components by assembly name. Do you have any other reason (except shortening resource names) to make this change?

Also it guess it might be fine to just truncate the distgit keys in some way or having a new field as suggested by @joepvd . What do you think?

thegreyd:
Yeah this change makes me weary a little bit.. we have to do a good amount of config admin work with application names / component names in konflux-release-dev repo for releases. So why exactly are we doing this? or what is the problem right now that this would solve?

PR #1108, opened 2024-11-05
Don't close 'customer portal' subtask automatically

thiagoalessio:
Release present in cincinnati fast channel doesn't mean that it is automatically visible on the customer portal. I'd rather check and close that subtask manually.


joepvd (approved):
/lgtm

PR #1110, opened 2024-11-06
ART-10882: Get RHCOS to build on new host

joepvd:


pyartcd/pyartcd/pipelines/build_rhcos.py
```diff
@@ -47,6 +45,9 @@ def __init__(self, runtime: Runtime, new_build: bool, ignore_running: bool, vers
         self.api_token = None
         self._stream = None  # rhcos stream the version maps to
         self.dry_run = self.runtime.dry_run
+        self.jenkins_base_url = "https://jenkins-rhcos--prod-pipeline.apps.int.prod-stable-spoke1-dc-iad2.itup.redhat.com"
+        if self.version in ["4.15", "4.16", "4.17", "4.18", "4.19"]:
```
Ximinhan, line 49: suggest using version compare

pyartcd/pyartcd/pipelines/build_rhcos.py
```diff
@@ -47,6 +45,9 @@ def __init__(self, runtime: Runtime, new_build: bool, ignore_running: bool, vers
         self.api_token = None
         self._stream = None  # rhcos stream the version maps to
         self.dry_run = self.runtime.dry_run
+        self.jenkins_base_url = "https://jenkins-rhcos--prod-pipeline.apps.int.prod-stable-spoke1-dc-iad2.itup.redhat.com"
```
Ximinhan, line 48: suggest putting this URL in the constants

joepvd: It used to be a module constant, which is appropriate, as this is the only place that consumes it. 
I intend to restore that next week once we get the go ahead to produce all builds on the new instance.

Ximinhan: ok

PR #1115, opened 2024-11-11
[konflux] skip-checks param

ashwindasr:
Till now we had been skipping post build steps so that we can get the builds working. But skipping them violates enterprise contract policy. So now is a good time to enable the checks and the hash out any issues if any.

But keeping the skip checks as a param if we need it in the future.


locriandev (commented):
/lgtm

PR #1117, opened 2024-11-12
[konflux] skip-checks param fix

ashwindasr:


ashwindasr:
Fix included in https://github.com/openshift-eng/art-tools/pull/1116

PR #1118, opened 2024-11-13
[konflux] add arches param

ashwindasr:
We want to limit the arches that we build as a param. But keeping s390x, and ppc64le commented out for now so that we don't accidentally build them and cause issues for RHEL


doozer/doozerlib/backend/konflux_image_builder.py
```diff
@@ -103,7 +103,13 @@ async def build(self, metadata: ImageMetadata):
             if failed_parents:
                 raise IOError(f"Couldn't build {metadata.distgit_key} because the following parent images failed to build: {', '.join(failed_parents)}")
 
-            arches = metadata.get_arches()
+            # Use the override from jenkins param if its present else use the arches from the component metadata
+            if metadata.runtime.arches:
```
thegreyd, line 107: I don't think we need to check this, https://github.com/openshift-eng/art-tools/blob/f183313d6dd4bbda33d801361a365fd434e3c7ef/doozer/doozerlib/metadata.py#L289 get_arches() will return appropriately

ashwindasr: Thanks. Tested, works.

PR #1125, opened 2024-11-15
improve release message email

ashwindasr:
To clarify the location of reference releases


joepvd:
/lgtm
/approve

May want https://issues.redhat.com/browse/ART-11382 later, but this is already an improvement.

PR #1129, opened 2024-11-15
Sync pullspec to latest only when rpm for assembly is present in latest

thegreyd:
Test runs:

Do not sync to latest when check fails:
https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fbuild-microshift-bootc/15/console

Sync to latest when check succeeds:
https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fbuild-microshift-bootc/16/console

### DIFF
--- a/pyartcd/pyartcd/pipelines/build_microshift.py
+++ b/pyartcd/pyartcd/pipelines/build_microshift.py
@@ -188,7 +188,8 @@ class BuildMicroShiftPipeline:
         major, minor = self._ocp_version
         version = f'{major}.{minor}'
         try:
-            jenkins.start_microshift_sync(version=version, assembly=self.assembly, dry_run=self.runtime.dry_run)
+            jenkins.start_microshift_sync(version=version, assembly=self.assembly,
+                                          dry_run=self.runtime.dry_run, block_until_complete=True)
             message = f"microshift_sync for version {version} and assembly {self.assembly} has been triggered\n" \
                       f"This will publish the microshift build to mirror"
             await self.slack_client.say_in_thread(message)
--- a/pyartcd/pyartcd/pipelines/build_microshift_bootc.py
+++ b/pyartcd/pyartcd/pipelines/build_microshift_bootc.py
@@ -104,12 +104,12 @@ class BuildMicroShiftBootcPipeline:
             await self.sync_to_mirror(arch, bootc_build.el_target, f"{repo_url}@{digest}")
 
     async def is_microshift_for_release_in_latest(self, packages_path):
-        cmd = ["aws", "s3", "ls", packages_path]
+        cmd = ["aws", "s3", "ls", f"s3://art-srv-enterprise{packages_path}"]
         try:
             _, out, _ = await exectools.cmd_gather_async(cmd)
         except Exception:
             return False
-        if self.microshift_nvr in out:
+        if out and self.microshift_nvr in out:
             return True
         return False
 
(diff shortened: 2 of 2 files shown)


pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -188,7 +188,8 @@ async def _trigger_microshift_sync(self):
         major, minor = self._ocp_version
         version = f'{major}.{minor}'
         try:
-            jenkins.start_microshift_sync(version=version, assembly=self.assembly, dry_run=self.runtime.dry_run)
+            jenkins.start_microshift_sync(version=version, assembly=self.assembly,
+                                          dry_run=self.runtime.dry_run, block_until_complete=True)
```
Ximinhan, line 192: what dose the `block_until_complete` used for?

pyartcd/pyartcd/pipelines/build_microshift_bootc.py
```diff
@@ -81,6 +82,10 @@ async def run(self):
             data_path=self._doozer_env_vars["DOOZER_DATA_PATH"]
         )
         self.assembly_type = get_assembly_type(self.releases_config, self.assembly)
+
+        microshift_nvrs = await get_microshift_builds(self.group, self.assembly, env=self._elliott_env_vars)
```
Ximinhan, line 86: for a named assembly, isn't the microshift nvr already pined in the assembly definition? why need to find builds again

pyartcd/pyartcd/pipelines/build_microshift_bootc.py
```diff
@@ -140,7 +157,16 @@ async def _run_for(s3_path):
             await exectools.cmd_assert_async(cmd)
 
         await _run_for(release_path)
-        await _run_for(latest_path)
+
+        # make sure that the latest path has the same microshift build as the given release assembly
+        # only then sync the pullspec to the latest path
+        latest_packages_path = f"{latest_path}/os/Packages/"
+        if await self.is_microshift_for_release_in_latest(latest_packages_path):
```
Ximinhan, line 164: Will the microshift_bootc job triggered by microshift job? if its triggered by microshift job can this be a parameter of the job to update latest?

thegreyd: Yes the `build-microshift-bootc` job is triggered by the `build-microshift` job.
So you're proposing that build-microshift job determine if latest needs update or not? And then sets the flag
Or are you proposing that someone set that flag by hand?
The enhancement here is to remove the manual part, i.e. we can do this basic check and only 
update latest when rpms are sync'd first

pyartcd/pyartcd/pipelines/build_microshift.py
thegreyd: It blocks the parent job until this child job is complete

pyartcd/pyartcd/pipelines/build_microshift_bootc.py
Ximinhan: I suggest we can directly load the nvr from assembly, the get_microshift_builds() is gong to invoke find-builds and the end result is the same as pined rpm.

pyartcd/pyartcd/pipelines/build_microshift_bootc.py
Ximinhan: I kind of get the propose you block microshift-sync job above the wait it populates the path? 
yeah, I think set a `--latest` arg will be simpler

PR #1139, opened 2024-11-20
Refactor batch handling and exclude microshift from batches

vfreex:
1. Move batch handling logic into separate functions.
2. Don't include microshift advisory in the batch.
3. Make sure batch is always unlocked before creating a advisory.


artcommon/artcommonlib/util.py
```diff
@@ -127,21 +127,19 @@ def is_future_release_date(date_str):
...
                 # convert date format for advisory usage, 2024-02-13 -> 2024-Feb-13
-                assembly_release_date = datetime.strptime(release['date_start'], "%Y-%m-%d").strftime("%Y-%b-%d")
-                break
-        return assembly_release_date
-
+                return datetime.strptime(release['date_start'], "%Y-%m-%d").strftime("%Y-%b-%d")
     except KeyError:
-        return None
+        pass
+    raise ValueError(f'Assembly release date not found for {assembly}')
```
vfreex, line 142: Would not like this function to return `None`. We easily forget to check it.

vfreex:
Tested with https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fprepare-release/458/console

thegreyd:
@vfreex I checked https://errata.devel.redhat.com/batches/3423 and I see it has microshift advisory.. should it not be auto-removed from it?

thegreyd:
In general code lgtm

pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -156,62 +159,49 @@ async def run(self):
...
+                    await self._slack_client.say_in_thread(
+                        f"{ad} advisory created with release date {release_date}")
+                    continue
+                batch_id = 0
+                if batch and ad != "microshift":
+                    # Set batch_id for advisories other than microshift
+                    batch_id = int(batch["id"])
+                    # Ensure that the batch is unlocked
+                    batch = await self._ensure_batch_status(self._errata_api, batch, lock=False, dry_run=self.dry_run)
+                advisories[ad] = self.create_advisory(advisory_type=advisory_type,
```
thegreyd, line 201: This makes me think if we should have a lightweight Advisory and Batch class and then it can internally call `batch.unlock()` before `create()`. Strictly optional suggestion/thought

vfreex:
@thegreyd Thanks

PR #1147, opened 2024-11-22
[ART-10837] Exclude microshift advisory from batch

vfreex:


vfreex:
https://github.com/openshift-eng/art-tools/pull/1139 forgot this line...

PR #1148, opened 2024-11-22
Introduce build plan for konflux builds

locriandev:
Instead of crashing the pipeline on rebase failures, exclude those images from the build step.

Prepare ocp4-konflux to accept an image build strategy similar to what we have in ocp4, where can build all images, a pinned subset, or exclude only specific images.

Do not spam us about rebase failures (would happen too often at the moment)


vfreex (commented):
/lgtm

locriandev:
/approve

Test build: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-konflux/343/console

PR #1150, opened 2024-11-25
Introduce regex matching in konflux_db.search_builds_by_fields()

locriandev:
In `konflux_db.search_builds_by_fields()` we currently use LIKE to perform simple pattern matching using wildcard characters. This is limited to  basic wildcard patterns (`%` to match zero or more characters, `_` to match exactly one character).

Replace `LIKE` with `REGEXP_CONTAINS` to allow more complex regular expression matching. Supported Features are:

- Anchors (^ for the beginning of a string, $ for the end).
- Character classes (e.g., [abc], [a-z]).
- Quantifiers (e.g., *, +, ?, {n,m}).
- Alternation (| for "or" conditions).


joepvd (approved):
/lgtm

PR #1152, opened 2024-11-25
Konflux: Add OLM bundle rebase/build support

vfreex:
Test command:
```sh
doozer --assembly=test --group=openshift-4.18 --latest-parent-version beta:images:konflux:bundle --konflux-kubeconfig=/path/to/kubeconfig -
-konflux-namespace=ocp-art-tenant --image-repo=quay.io/openshift-release-dev/ocp-v4.0-art-dev-test ose-metallb-operator-container-v4.18.0-202410251504.p0.gc
5a3b07.assembly.test.el9
```

Test build: https://konflux.apps.kflux-ocp-p01.7ayg.p1.openshiftapps.com/application-pipeline/workspaces/ocp-art/applications/openshift-4-18/pipelineruns/ose-metallb-operator-bundle-container-s2724

### DIFF
--- a/artcommon/artcommonlib/konflux/konflux_db.py
+++ b/artcommon/artcommonlib/konflux/konflux_db.py
@@ -16,6 +16,8 @@ from artcommonlib.konflux import konflux_build_record
 from artcommonlib.konflux.konflux_build_record import KonfluxBuildOutcome, ArtifactType, KonfluxRecord, Engine
 
 SCHEMA_LEVEL = 1
+DEFAULT_SEARCH_WINDOW = 90
+DEFAULT_SEARCH_DAYS = 360  # By default, search for the last 360 days of data
 
 
 class KonfluxDb:
@@ -109,67 +111,109 @@ class KonfluxDb:
         with concurrent.futures.ThreadPoolExecutor() as pool:
             await asyncio.gather(*(loop.run_in_executor(pool, self.add_build, build) for build in builds))
 
-    async def search_builds_by_fields(self, start_search: datetime, end_search: datetime = None,
-                                      where: typing.Dict[str, typing.Any] = None, extra_patterns: dict = None,
-                                      order_by: str = '', sorting: str = 'DESC', limit: int = None) -> RowIterator:
+    async def search_builds_by_fields(
+            self,
+            start_search: typing.Optional[datetime] = None,
+            end_search: typing.Optional[datetime] = None,
+            window_size: int = DEFAULT_SEARCH_WINDOW,
+            where: typing.Optional[typing.Dict[str, typing.Any]] = None,
+            extra_patterns: typing.Optional[dict] = None,
+            order_by: str = '',
+            sorting: str = 'DESC',
+            limit: typing.Optional[int] = None,
+    ):
         """
         Execute a SELECT * from the BigQuery table.
 
         "where" is an optional dictionary that maps names and values to define a WHERE clause.
-        "start_search" is a lower bound to be applied to the partitioning field `start_time`.
-        "end_search" can optionally be provided as an upper bound for the same field.
+        "start_search" is a lower bound to be applied to the partitioning field `start_time`. If None, the search starts 360 days ago.
+        "end_search" can optionally be provided as an upper bound for the same field. If None, the search ends now.
+        "window_size" is the number of days to search in each iteration. If 0, the whole period is searched at once.
+        "extra_patterns" is an optional dictionary that maps names and values to define extra patterns to be matched.
+        "order_by" is the column to order by.
+        "sorting" is the sorting order.
+        "limit" is the maximum number of results to return. None for no limit.
 
-        Return a (possibly empty) list of results
+        Return a generator that yields KonfluxRecord objects.
         """
 
-        where_clauses = [
-            Column('start_time', DateTime) >= start_search,
-        ]
+        if start_search and end_search and start_search >= end_search:
+            raise ValueError(f"start_search {start_search} must be earlier than end_search {end_search}")
+        end_search = end_search or datetime.now(tz=timezone.utc)
+        start_search = start_search or end_search - timedelta(days=DEFAULT_SEARCH_DAYS)
+        if window_size < 0:
+            raise ValueError(f"search_window {window_size} must be a non-negative integer")
+        elif window_size == 0:
+            # If window_size is 0, search the whole period at once
+            window_size = (end_search - start_search).days
 
-        if end_search:
-            where_clauses.append(Column('start_time', DateTime) < end_search)
-
-        where = where if where else {}
+        base_clauses = []
+        where = where or {}
         for col_name, col_value in where.items():
             if col_value is not None:
-                where_clauses.append(Column(col_name, String) == col_value)
+                base_clauses.append(Column(col_name, String) == col_value)
             else:
-                where_clauses.append(Column(col_name, String).is_(None))
-
-        extra_patterns = extra_patterns if extra_patterns else {}
+                base_clauses.append(Column(col_name, String).is_(None))
+        extra_patterns = extra_patterns or {}
         for col_name, col_value in extra_patterns.items():
             regexp_condition = func.REGEXP_CONTAINS(Column(col_name, String), col_value)
-            where_clauses.append(regexp_condition)
+            base_clauses.append(regexp_condition)
 
         order_by_clause = Column(order_by if order_by else 'start_time', quote=True)
         order_by_clause = order_by_clause.desc() if sorting == 'DESC' else order_by_clause.asc()
 
-        results = await self.bq_client.select(
-            where_clauses=where_clauses,
-            order_by_clause=order_by_clause,
-            limit=limit)
+        # Table is partitioned by start_time. Perform an iterative search within given days of windows, going back to start_search
+        total_rows = 0
+        for window in range(0, (end_search - start_search).days, window_size):
+            end_window = end_search - timedelta(days=window)
+            start_window = max(end_window - timedelta(days=window_size), start_search)
+            where_clauses = base_clauses + [Column('start_time', DateTime) >= start_window, Column('start_time', DateTime) < end_window]
+            rows = await self.bq_client.select(
+                where_clauses=where_clauses,
+                order_by_clause=order_by_clause,
+                limit=limit-total_rows if limit is not None else None)
+            self.logger.debug('Found %s builds in batch %s', rows.total_rows, (window // window_size) + 1)
+            for row in rows:
+                total_rows += 1
+                yield self.from_result_row(row)
+                if limit is not None and total_rows >= limit:
+                    return
 
-        self.logger.debug('Found %s builds', results.total_rows)
-        return results
-
-    async def get_latest_builds(self, names: typing.List[str], group: str,
-                                outcome: KonfluxBuildOutcome = KonfluxBuildOutcome.SUCCESS, assembly: str = 'stream',
-                                el_target: str = None, artifact_type: ArtifactType = None,
-                                engine: Engine = None, completed_before: datetime = None)\
-            -> typing.List[konflux_build_record.KonfluxRecord]:
+    async def get_latest_builds(
+            self,
+            names: typing.List[str],
+            group: str,
+            outcome: KonfluxBuildOutcome = KonfluxBuildOutcome.SUCCESS,
+            assembly: str = 'stream',
+            el_target: typing.Optional[str] = None,
+            artifact_type: typing.Optional[ArtifactType] = None,
+            engine: typing.Optional[Engine] = None,
+            completed_before: typing.Optional[datetime] = None,
+            extra_patterns: dict = {},
+            strict: bool = False,
+    ) -> typing.List[typing.Optional[KonfluxRecord]]:
         """
         For a list of component names, run get_latest_build() in a concurrent pool executor.
         """
 
         return await asyncio.gather(*[
-            self.get_latest_build(name, group, outcome, assembly, el_target, artifact_type, engine, completed_before)
+            self.get_latest_build(name, group, outcome, assembly, el_target, artifact_type, engine, completed_before,
+                                  extra_patterns, strict)
             for name in names])
 
-    async def get_latest_build(self, name: str, group: str, outcome: KonfluxBuildOutcome = KonfluxBuildOutcome.SUCCESS,
-                               assembly: str = 'stream', el_target: str = None, artifact_type: ArtifactType = None,
-                               engine: Engine = None, completed_before: datetime = None, extra_patterns: dict = {},
-                               strict: bool = False) \
-            -> typing.Optional[konflux_build_record.KonfluxRecord]:
+    async def get_latest_build(
+            self,
+            name: str,
+            group: str,
+            outcome: KonfluxBuildOutcome = KonfluxBuildOutcome.SUCCESS,
+            assembly: str = 'stream',
+            el_target: typing.Optional[str] = None,
+            artifact_type: typing.Optional[ArtifactType] = None,
+            engine: typing.Optional[Engine] = None,
+            completed_before: typing.Optional[datetime] = None,
+            extra_patterns: dict = {},
+            strict: bool = False,
+    ) -> typing.Optional[KonfluxRecord]:
         """
         Search for the latest Konflux build information in BigQuery.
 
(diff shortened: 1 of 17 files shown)


vfreex:
@ashwindasr @ashwindasr @locriandev This PR is ready to review if you guys are interested in, while I am in the progress of splitting it into smaller commits and adding unit tests.

artcommon/artcommonlib/konflux/konflux_db.py
```diff
@@ -234,10 +237,40 @@ async def get_latest_build(self, name: str, group: str, outcome: KonfluxBuildOut
...
+
+        :param nvr: The NVR of the build.
+        :param outcome: The outcome of the build.
+        :param strict: If True, raise an exception if the build record is not found.
+        :return: The build record; None if the build record is not found.
+        :raise: IOError if the build record is not found and strict is True.
+        """
+        where = {"nvr": nvr, "outcome": str(outcome)}
+        started_before = datetime.now(tz=timezone.utc)
+        for window in range(12):
```
locriandev, line 257: Since `start_time` is something we always have to provide `search_builds_by_fields()` with, I think it could useful to embed this window search into that function. the `start_search` param could be made optional, and if None we would go through this loop. this way, you can get rid of the new `get_build_record_by_nvr()` function and just call `search_builds_by_fields()` appropriately

doozer/doozerlib/cli/images_konflux.py
```diff
@@ -188,3 +196,192 @@ async def images_konflux_build(
...
+        self.dry_run = dry_run
+        self.konflux_kubeconfig = konflux_kubeconfig
+        self.konflux_context = konflux_context
+        self.konflux_namespace = konflux_namespace
+        self.image_repo = image_repo
+        self.skip_checks = skip_checks
+        self.release = release
+
+    @start_as_current_span_async(TRACER, "images:konflux:bundle")
+    async def run(self):
```
locriandev, line 227: I suggest splitting it into smaller functions to enhance readability and maintainability

doozer/doozerlib/cli/images_konflux.py
```diff
@@ -188,3 +196,192 @@ async def images_konflux_build(
...
+            db=db_for_bundles,
+            konflux_namespace=self.konflux_namespace,
+            konflux_kubeconfig=self.konflux_kubeconfig,
+            konflux_context=self.konflux_context,
+            image_repo=self.image_repo,
+            skip_checks=self.skip_checks,
+            dry_run=self.dry_run,
+        )
+
+        async def _rebase_and_build(image_meta: ImageMetadata, operator_build: KonfluxBuildRecord):
```
locriandev, line 317: Is there a strong need for an embedded function? moving it at the class scope would facilitate readability and testing

doozer/doozerlib/util.py
```diff
@@ -550,6 +550,29 @@ def oc_image_info(pull_spec: str, go_arch: str = 'amd64') -> Dict:
     return json.loads(out)
 
 
+@retry(reraise=True, stop=stop_after_attempt(3), wait=wait_fixed(3))
+async def oc_image_info_async(pull_spec: str, go_arch: str = 'amd64') -> Dict:
```
locriandev, line 554: already defined [here](https://github.com/openshift-eng/art-tools/pull/1152/files#diff-4504181f403e20b264f5ab2a4d5706e3859873b729a2cda5048b2b245a46d795R587)

doozer/doozerlib/util.py
```diff
@@ -550,6 +550,29 @@ def oc_image_info(pull_spec: str, go_arch: str = 'amd64') -> Dict:
...
+    during the course of doozer's execution.
+    """
+    # Filter by os because images can be multi-arch manifest lists (which cause oc image info to throw an error if not filtered).
+    cmd = ['oc', 'image', 'info', f'--filter-by-os={go_arch}', '-o', 'json', pull_spec]
+    _, out, _ = await exectools.cmd_gather_async(cmd)
+    return json.loads(out)
+
+
+@alru_cache(maxsize=1000)
+async def oc_image_info_async__caching(pull_spec: str, go_arch: str = 'amd64') -> Dict:
```
locriandev, line 567: already defined [here](https://github.com/openshift-eng/art-tools/pull/1152/files#diff-4504181f403e20b264f5ab2a4d5706e3859873b729a2cda5048b2b245a46d795R601)

vfreex:
@locriandev Thanks for the review. I've updated the PR to address your suggestions. PTAL.

artcommon/artcommonlib/konflux/konflux_db.py
vfreex: I feel it might be reasonable to have `search_builds_by_fields` only search for once while adding another wrapper function to do windowed searches. It will also keep the function simpler. 
I added `search_builds_by_fields_with_window`

artcommon/artcommonlib/konflux/konflux_db.py
```diff
@@ -152,11 +152,54 @@ async def search_builds_by_fields(self, start_search: datetime, end_search: date
         self.logger.debug('Found %s builds', results.total_rows)
         return results
 
+    async def search_builds_by_fields_with_window(
```
locriandev, line 155: fine with me, but I think we could just have one single `search_builds_by_fields()` function that always scans within a moving window of fixed size. `start_time` should be optional, and if None just assume a very old date (you used 3 years, that can be adjusted if we want)

artcommon/artcommonlib/konflux/konflux_db.py
```diff
@@ -234,10 +280,32 @@ async def get_latest_build(self, name: str, group: str, outcome: KonfluxBuildOut
                 continue
 
         # If we got here, no builds have been found in the whole 36 months period
+        if strict:
+            raise ValueError(f"Build record for {name} not found.")
         self.logger.warning('No builds found for %s in %s with status %s in assembly %s and target %s',
                             name, group, outcome.value, assembly, el_target)
         return None
 
+    async def get_build_record_by_nvr(self, nvr: str, outcome: KonfluxBuildOutcome = KonfluxBuildOutcome.SUCCESS, strict: bool = True) -> typing.Optional[KonfluxRecord]:
```
locriandev, line 289: Once we unify the `search_builds_by_fields()` implementation, this function won't be needed as you can just search with no start_time param

artcommon/artcommonlib/konflux/konflux_db.py
```diff
@@ -152,11 +152,54 @@ async def search_builds_by_fields(self, start_search: datetime, end_search: date
...
+            where: typing.Optional[typing.Dict[str, typing.Any]] = None,
+            extra_patterns: typing.Optional[typing.Dict] = None,
+            order_by: str = '',
+            sorting: str = 'DESC',
+            limit: typing.Optional[int] = None) -> typing.AsyncGenerator[Row, None]:
+        """ Execute search_builds_by_fields() with a window of given days.
+
+        :param start_search: The start time of the search (inclusive). If None, the search starts 3 years ago.
+        :param end_search: The end time of the search (exclusive). If None, the search ends now.
+        :param search_window: The number of days in the search window. Default is 90 days.
```
locriandev, line 169: `get_latest_build()` defines a window size of 30 days. I think we should define a unique window size and store it as a constant at class or module level.

artcommon/artcommonlib/konflux/konflux_db.py
```diff
@@ -152,11 +152,54 @@ async def search_builds_by_fields(self, start_search: datetime, end_search: date
...
+        :param limit: The limit of the search. None for no limit.
+        :return: The search results.
+        """
+        # Table is partitioned by start_time. Perform an iterative search within given days of windows, going back to start_search
+        # at most. This will let us reduce the amount of scanned data (and the BigQuery usage cost), as in the vast
+        # majority of cases we would find a build in the first search_window interval.
+        if start_search and end_search and start_search >= end_search:
+            raise ValueError(f"start_search {start_search} must be earlier than end_search {end_search}")
+        end_search = end_search or datetime.now(tz=timezone.utc)
+        start_search = start_search or end_search - timedelta(days=365*3)
```
locriandev, line 183: `get_latest_build()` goes back to 1 years at most (12 windows * 30 days). We should define a unique threshold for all searches and store it in a centralized place as a constant.

artcommon/artcommonlib/runtime.py
```diff
@@ -54,6 +54,8 @@ def initialize_logging(self):
         self._logger = logging.getLogger('artcommonlib')
 
     def initialize_konxflux_db(self):
+        if self.initialized or self.konflux_db:
```
locriandev, line 57: I am not seeing where `self.initialized` is being set to True. It works anyway because you're checking for `self.konflux_db` to be != None, but maybe it makes sense to set `self.initialized = True` at the end of the initialization function

vfreex: I was following initialize_logging (https://github.com/openshift-eng/art-tools/blob/eed0c8e6d883a2c2cc6473f48f60f3a03dd86691/artcommon/artcommonlib/runtime.py#L38). However we expect self.initialized to be set by the child class's initialize function.

I don't plan to make a lot of changes for the Runtime. It seems make more sense to me to remove the check for self.initialized.

artcommon/artcommonlib/konflux/konflux_db.py
vfreex: Done merging into one single `search_builds_by_fields`

artcommon/artcommonlib/konflux/konflux_db.py
vfreex: I added 2 constants for both of the functions.

vfreex:
@locriandev I pushed another update.

vfreex:
unit tests added

artcommon/artcommonlib/konflux/konflux_db.py
```diff
@@ -109,66 +111,109 @@ async def add_builds(self, builds: typing.List[konflux_build_record.KonfluxBuild
         with concurrent.futures.ThreadPoolExecutor() as pool:
             await asyncio.gather(*(loop.run_in_executor(pool, self.add_build, build) for build in builds))
 
-    async def search_builds_by_fields(self, start_search: datetime, end_search: datetime = None,
-                                      where: typing.Dict[str, typing.Any] = None, extra_patterns: dict = None,
-                                      order_by: str = '', sorting: str = 'DESC', limit: int = None) -> RowIterator:
```
locriandev, line 114: I would keep the `RowIterator` type hint for the return type, as it tells us that the return value of this function must be looped over to start constructing the build record objects

artcommon/artcommonlib/konflux/konflux_db.py
```diff
@@ -109,66 +111,109 @@ async def add_builds(self, builds: typing.List[konflux_build_record.KonfluxBuild
...
-
-        if end_search:
-            where_clauses.append(Column('start_time', DateTime) < end_search)
-
-        where = where if where else {}
+        if start_search and end_search and start_search >= end_search:
+            raise ValueError(f"start_search {start_search} must be earlier than end_search {end_search}")
+        end_search = end_search or datetime.now(tz=timezone.utc)
+        start_search = start_search or end_search - timedelta(days=DEFAULT_SEARCH_DAYS)
+        if window_size < 0:
```
locriandev, line 144: I think this might be clearer if we enforce `window_size` to be > 0, and default it to None
```
    async def search_builds_by_fields(
            ...
            window_size: int = None,
...
        window_size =  window_size if window_size else DEFAULT_SEARCH_WINDOW
        assert window_size > 0, f"search_window {window_size} must be a non-negative integer"
```

artcommon/artcommonlib/konflux/konflux_db.py
locriandev: It seems to me all `get_build_record_by_nvr()` does is call `search_builds_by_fields` and raise an IOError if `strict` is True. I would rather avoid defining extra functions to search for specific fields, because if we follow with this approach more function could be defined in the future, and their implementation would eventually diverge leaving us with a codebase that's hard to maintain and keep consistent.

Instead, I think we can give the `strict` param to `search_builds_by_fields` (default to False), and only leave the `get_build_records_by_nvrs()` implementation which is more specific.

artcommon/artcommonlib/konflux/konflux_db.py
```diff
@@ -213,15 +261,15 @@ async def get_latest_build(self, name: str, group: str, outcome: KonfluxBuildOut
...
             base_clauses.append(Column(col_name, String).like(f"%{col_value}%"))
 
-        started_before = datetime.now(tz=timezone.utc) if not completed_before else completed_before
-        for window in range(12):
-            end_search = started_before - window * 3 * timedelta(days=30)
-            start_search = end_search - 3 * timedelta(days=30)
-
+        end_search = datetime.now(tz=timezone.utc) if not completed_before else completed_before
+        start_search = end_search - timedelta(days=DEFAULT_SEARCH_DAYS)
+        for window in range(0, (end_search - start_search).days, DEFAULT_SEARCH_WINDOW):
```
locriandev, line 266: `end_search - start_search` is always equal to `DEFAULT_SEARCH_DAYS`

artcommon/artcommonlib/konflux/konflux_db.py
vfreex: `self.from_result_row` is called for each row in this function, so the return type is actually an AsyncGenerator of KonfluxRecord. I'll add type hint to make it more obvious

artcommon/artcommonlib/konflux/konflux_db.py
vfreex: makes sense

artcommon/artcommonlib/konflux/konflux_db.py
vfreex: sounds good to me

artcommon/artcommonlib/konflux/konflux_db.py
```diff
@@ -213,15 +264,15 @@ async def get_latest_build(self, name: str, group: str, outcome: KonfluxBuildOut
...
+            start_window = max(end_window - timedelta(days=DEFAULT_SEARCH_WINDOW), start_search)
             where_clauses = copy.copy(base_clauses)
             where_clauses.extend([
-                Column('start_time', DateTime) >= start_search,
-                Column('start_time', DateTime) < end_search,
+                Column('start_time', DateTime) >= start_window,
+                Column('start_time', DateTime) < end_window,
             ])
 
             results = await self.bq_client.select(where_clauses, order_by_clause=order_by_clause, limit=1)
```
locriandev, line 278: same as above, it would be nice to have unit tests here.

artcommon/artcommonlib/konflux/konflux_db.py
```diff
@@ -109,66 +111,112 @@ async def add_builds(self, builds: typing.List[konflux_build_record.KonfluxBuild
...
             await asyncio.gather(*(loop.run_in_executor(pool, self.add_build, build) for build in builds))
 
-    async def search_builds_by_fields(self, start_search: datetime, end_search: datetime = None,
-                                      where: typing.Dict[str, typing.Any] = None, extra_patterns: dict = None,
-                                      order_by: str = '', sorting: str = 'DESC', limit: int = None) -> RowIterator:
+    async def search_builds_by_fields(
+            self,
+            start_search: typing.Optional[datetime] = None,
+            end_search: typing.Optional[datetime] = None,
+            window_size: int | None = None,
```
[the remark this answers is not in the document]
vfreex: No differences. Changed.

artcommon/artcommonlib/konflux/konflux_db.py
```diff
@@ -109,66 +111,112 @@ async def add_builds(self, builds: typing.List[konflux_build_record.KonfluxBuild
...
-                                el_target: str = None, artifact_type: ArtifactType = None,
-                                engine: Engine = None, completed_before: datetime = None)\
-            -> typing.List[konflux_build_record.KonfluxBuildRecord]:
+        # Table is partitioned by start_time. Perform an iterative search within given days of windows, going back to start_search
+        total_rows = 0
+        for window in range(0, (end_search - start_search).days, window_size):
+            end_window = end_search - timedelta(days=window)
+            start_window = max(end_window - timedelta(days=window_size), start_search)
+            where_clauses = base_clauses + [Column('start_time', DateTime) >= start_window, Column('start_time', DateTime) < end_window]
+            rows = await self.bq_client.select(
```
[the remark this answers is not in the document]
vfreex: Added `test_search_builds_by_fields_windowed`

artcommon/artcommonlib/konflux/konflux_db.py
vfreex: Added test_get_latest_build_windowed

locriandev (commented):
/lgtm

PR #1161, opened 2024-11-28
Fix batch creation

vfreex:
The returned object of `AsyncErrataAPI.create_batch` is flattened since https://github.com/openshift-eng/art-tools/pull/1139. Getting the "data" field is no longer needed.


locriandev (approved):
/lgtm

PR #1162, opened 2024-11-28
move yaml lib into artcommon lib

Ximinhan:
Add yaml_handler function to artcommon lib to set yaml config and load by others to reduce duplication
also set indent config for yaml, this will change list item
```
arches:
- x86_64
- ppc64le
- s390x
- aarch64
```
to
```
arches:
  - x86_64
  - ppc64le
  - s390x
  - aarch64
```
for better reading

### DIFF
--- a/artcommon/artcommonlib/util.py
+++ b/artcommon/artcommonlib/util.py
@@ -34,7 +34,17 @@ def remove_suffix(s: str, suffix: str) -> str:
         return s[:]
 
 
-def yaml_handler():
+def new_roundtrip_yaml_handler():
+    """
+    Creates and returns a configured YAML handler with specific formatting settings.
+    Returns:
+        YAML: A YAML handler configured with:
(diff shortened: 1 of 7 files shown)


artcommon/artcommonlib/util.py
```diff
@@ -34,6 +34,15 @@ def remove_suffix(s: str, suffix: str) -> str:
         return s[:]
 
 
+def yaml_handler():
```
vfreex, line 37: This function needs more describable name, like `new_roundtrip_yaml_processor`

vfreex, line 37: This function needs more describable name, like `new_roundtrip_yaml_processor`

Ximinhan: updated

vfreex (approved):
/lgtm

PR #1164, opened 2024-11-28
Start RHCOS arch jobs earlier

joepvd:


jlebon (approved):
Basically the downside of this is that any failure related to building artifacts or testing the live artifacts will now happen _after_ we've already forked off the other arches. Which means that instead of getting one failure in the pipeline, we'll get four.

SGTM if you want to try this, but it's a trade-off, and we might decide to toggle it back off or refine it (so that it's only enabled e.g. once we're past a certain milestone).

This also relates to https://github.com/openshift/enhancements/pull/1637. A big part of the churn are the OpenShift packages, and it's going to be much cheaper to rebuild just the OCP layer on top.

dustymabe:
> Basically the downside of this is that any failure related to building artifacts or testing the live artifacts will now happen _after_ we've already forked off the other arches. Which means that instead of getting one failure in the pipeline, we'll get four.

It also has the side effect of making things more confusing and also wrong. For example there is state that gets uploaded to the builds dir and now the builds browser shows artifacts and build IDs from failed builds. It also causes new triggered builds to detect "No new build" and then do nothing OR try to fix multi-arch builds. 

Overall I would say the benefits aren't worth the drawbacks right now.

dustymabe:
revert PR in https://github.com/openshift-eng/art-tools/pull/1310

PR #1167, opened 2024-12-02
[konflux] set build retires to 3

ashwindasr:
Now that we have our own cluster, we can reduce the number of retires to 3, as it was originally set. If builds fail here, its most likely a legit issue which warrants further investigation and not a flake, so no need for additional retires


locriandev (approved):
/lgtm

PR #1170, opened 2024-12-03
Make quay-doomsday-backup async and send its results to Slack

adobes1:
Make `quay-doomsday-backup` asynchronous
- For running AWS and mirror commands, use `AsyncRetrying` from `tenacity` directly because `cmd_assert_async` does not contain retry logic and to avoid having to create a wrapper function around `cmd_assert_async` to use with `tenacity.retry` decorator

Sending results to Slack
- Create a new thread in a version specific release channel and send updates there
- After the sync of all arches either succeeds or fails, send final message to the thread and to the channel
- Modify `say()` and `say_in_thread()` from `pyartcd/slack.py` to include optional parameter -`broadcast` - to send thread replies to the channel as well
- Example of a successful run [here](https://redhat-internal.slack.com/archives/C05HB1F1THP/p1733224749354249)
- Example of a failed run [here](https://redhat-internal.slack.com/archives/C05HB1F1THP/p1733225204280459)

Also fix the working directory path which is deleted at the end

### DIFF
--- a/pyartcd/pyartcd/pipelines/quay_doomsday_backup.py
+++ b/pyartcd/pyartcd/pipelines/quay_doomsday_backup.py
@@ -25,10 +25,12 @@ class QuayDoomsdaySync:
         self.runtime = runtime
         self.version = version
         self.workdir = "./workspace"
+        self.slack_client = self.runtime.new_slack_client()
+        self.slack_client.bind_channel(version)
 
         self.arches = arches.split(",") if arches else ALL_ARCHES_LIST
 
-    async def sync_arch(self, arch: str):
+    async def sync_arch(self, arch: str) -> bool:
         if arch not in ALL_ARCHES_LIST:
             raise Exception(f"Invalid arch: {arch}")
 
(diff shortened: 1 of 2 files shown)


ashwindasr (changes requested):

pyartcd/pyartcd/pipelines/quay_doomsday_backup.py
```diff
@@ -44,35 +48,66 @@ def sync_arch(self, arch: str):
...
+        retry = AsyncRetrying(reraise=True, stop=stop_after_attempt(N_RETRIES))
+        try:
+            self.runtime.logger.info("[%s] Running mirror command: %s", arch, mirror_cmd)
+            await retry(cmd_assert_async, mirror_cmd)
+            self.runtime.logger.info("[%s] Mirror command ran successfully", arch)
+            if self.runtime.dry_run:
+                self.runtime.logger.info("[DRY RUN] [%s] Would have run %s", arch, " ".join(aws_cmd))
+                self.runtime.logger.info("[DRY RUN] [%s] Would have messaged slack", arch)
+            else:
+                sleep(5)
```
vfreex, line 62: `time.sleep` will block the current coroutine. `await asyncio.sleep` should be used instead.

adobes1: Thank you! Fixed

pyartcd/pyartcd/pipelines/quay_doomsday_backup.py
```diff
@@ -44,35 +48,66 @@ def sync_arch(self, arch: str):
...
-    def run(self):
-        mkdirs(self.workdir)
-
-        for arch in self.arches:
-            self.runtime.logger.info("Now syncing arch %s", arch)
-            self.sync_arch(arch)
+        # Report the results to Slack
+        if not self.runtime.dry_run:
+            if all(results):
+                await self.slack_client.say_in_thread(":done_it_is: All arches synced successfully", broadcast=True)
```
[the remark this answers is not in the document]
adobes1: You are right, thank you. 
I've added a new method `add_reaction` to `slack.py` which is then used to add :done_it_is: on the main message (I find :done_it_is: to be clearer than :done:) like [this](https://redhat-internal.slack.com/archives/C05HB1F1THP/p1733752159971629)
However, the underlying Slack api function [reactions_add](https://api.slack.com/methods/reactions.add) requires channel ID instead of just channel name and getting the ID is a little bit tricky.
I've added another method (`get_channel_id`) to `slack.py` which tries to get the ID of the bound channel. It only works for private channels for now because there's too many public channels and we hit request ratelimit while searching through them.

vfreex (commented):
/lgtm
/approve

pyartcd/pyartcd/slack.py
```diff
@@ -42,15 +42,26 @@ def bind_channel(self, channel_or_release: Optional[str]):
         else:
             raise ValueError(f"Invalid channel_or_release value: {channel_or_release}")
 
-    async def say_in_thread(self, message: str, reaction: Optional[str] = None):
+    async def get_channel_id(self):
+        """
+        Get ID of the currently bound channel
+        """
+        # Can only get private channel ids for now to avoid longer waiting caused by request ratelimit
+        response = await self._client.conversations_list(types="private_channel", limit=999)
```
ashwindasr, line 50: @adobes1 I thought we discussed that we can get the channel ID directly from the message response, when we are posting to the channel?

Querying all channels from RH Slack is not efficient. PTAL and fix this once you're back. Reverting PR.

PR #1173, opened 2024-12-03
[konflux] add ignore locks function

ashwindasr:
Sometimes we need to run other builds even though the lock is held by one job, if we know for sure that there is no conflict.

Alongside https://github.com/openshift-eng/aos-cd-jobs/pull/4304


locriandev (commented):
/lgtm

PR #1175, opened 2024-12-04
ART-11446: images-health should report failures only when fail count > 1

mbiarnes:
rh-pre-commit.version: 2.3.2


locriandev:
/hold
this does not protect us from the case where only 1 failure is found

locriandev (commented):
/lgtm

PR #1182, opened 2024-12-05
allow go previous is none

Ximinhan:
GO_PREVIOUS will be removed so only have latest


joepvd:
We may need this in the future. Possibly better to query if this var is set, and act if so?

PR #1183, opened 2024-12-05
Konflux: Ignore ConflictError when creating a resource

vfreex:
Another process may create the same resource at the same time. Ignore this kind of error.


locriandev (commented):
/lgtm

ashwindasr (approved):

PR #1184, opened 2024-12-05
Deal with kernel stop ship tag better pre-GA

joepvd:
Example config to go with this: https://github.com/openshift-eng/ocp-build-data/pull/5795


joepvd:
/hold
timing needs care

locriandev (approved):
/lgtm

PR #1177, opened 2024-12-04
[konflux] update apply-tags timeout

ashwindasr:
microshift-bootc times out after the default '2h' for the apply-tags task. So increasing to '4h'

PR #1185, opened 2024-12-05
Revert "[konflux] update apply-tags timeout"

ashwindasr:
Reverts openshift-eng/art-tools#1177

Apply tags should only take a couple of minutes. Something else might be going on, so additional investigation might be required.


vfreex (commented):
/lgtm

vfreex (approved):

PR #1190, opened 2024-12-06
load batch relase from config

Ximinhan:
the batch release should align with advisory config, so load from erratatool.yml to get it aligned instead of using hardcode value.
In 4.13 the advisory release is updated to use "RHOSE ASYNC - Silent Release - AUTO"


vfreex (approved):
/lgtm

PR #1196, opened 2024-12-09
[konflux] set build source image as true

ashwindasr:
Enterprise Contract (EC) requires us to build source image always. This value was set in the yaml file, but the default value there is `false`. Setting that value in code as per current pattern.


vfreex (commented):
/lgtm
/approve

PR #1197, opened 2024-12-09
[konflux] pull template from art-konflux-template repo

ashwindasr:
https://github.com/openshift-priv/art-konflux-template repo was created and setup as the feedback loop with konflux. Konflux raises periodic PRs to that repo to update task SHAs (like [this](https://github.com/openshift-priv/art-konflux-template/pull/31)). We need to make sure that we use the new tasks to:
- Satisfy Enterprise Contract policy
- Make sure that we pick up bug fixes for those tasks that are being updated

Till now we had been manually bumping the task SHAs (like [this](https://github.com/openshift-eng/art-tools/pull/1195)) but since we are overriding the defaults in code, we can safely use the yaml from the art-konflux-template repo now.

Hence from now on, the template will be from https://github.com/openshift-priv/art-konflux-template/blob/main/.tekton/art-konflux-template-push.yaml

Test run [worked](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-konflux/477/console)


ashwindasr:
/hold

Until https://github.com/openshift-eng/art-tools/pull/1196 is merged. Can be unholded after

doozer/doozerlib/backend/konflux_client.py
```diff
@@ -296,9 +296,8 @@ def _new_pipelinerun_for_image_build(generate_name: str, namespace: Optional[str
         if additional_tags is None:
             additional_tags = []
         https_url = art_util.convert_remote_git_to_https(git_url)
-        # TODO: In the future the PipelineRun template should be loaded from a remote git repo.
-        template_content = files("doozerlib").joinpath("backend").joinpath("konflux_image_build_pipelinerun.yaml").read_text()
-        template = jinja2.Template(template_content, autoescape=True)
+        response = requests.get(constants.KONFLUX_PlR_TEMPLATE_URL)
```
vfreex, line 299: We don't have to pull the template for each PipelineRun. Better to cache it on start.

ashwindasr: Thanks - updated and retested. Works as expected

vfreex (commented):
/lgtm
/approve
Let's try it

PR #1207, opened 2024-12-11
images:list print just the count

ashwindasr:
Sometimes we just need the count, so adding a flag for that purpose


locriandev (approved):
/lgtm

locriandev (commented):
/hold

PR #1212, opened 2024-12-12
Improve scan-sources-konflux

locriandev:
It happened (due to manual build actions mostly) to have multiple builds from the same upstream commit. In this case, scan-sources for konflux compares the NVRs and finds them different (rebase times differ), so concluded that the image needed to be rebuilt. 

To avoid that, compare the commitish of the latest retrieved build with the most recent upstream commit hash. If they match, no need to rebuild even if NVRs differ. Otherwise, scan-sources can conclude that an upstream commit has been reverted, so we need a more recent build from the HEAD commit.


vfreex (commented):
/lgtm
/approve

PR #1222, opened 2024-12-17
doozer: Don't send HTTP requests in unit test

vfreex:
No actual HTTP requests to GitHub should be made by unit test. `what_is_in_master` function should be mocked.


joepvd (approved):
/lgtm

PR #1223, opened 2024-12-17
ART-11638 remove art-ci

Ximinhan:
remove art-ci on cluster


ocp-build-data-validator/validator/__main__.py
```diff
@@ -10,8 +10,12 @@
 def validate(file, exclude_vpn, schema_only):
     (parsed, err) = format.validate(open(file).read())
     if err:
-        msg = '{} is not a valid YAML\nReturned error: {}'.format(file, err)
-        support.fail_validation(msg, parsed)
+        print(f'{file} is not a YAML file')
```
vfreex, line 13: This will make the validator completely ignore syntax errors in yaml files. We need to either exclude non-yaml files from the invocation, or determine if a file is yaml or not in a better way.

ocp-build-data-validator/validator/__main__.py
```diff
@@ -10,8 +10,12 @@
...
+        return
+
+    if support.is_not_release(parsed):
++       print(f'Skipping validation of non release file {file}')
```
vfreex, line 17: Could you explain why non_release files should be skipped?

ocp-build-data-validator/validator/support.py
```diff
@@ -18,6 +19,10 @@ def is_disabled(parsed):
     return 'mode' in parsed and parsed['mode'] == 'disabled'
 
 
+def is_not_release(parsed)
+    return 'content' in parsed or parsed.get('for_release') == 'false'
```
vfreex, line 23: The condition doesn't seem right. Looks like it should be `content' in parsed and not parsed.get('for_release')`

ocp-build-data-validator/validator/support.py
```diff
@@ -18,6 +19,10 @@ def is_disabled(parsed):
     return 'mode' in parsed and parsed['mode'] == 'disabled'
 
 
+def is_not_release(parsed)
```
vfreex, line 22: missing `:`

ocp-build-data-validator/validator/__main__.py
Ximinhan: updated, used an other way to check

ocp-build-data-validator/validator/support.py
Ximinhan: updated, used an other way to check

ocp-build-data-validator/entrypoint.sh
```diff
@@ -4,8 +4,8 @@ pr=`echo $PR_URL | sed 's/github.com/api.github.com\/repos/g'| sed 's/pull/pulls
 git clone --branch `curl -s $pr | jq -r .head.ref` --single-branch --depth=1  `curl -s $pr | jq -r .head.repo.html_url`
 # cd to repo
 cd `curl -s $pr| jq -r .head.repo.name`
-# validate changed files and output to result.output
-curl -s $pr/files | jq -r .[].filename | xargs validate-ocp-build-data -s $1 2>&1 | tee result.output
+# validateonly .yml files in root, rpms/, or images/ directories and output to result.output
+curl -s $pr/files | jq -r .[].filename | grep -E '^[^/]+\.yml$|^(rpms|images)/.*\.yml$' | xargs validate-ocp-build-data -s $1 2>&1 | tee result.output
```
Ximinhan, line 8: align with https://github.com/openshift-eng/ocp-build-data/blob/openshift-4.19/.github/workflows/validate.yaml#L8-L10

vfreex (commented):
/lgtm
/approve

PR #1225, opened 2024-12-17
trigger rhcos_sync with shortname

ashwindasr:
Noticed that rhcos_sync jobs were [failing](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Frhcos_sync/115/parameters/) because we were passing the entire pull spec as param.


joepvd (approved):
/lgtm

PR #1227, opened 2024-12-18
Elliott: Fix unit tests that take too much time

vfreex:
`test_get_brew_build_failure` has a few issues that causes the execution takes too much time (> 30 seconds):
- `requests.get` is not property mocked, causing actual HTTP requests send to Errata.
- It retries 10 times on error. For each retry it waits for 3 seconds.

This PR fixes the issue and also updates `test_get_brew_build_success` and `test_get_brew_build_success_session` to follow the same pattern.


joepvd (approved):
/lgtm

PR #1231, opened 2024-12-18
ART-11446: images-health should report only failures

mbiarnes:
when there are more than one build

### DIFF
--- a/doozer/doozerlib/cli/images_health.py
+++ b/doozer/doozerlib/cli/images_health.py
@@ -107,6 +107,9 @@ class ImagesHealthPipeline:
             elif latest_success_idx > 1:
                 msg += (f'Last {self.url_text(latest_success_bi_task_url, "success")} '
                         f'was {latest_success_idx} attempts ago')
+            elif latest_success_idx == 1:
+                # Do nothing
+                pass
 
             # Append the Art-Dash link to the message
             msg += f'. {self.url_text(art_dash_link, "Details")}'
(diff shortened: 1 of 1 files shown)


locriandev (changes requested):
I don't think this is doing what we need. If only latest attempt failed, we will get a notification anyway. [This](https://github.com/openshift-eng/art-tools/pull/1231/files#diff-5f83f9cc1855d0e51814636f02bca1bba765fe0f6cc0818fbf5d93447454287eR99) line needs to change to account only for a number of failures > 1

locriandev:
This is still not fixing the issue. [This line](https://github.com/openshift-eng/art-tools/pull/1231/files#diff-5f83f9cc1855d0e51814636f02bca1bba765fe0f6cc0818fbf5d93447454287eR116) is executed when only latest attempt failed

doozer/doozerlib/cli/images_health.py
```diff
@@ -107,6 +107,9 @@ async def get_concerns(self, image_meta):
             elif latest_success_idx > 1:
                 msg += (f'Last {self.url_text(latest_success_bi_task_url, "success")} '
                         f'was {latest_success_idx} attempts ago')
+            elif latest_success_idx == 1:
+                # Do nothing
+                pass
```
locriandev, line 112:
```suggestion
                return  # skipping notifications when only latest attempt failed
```

locriandev (changes requested):

PR #1239, opened 2024-12-23
update scan sources to override dockerfile

ashwindasr:
Modify scan-sources to compute the digest with the overridden Dockerfile, from Konflux config, if it exists, so that we correct the digest correctly.


ashwindasr:
/hold

If this idea is good, we can create a new function in artcommonlib, to reduce code duplication. Since the new changes was copied over from Konflux Rebaser.

PR #1240, opened 2024-12-23
[konflux] override dockerfile after digest calculation

ashwindasr:
Move config digest calculation before we replace the Dockerfile


locriandev (approved):
/lgtm

PR #1262, opened 2025-01-10
KonfluxDB: Always use timezone aware datetime objects

vfreex:
- Move the datetime.replace(tzinfo=…) call from search_builds_by_fields to the scan_sources caller. This function is to force set the timezone for a given datetime object. If it is already timezone aware, it will override its timezone and cause an unexpected timezone change. We should only use it on timezone “naive” objects.
- Use datetime.astimezone() to convert all passed datetime objects into timezone aware objects. This makes sure all datetime calculations are timezone aware, and all datetime objects passed to BigQuery are also timezone aware.


locriandev (commented):
/lgtm

PR #1266, opened 2025-01-14
Define metadata.get_latest_build_async

locriandev:
Define an async version of `metadata.get_latest_build`. The final goal is to have a common interface for Metadata to get latest builds for Brew and Konflux.

The PR also updates `check_for_image_changes()` in scan-sources to use the new async function as a POC.
Test build: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4_scan/97330/


doozer/doozerlib/cli/scan_sources.py
```diff
@@ -292,16 +292,16 @@ def scan_for_upstream_changes(self, koji_api):
             else:
                 raise IOError(f'Unsupported meta type: {meta.meta_type}')
 
-    def check_for_image_changes(self, koji_api):
-        for image_meta in self.runtime.image_metas():
+    async def check_for_image_changes(self, koji_api):
+        async def _inner(image_meta):
```
vfreex, line 296: Invoking blocking function calls (e.g. find_oldest_newest, check_dependents, check_config_changes, etc) in a coroutine will effectively block the coroutine, so that you will not get a lot of performance benefits on `asyncio.gather`. Thinking it is better to wrap all those blocking calls in `asyncio.to_thread`.

locriandev: Yeah, but the goal here is not boosting performances. There's indeed a tiny advantage in `gather`ing those get_latest_build_async calls, but my intention here is to define the async function that will be used in later PRs.
If we want to make use of this new function to improve pipeline performances, we can do that in a follow up PR, as I meant to make this change as small as possible and merge it ASAP

vfreex: Sounds good to me. After you fix the flake error, I am ok with merging this PR

locriandev: Done, thanks

PR #1268, opened 2025-01-14
Define common entrypoint for metadata.get_latest_build

locriandev:
Test build: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4_scan/97351/console

PR #1267, opened 2025-01-14
Rename medatata.get_latest_build into get_latest_brew_build

locriandev:
`get_latest_build` will serve as a unique entry point to find builds in both Brew and Konflux, based on a new `build-system`  option that will be given to the runtime. Renaming `medatata.get_latest_build()` into `get_latest_brew_build()` will later let us define a `get_latest_konflux_build()` function and make `get_latest_build` a unique entrypoint for both, being controlled by the build-system option.

Test builds:
- [gen-assembly](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fgen-assembly/356/console)
- [build-sync](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fbuild-sync/13532/console)
- [ocp4](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4/13759/console)
- [ocp4-konflux](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-konflux/1945/console)
- [ocp4-scan-konflux](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-scan-konflux/7862/console)
- [ocp4-scan](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4_scan/97337/console)

Also tested elliott find-builds with `elliott --group=openshift-4.18 --assembly rc.4 find-builds --kind=image --non-payload`, which returned
```text
 cloud-event-proxy-container-v4.18.0-202412200437.p0.g7cfad7b.assembly.stream.el9
 cluster-nfd-operator-container-v4.18.0-202501071837.p0.gb8c54f7.assembly.stream.el9
 dpu-cni-container-v4.18.0-202412170221.p0.g99f3e30.assembly.stream.el9
 dpu-daemon-container-v4.18.0-202412170221.p0.g99f3e30.assembly.stream.el9
 dpu-operator-container-v4.18.0-202412170221.p0.g99f3e30.assembly.stream.el9
 ib-sriov-cni-container-v4.18.0-202412170221.p0.gceb851e.assembly.stream.el9
[...]

PR #1271, opened 2025-01-15
Define metadata.get_latest_konflux_build()

locriandev:
Follows:
- https://github.com/openshift-eng/art-tools/pull/1266
- https://github.com/openshift-eng/art-tools/pull/1267
- https://github.com/openshift-eng/art-tools/pull/1268

Implement the new `metadata.get_latest_konflux_build()` function and start using it in scan-sources-konflux.
Requires doozer to be called with the `--build-system=konflux` option.

Test build: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-scan-konflux/8122/console

### DIFF
--- a/artcommon/artcommonlib/konflux/konflux_db.py
+++ b/artcommon/artcommonlib/konflux/konflux_db.py
@@ -210,7 +210,7 @@ class KonfluxDb:
             name: str,
             group: str,
             outcome: KonfluxBuildOutcome = KonfluxBuildOutcome.SUCCESS,
-            assembly: str = None,
+            assembly: typing.Optional[str] = None,
             el_target: typing.Optional[str] = None,
             artifact_type: typing.Optional[ArtifactType] = None,
             engine: typing.Optional[Engine] = None,
--- a/artcommon/artcommonlib/konflux/konflux_db.py
+++ b/artcommon/artcommonlib/konflux/konflux_db.py
@@ -210,7 +210,7 @@ class KonfluxDb:
             name: str,
             group: str,
             outcome: KonfluxBuildOutcome = KonfluxBuildOutcome.SUCCESS,
-            assembly: str = 'stream',
+            assembly: str = None,
             el_target: typing.Optional[str] = None,
             artifact_type: typing.Optional[ArtifactType] = None,
             engine: typing.Optional[Engine] = None,
@@ -224,7 +224,7 @@ class KonfluxDb:
         :param name: component name, e.g. 'ironic'
         :param group: e.g. 'openshift-4.18'
         :param outcome: 'success' | 'failure'
-        :param assembly: non-standard assembly name, defaults to 'stream' if omitted
+        :param assembly: assembly name, if omitted any assembly is matched
         :param el_target: e.g. 'el8'
         :param artifact_type: 'rpm' | 'image'
         :param engine: 'brew' | 'konflux'
@@ -242,8 +242,10 @@ class KonfluxDb:
             Column('name', String) == name,
             Column('group', String) == group,
             Column('outcome', String) == str(outcome),
-            Column('assembly', String) == assembly,
         ]
+        if assembly:
+            base_clauses.append(Column('assembly', String) == assembly)
+
         order_by_clause = Column('start_time', quote=True).desc()
 
         if completed_before:
--- a/doozer/doozerlib/metadata.py
+++ b/doozer/doozerlib/metadata.py
@@ -21,6 +21,7 @@ import artcommonlib
 import doozerlib
 from artcommonlib import logutil, exectools, assertion
 from artcommonlib.assembly import assembly_basis_event, assembly_metadata_config
+from artcommonlib.konflux.konflux_build_record import KonfluxBuildRecord, KonfluxBuildOutcome, Engine
 from artcommonlib.pushd import Dir
 from artcommonlib.model import Missing, Model
 from doozerlib.brew import BuildStates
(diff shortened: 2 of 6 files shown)


artcommon/artcommonlib/konflux/konflux_db.py
```diff
@@ -224,7 +224,7 @@ async def get_latest_build(
         :param name: component name, e.g. 'ironic'
         :param group: e.g. 'openshift-4.18'
         :param outcome: 'success' | 'failure'
-        :param assembly: non-standard assembly name, defaults to 'stream' if omitted
+        :param assembly: assembly name, if omitted any assembly is matched
```
thegreyd, line 227: Hmm this behavior might do unexpected things, wondering why change this?

doozer/doozerlib/metadata.py
```diff
@@ -401,11 +402,117 @@ async def get_latest_build(self, **kwargs):
...
+        :param outcome: one in KonfluxBuildOutcome[FAILURE, SUCCESS, PENDING]
+        :param el_target: specify 'el7', 'el8', etc.
+                          For an RPM, leave as None if you want the true latest.
+                          For an image, leaving as none will default to the RHEL version in the branch
+                          associated with the metadata.
+        :param honor_is: If True, and an assembly component specifies 'is', that nvr will be returned.
+        :param completed_before: cut off timestamp for builds completion time
+        :param extra_patterns: e.g. {'release': 'b45ea65'} will result in adding "AND release LIKE '%b45ea65%'" to the query
+        """
+
```
thegreyd, line 461: I'm looking at the code for `get_latest_brew_build` 
https://github.com/openshift-eng/art-tools/blob/de3d8b791c0bf29f012f103e6d4671e0ff29d520/doozer/doozerlib/metadata.py#L412
We have params that are not used in this func, like component_name. Now that it's irrelevant, maybe we have a check that fails if the component name is different from what we determine

artcommon/artcommonlib/konflux/konflux_db.py
locriandev: That is required to mimic the behavior of [metadata.get_latest_build](https://github.com/openshift-eng/art-tools/blob/main/doozer/doozerlib/metadata.py#L417). Always defaulting to `stream` would make this impossible. I've also checked that we always pass in the assembly param when using the konflux_db function directly, so I don't see any risk in here.

doozer/doozerlib/metadata.py
locriandev: `component_name` is used by `get_latest_brew_build` to get the package ID, and this is later used to fetch the builds:

```
builds = koji_api.listBuilds(packageID=package_id,
```
Overriding it to match something different from the metadata component name could be used in particular cases (e.g. olm bundles?) although I haven't found any occurrences of  this. Either way, I feel this and other refactoring changes should be postponed as things are getting messy already and I want to limit changes to what's strictly needed for this task

artcommon/artcommonlib/konflux/konflux_db.py
```diff
@@ -210,7 +210,7 @@ async def get_latest_build(
             name: str,
             group: str,
             outcome: KonfluxBuildOutcome = KonfluxBuildOutcome.SUCCESS,
-            assembly: str = 'stream',
+            assembly: str = None,
```
vfreex, line 213:
```suggestion
            assembly: typing.Optional[str] = None,
```

`str` doesn't actually accept None value and will cause type checking error if you have.

doozer/doozerlib/metadata.py
```diff
@@ -401,11 +402,121 @@ async def get_latest_build(self, **kwargs):
             return await self.get_latest_brew_build_async(**kwargs)
 
         elif self.runtime.build_system == 'konflux':
-            raise NotImplementedError
+            return await self.get_latest_konflux_build(**kwargs)
```
vfreex, line 405: Considering this option is only to switch implementations for finding builds and ideally won't cause any user-facing differences, this might be better to be controlled by an env var. With that env var, we can start to having certain jobs to use the konflux implementation, then gradually make that default and remove the brew implementation without bothering removing this option.

But this is only an option suggestion and I can live with this if you have strong opinion.

doozer/doozerlib/metadata.py
```diff
@@ -401,11 +402,121 @@ async def get_latest_build(self, **kwargs):
...
+
+        is_config = self.config['is']
+
+        if self.meta_type == 'rpm':
+            if el_target is None:
+                raise ValueError(
+                    f'Expected el_target to be set when querying a pinned RPM component {self.distgit_key}')
+            is_nvr = is_config[f'el{el_target}']
+            if not is_nvr:
+                self.logger.warning('No pinned NVR found for RPM %s and target %s', self.rpm_name, el_target)
```
vfreex, line 424: For consistency with line 436, an error should be raised instead of returning none.

doozer/doozerlib/metadata.py
```diff
@@ -401,11 +402,121 @@ async def get_latest_build(self, **kwargs):
...
+            # Pinning also informs gen-payload when attempting to assemble a release.
+            is_nvr = is_config.nvr
+            if not is_nvr:
+                raise ValueError(f'Did not find nvr field in pinned Image component {self.distgit_key}')
+
+        # strict means raise an exception if not found.
+        return await self.runtime.konflux_db.get_build_record_by_nvr(is_nvr, strict=True)
+
+    async def get_latest_konflux_build(self, assembly: Optional[str] = None,
+                                       outcome: KonfluxBuildOutcome = KonfluxBuildOutcome.SUCCESS,
```
vfreex, line 442: Any use case to return a build with outcome other than SUCESS?

doozer/doozerlib/metadata.py
```diff
@@ -401,11 +402,121 @@ async def get_latest_build(self, **kwargs):
...
+        """
+
+        assert self.runtime.konflux_db is not None, 'Konflux DB must be initialized with GCP credentials'
+        self.runtime.konflux_db.bind(KonfluxBuildRecord)
+
+        # Is the component pinned in config?
+        if honor_is and self.config['is']:
+            if outcome != KonfluxBuildOutcome.SUCCESS:
+                # If this component is defined by 'is', history failures, etc, do not matter.
+                return None
```
vfreex, line 469: Here with a non-SUCCESS outcome (including None) and pinned by "is", this method always returns None. This seems to have inconsistent behaviors comparing to normal search.

doozer/doozerlib/metadata.py
locriandev: I don't either, also an env var sounds good to me. Would adding a `envar=BUILD_SYSTEM` [here](https://github.com/openshift-eng/art-tools/blob/main/doozer/doozerlib/cli/__init__.py#L107) work?

doozer/doozerlib/metadata.py
locriandev: I followed the logic we already had [here](https://github.com/openshift-eng/art-tools/blob/main/doozer/doozerlib/metadata.py#L552). It looks like for RPMs `get_latest_brew_build` isn't throwing an exception if a default != -1 is set (which is always the case, we pass in default=None everywhere but [here](https://github.com/openshift-eng/art-tools/blob/main/doozer/doozerlib/cli/release_gen_assembly.py#L477))

doozer/doozerlib/metadata.py
locriandev: As usual, I followed the prototype we already had in `get_latest_brew_build`. Looks like we never pass in something different from a successful build in there too, so I would either change both or None to keep them as similar as possible

doozer/doozerlib/metadata.py
locriandev: I've tried to follow the same pattern we already had [here](https://github.com/openshift-eng/art-tools/blob/main/doozer/doozerlib/metadata.py#L540-L543), do you see differences in behavior?

doozer/doozerlib/metadata.py
vfreex: Seems Doozer defines a lot of global env vars here: https://github.com/openshift-eng/art-tools/blob/27708d09b004e12ab590f95ca5e42271cadb7475/doozer/doozerlib/cli/cli_opts.py#L28

doozer/doozerlib/metadata.py
vfreex: If so, maybe always searching for SUCCESS build is better choice to keep the logic simpler and less confusing.

doozer/doozerlib/metadata.py
vfreex: default_return (https://github.com/openshift-eng/art-tools/blob/27708d09b004e12ab590f95ca5e42271cadb7475/doozer/doozerlib/metadata.py#L485) raises IOError when default == -1.

locriandev: yeah so for images we always pass in defautl = None so an exception is never thrown if a container build that was pinned isn't found. If you feel that should change, we should probably change it in legacy code as well

doozer/doozerlib/metadata.py
locriandev: Done with 5730f753935092b0ec307f3b5bb2c4a2d2af4eb2

doozer/doozerlib/metadata.py
locriandev: Done with f19fd0546a3df6a2ddacdcffb5bbfc8697de5ef3.

locriandev: @thegreyd can you please double check if this looks good to you? Although I don't see cases where we search for anything different from COMPLETED it might have slipped

locriandev: actually, scan-sources searches for failed builds [here](https://github.com/openshift-eng/art-tools/blob/main/doozer/doozerlib/metadata.py#L745), so it looks like that param must stay. and for the new `get_latest_konflux_build`, it will be needed once we decide to swap out `konflux_db.get_latest_build`

vfreex (commented):
/lgtm

PR #1274, opened 2025-01-16
promote-assembly: retry on 502

vfreex:


vfreex:
Updated to be less specific about the error.
/hold cancel

pyartcd/tests/pipelines/test_promote.py
```diff
@@ -1046,13 +1046,11 @@ def test_build_create_symlink(self, _):
             dry_run=False
         )
         pipeline = PromotePipeline(runtime, group="openshift-4.10", assembly="4.10.99", signing_env="prod")
-        temp_dir = tempfile.mkdtemp()
-        os.chdir(temp_dir)
```
vfreex, line 1050: `os.chdir` causes issues to other unit test cases. It is not directly related to this PR, but I would like to get it fixed.

PR #1277, opened 2025-01-16
[konflux] do not use the -container suffix in the NVRs

locriandev:
The -container is associated with the Brew package which is not relevant in the Konflux world. Removing this suffix would also let us distinguish the build system that produced a certain NVR.


joepvd:
/approve
but do not oversee potential ramifications. @ashwindasr, could you review?

vfreex (approved):
/lgtm

PR #1278, opened 2025-01-16
[ART-11306] Make quay-doomsday-backup async and send its results to Slack

adobes1:
Nearly identical changes as in https://github.com/openshift-eng/art-tools/pull/1170 (which has been reverted).
The only difference is that now the code does not query all RH Slack channels to get the channel ID. It simply read the ID from the Slack API response.

- [Example](https://redhat-internal.slack.com/archives/C05HB1F1THP/p1737042238643629) of a message after a successful run
- [Example](https://redhat-internal.slack.com/archives/C05HB1F1THP/p1737042297567809) of a message after a failed run

### DIFF
--- a/pyartcd/pyartcd/pipelines/quay_doomsday_backup.py
+++ b/pyartcd/pyartcd/pipelines/quay_doomsday_backup.py
@@ -1,12 +1,13 @@
 import os
 from typing import Optional
 import click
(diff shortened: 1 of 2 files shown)


joepvd (approved):
/lgtm

PR #1280, opened 2025-01-16
[konflux] network modes

ashwindasr:
Support two new network modes `hermetic` and `internal-only`. The latter is set by using proxies to allow traffic only to RH internal network, while the former is configured by setting a value `hermetic` to `true` for Konflux PipelineRun


ashwindasr:
Fixed merge conflict, reapplying lgtm

PR #1281, opened 2025-01-17
[Konflux] Support registry auth for bundle build

vfreex:


doozer/doozerlib/util.py
```diff
@@ -561,27 +563,101 @@ def oc_image_info__caching(pull_spec: str, go_arch: str = 'amd64') -> Dict:
...
-    cmd = ['oc', 'image', 'info', f'--filter-by-os={go_arch}', '-o', 'json', pull_spec]
+    options = [f'--filter-by-os={go_arch}', '-o', 'json']
+    if registry_config:
+        options.extend([f'--registry-config={registry_config}'])
+    cmd = ['oc', 'image', 'info'] + options + [pull_spec]
     _, out, _ = await exectools.cmd_gather_async(cmd)
     return json.loads(out)
 
 
+async def oc_image_info_with_auth_async(
```
locriandev, line 593: it looks like we could just define a single `oc_image_info` function that accepts optional auth params instead of having `oc_image_info_with_auth_async` fall back to `oc_image_info_async` in case username and password are not provided?

vfreex: That is possible. I was just lazy to do that. Let me check how easy or hard to change that.

vfreex: Updated

locriandev (commented):
/lgtm

PR #1283, opened 2025-01-17
[ART-11310] add protential key errors in response data

Ximinhan:
handle the protential errors when request pp


vfreex:
I don't understand the purpose of this PR. Catching an exception and wrap it into another exception doesn't show any meaningful benefit and is actually an no-op to me.

Ximinhan:
> I don't understand the purpose of this PR. Catching an exception and wrap it into another exception doesn't show any meaningful benefit and is actually an no-op to me.

I plan to catch the potential key not found error in response data

PR #1286, opened 2025-01-20
[konflux] ENV variables for cachito/cachi2 and network mode identifcation

ashwindasr:
Per [doc](https://docs.google.com/document/d/1Ajc6MSNJz20b34L7QNwoPcWKxId7Fy-eRt7oqyaJl8c/edit?tab=t.0#heading=h.4fbec24ysdh3)

Inserting ENV variables so that upstream can indentify if the build environment is brew/konflux and if the dependency manager that's available is cachito or cachi2


ashwindasr:
TODO:

- [ ] Create a subdirectory for cachito-emulation

PR #1292, opened 2025-01-21
Revert "Revert "[konflux] network modes""

ashwindasr:
Reverts openshift-eng/art-tools#1288

Includes fixes: https://github.com/openshift-eng/art-tools/pull/1292/commits/1901a5fb3af4afc5456fbcba40549cace36a6675


ashwindasr:
Tested: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-konflux/2433/console

doozer/doozerlib/backend/konflux_client.py
```diff
@@ -359,6 +360,9 @@ def _modify_param(params: List, name: str, value: Union[str, bool, list[str]]):
         _modify_param(params, "image-expires-after", "6w")
         _modify_param(params, "build-platforms", list(build_platforms))
 
+        if image_metadata.config.get("konflux", {}).get("network-mode") == "hermetic":
```
locriandev, line 363: looks like the hermetic build condition is the only thing we need from the Metadata object. Consider passing this around instead of the whole Metadata.

locriandev: either way,
/lgtm

ashwindasr: Have a couple of follow up PRs that might make use of metadata, so hoping to keep it in for now.

Thanks for the review

PR #1294, opened 2025-01-22
Prevent form-urlencoded/parm interpretation of + in URI

jupierce:
Given a uri like 'https://.../x86_64/dependencies/rpms/4.19-el9-beta/perl-Text-Tabs+Wrap-2013.0523-460.el9.noarch.rpm' unquote_plus is interpretting and replacing the '+' as a space (' '). This encoding rule is only valid in a form submission (which the CloudFront request is not) and query string parameters (which we are not parsing here).
Applied here, the encoding rule mangles the filename which does include a literal +. unquote (vs unquote_plus) does not replace + with ' '.


jlebon (approved):
Looks sane to me!

PR #1296, opened 2025-01-23
[konflux] persist images on quay repo

ashwindasr:
Images will not be purged, if we do not specify images-expires-after


joepvd (approved):
/lgtm

PR #1299, opened 2025-01-24
Update build-sync title before exiting

thegreyd:
For non-stream assemblies we were raising before updating title of the run.
example: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fbuild-sync/14035/
So fix that.


locriandev (approved):
/lgtm

PR #1303, opened 2025-01-27
Make oc_image_info option generic

locriandev:
Currently, we run `oc image info` with --filter-by-os always on. This PR suggests making the `oc_image_info()` function generic, and create new specific functions to be used with the `--filter-by-os` and `--show-multiarch` options instead.

### DIFF
--- a/doozer/doozerlib/util.py
+++ b/doozer/doozerlib/util.py
@@ -489,8 +489,8 @@ def get_release_calc_previous(version, arch,
     return sort_semver(list(upgrade_from))
 
 
-async def find_manifest_list_sha(pull_spec):
-    image_data = oc_image_info__caching(pull_spec)
+async def find_manifest_list_sha(pullspec):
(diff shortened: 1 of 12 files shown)


doozer/doozerlib/util.py
```diff
@@ -540,84 +540,135 @@ def get_release_name_for_assembly(group_name: str, releases_config: Model, assem
...
+    Will return a single dictionary repesenting the amd64 arch
+    """
+    return await oc_image_info_async(pull_spec, f'--filter-by-os={go_arch}',
+                                     registry_config=registry_config,
+                                     registry_username=registry_username,
+                                     registry_password=registry_password)
+
+
+async def oc_image_info_show_multiarch_async(
+        pull_spec: str,
```
thegreyd, line 654: nitpik but if I remember we've been using `pullspec` as one word in most places

locriandev: yes, in fact I also renamed some of our old occurrences. let me be more precise with this

PR #1304, opened 2025-01-27
[konflux] Build only for x86 for now

ashwindasr:
In the Konflux call, it was discussed that we build only x86, for now, for rapid development, until a single arch payload is generated.


locriandev (approved):
/lgtm

PR #1308, opened 2025-01-28
Give oc_image_info() auth powers

locriandev:
Give the sync version of oc_image_info the same auth capabilities that the async version already has.

### DIFF
--- a/doozer/doozerlib/util.py
+++ b/doozer/doozerlib/util.py
@@ -540,7 +540,13 @@ def get_release_name_for_assembly(group_name: str, releases_config: Model, assem
     return get_release_name(assembly_type, group_name, assembly_name, patch_version)
 
 
-def oc_image_info(pullspec: str, *options):
+def oc_image_info(
+        pullspec: str,
+        *options,
+        registry_config: Optional[str] = None,
+        registry_username: Optional[str] = None,
+        registry_password: Optional[str] = None
+) -> Union[Dict, List]:
     """
     Returns a Dict of the parsed JSON output of `oc image info` for the specified
     pullspec. Use oc_image_info__caching if you do not believe the image will change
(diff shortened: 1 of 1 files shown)


doozer/doozerlib/util.py
```diff
@@ -540,20 +540,57 @@ def get_release_name_for_assembly(group_name: str, releases_config: Model, assem
...
+        cmd = ['oc', 'image', 'info', '-o', 'json', pullspec]
+        cmd.extend(options)
+        if auth_file:
+            cmd.extend([f'--registry-config={auth_file}'])
+        out, _ = exectools.cmd_assert(cmd, retries=3)
+        return json.loads(out)
+
+    # If neither a username nor a password were provided, assume an auth file is being used or no auth at all
+    if not registry_username and not registry_password:
+        return run_oc(auth_file=registry_config)
```
thegreyd, line 572: here registry_config could be `None` ?

doozer/doozerlib/util.py
```diff
@@ -540,20 +540,57 @@ def get_release_name_for_assembly(group_name: str, releases_config: Model, assem
...
+
+    # Otherwise, a password is always needed for basic auth
+    assert registry_password, 'registry_password must be provided if auth is needed.'
```
thegreyd, line 575: So we need either auth_file or password to use this func. I would say it would  be fine to run if none is present, as we were doing before?

doozer/doozerlib/util.py
locriandev: yes it could. and if it is, the `--registry-login` option is omitted. same logic as https://github.com/openshift-eng/art-tools/blob/main/doozer/doozerlib/util.py#L630

doozer/doozerlib/util.py
locriandev: if you don't provide anything, oc runs without auth options. If you provide a `registry_config`, it must point to a file where auth credentials are stored. Otherwise, you must provide username and password to be used with basic auth (which will be used to create a temporary registry config file)

doozer/doozerlib/util.py
thegreyd: ah sorry i missed that, thanks for clarifying it

locriandev (commented):
/approve

PR #1309, opened 2025-01-28
[find-bugs:golang] Validate cve_id is set or skip

thegreyd:
Make sure bugs have a valid cve_id otherwise the command crashes later.


locriandev (commented):
/lgtm

PR #1310, opened 2025-01-28
Revert "Start RHCOS arch jobs earlier"

dustymabe:
This reverts commit 45088161971df1aace0527b32b7d3a6bd0e2d74e.

As mentioned in [1] this can have a few side effects that I don't really think we want to deal with on a continuing basis. Let's just revert back to letting the x86_64 buils pass tests and upload to clouds first (i.e. error prone operations) first.

[1] https://github.com/openshift-eng/art-tools/pull/1164#issuecomment-2620026207


joepvd (approved):
/lgtm

PR #1318, opened 2025-01-30
Retry failed arch only

thegreyd:
Currently `@retry` applies for the func which runs `oc adm release new` for all arches.
Instead wrap it on the command itself, so only the failed arch invocation is retried.

### DIFF
--- a/doozer/doozerlib/cli/release_gen_payload.py
+++ b/doozer/doozerlib/cli/release_gen_payload.py
@@ -1287,7 +1287,6 @@ class GenPayloadCli:
         sha = await find_manifest_list_sha(output_pullspec)
         return exchange_pullspec_tag_for_shasum(output_pullspec, sha)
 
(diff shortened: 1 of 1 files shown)


doozer/doozerlib/cli/release_gen_payload.py
```diff
@@ -1305,6 +1304,19 @@ async def create_multi_release_image(self, imagestream_name: str, multi_release_
         async with aiofiles.open(multi_release_is_path, mode="w+") as mf:
             await mf.write(yaml.safe_dump(multi_release_is))
 
+        @retry(reraise=True, stop=stop_after_attempt(3), wait=wait_fixed(60))
+        async def _run(to_image, to_image_base):
+            await exectools.cmd_assert_async([
```
thegreyd, line 1309:
```suggestion
            return await exectools.cmd_assert_async([
```

PR #1323, opened 2025-02-04
Ignore invalid version-release rpms in rhcos

thegreyd:
context: https://redhat-internal.slack.com/archives/CB95J6R4N/p1738691928076439
and https://github.com/openshift/os/pull/1737


sosiouxme:
seems legit
/lgtm
(what, no test case added? :-P )

PR #1334, opened 2025-02-10
Fix update streams.yml func

thegreyd:


pyartcd/pyartcd/pipelines/update_golang.py
```diff
@@ -355,7 +355,8 @@ async def update_golang_streams(self, go_version, builder_nvrs):
             for branch in fork_repo.get_branches():
```
Ximinhan, line 355: error "Branch not found" in https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fgolang-builder/101/console is due to the `branch` var is already defined above as upstream f"openshift-{self.ocp_version}" so this part could be
```
for branch_item in fork_repo.get_branches():
      if branch_item.name == branch_name:
          fork_repo.get_git_ref(f"heads/{branch_name}").delete()
```

PR #1339, opened 2025-02-13
[konflux] build/sync microshift-bootc location

ashwindasr:
ref [thread](https://redhat-internal.slack.com/archives/GDBRP5YJH/p1739373251573739?thread_ts=1739372936.015079&cid=GDBRP5YJH)

Konflux needs images to be built to the default build repo in order for it to be released. But the microshift team has an additional requirement that it has to be synced to `quay.io/openshift-release-dev/ocp-v4.0-art-dev` as well

### DIFF
--- a/artcommon/artcommonlib/konflux/konflux_db.py
+++ b/artcommon/artcommonlib/konflux/konflux_db.py
@@ -13,7 +13,7 @@ from sqlalchemy import Column, String, DateTime, func
 
 from artcommonlib import bigquery
 from artcommonlib.konflux import konflux_build_record
-from artcommonlib.konflux.konflux_build_record import KonfluxBuildOutcome, ArtifactType, KonfluxRecord, Engine, KonfluxBuildRecord
+from artcommonlib.konflux.konflux_build_record import KonfluxBuildOutcome, ArtifactType, KonfluxRecord, Engine
 
 SCHEMA_LEVEL = 1
 DEFAULT_SEARCH_WINDOW = 90
@@ -216,7 +216,7 @@ class KonfluxDb:
             completed_before: typing.Optional[datetime] = None,
             extra_patterns: dict = {},
             strict: bool = False,
-    ) -> typing.Optional[KonfluxBuildRecord]:
+    ) -> typing.Optional[KonfluxRecord]:
         """
         Search for the latest Konflux build information in BigQuery.
 
--- a/artcommon/artcommonlib/konflux/konflux_db.py
+++ b/artcommon/artcommonlib/konflux/konflux_db.py
@@ -13,7 +13,7 @@ from sqlalchemy import Column, String, DateTime, func
 
 from artcommonlib import bigquery
 from artcommonlib.konflux import konflux_build_record
-from artcommonlib.konflux.konflux_build_record import KonfluxBuildOutcome, ArtifactType, KonfluxRecord, Engine
+from artcommonlib.konflux.konflux_build_record import KonfluxBuildOutcome, ArtifactType, KonfluxRecord, Engine, KonfluxBuildRecord
 
 SCHEMA_LEVEL = 1
 DEFAULT_SEARCH_WINDOW = 90
@@ -216,7 +216,7 @@ class KonfluxDb:
             completed_before: typing.Optional[datetime] = None,
             extra_patterns: dict = {},
             strict: bool = False,
-    ) -> typing.Optional[KonfluxRecord]:
+    ) -> typing.Optional[KonfluxBuildRecord]:
         """
         Search for the latest Konflux build information in BigQuery.
 
--- a/pyartcd/pyartcd/pipelines/build_microshift_bootc.py
+++ b/pyartcd/pyartcd/pipelines/build_microshift_bootc.py
@@ -24,7 +24,7 @@ from pyartcd.util import (get_assembly_type,
                           get_release_name_for_assembly,
                           get_microshift_builds)
 from pyartcd.plashets import build_plashets, plashet_config_for_major_minor
-from doozerlib.constants import ART_PROD_IMAGE_REPO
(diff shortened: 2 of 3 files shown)


ashwindasr:
/hold

Until a test is run

artcommon/artcommonlib/konflux/konflux_db.py
```diff
@@ -216,7 +216,7 @@ async def get_latest_build(
             completed_before: typing.Optional[datetime] = None,
             extra_patterns: dict = {},
             strict: bool = False,
-    ) -> typing.Optional[KonfluxRecord]:
+    ) -> typing.Optional[KonfluxBuildRecord]:
```
ashwindasr, line 219: I think this was supposed to be `KonfluxBuildRecord` since we are retrieving the build from the table, hence made the change, so that I can access the embargo status here: https://github.com/openshift-eng/art-tools/pull/1339/files#diff-bc471e2aee7ba5ac9bc27eca1f0ddc68963a7045561de4be6a3c8a77e0362364R102

vfreex: This method is also used to query `KonfluxBundleBuildRecord`. That is the reason why we use the parent class `KonfluxRecord` here.

vfreex (approved):
/lgtm

pyartcd/pyartcd/pipelines/build_microshift_bootc.py
```diff
@@ -98,6 +98,26 @@ async def run(self):
         for arch, digest in digest_by_arch.items():
             await self.sync_to_mirror(arch, bootc_build.el_target, f"{repo_url}@{digest}")
 
+        if not self.runtime.dry_run:
+            if bootc_build.embargoed:
+                await self.sync_to_quay(bootc_build.image_pullspec, ART_PROD_PRIV_IMAGE_REPO)
+            else:
+                await self.sync_to_quay(bootc_build.image_pullspec, ART_PROD_IMAGE_REPO)
+        else:
+            self._logger("Skipping sync to quay.io/openshift-release-dev/ocp-v4.0-art-dev since in dry-run mode")
```
thegreyd, line 107: Imo this can move to the `self.sync_to_quay` func

pyartcd/pyartcd/pipelines/build_microshift_bootc.py
```diff
@@ -98,6 +98,26 @@ async def run(self):
...
+
+    async def sync_to_quay(self, source_pullspec, destination_repo):
+        """
+        Microshift team needs the image on quay.io/openshift-release-dev/ocp-v4.0-art-dev as well.
+        """
+        # eg: extract 'ose-baremetal-installer-rhel9-v4.19.0-20250210.224515' from
+        # quay.io/redhat-user-workloads/ocp-art-tenant/art-images:ose-baremetal-installer-rhel9-v4.19.0-20250210.224515
```
thegreyd, line 114: Comments don't look particularly relevant here, would remove them

pyartcd/pyartcd/pipelines/build_microshift_bootc.py
```diff
@@ -98,6 +98,26 @@ async def run(self):
...
 
+        if not self.runtime.dry_run:
+            if bootc_build.embargoed:
+                await self.sync_to_quay(bootc_build.image_pullspec, ART_PROD_PRIV_IMAGE_REPO)
+            else:
+                await self.sync_to_quay(bootc_build.image_pullspec, ART_PROD_IMAGE_REPO)
+        else:
+            self._logger("Skipping sync to quay.io/openshift-release-dev/ocp-v4.0-art-dev since in dry-run mode")
+
+    async def sync_to_quay(self, source_pullspec, destination_repo):
```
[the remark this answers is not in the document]
thegreyd: we can run `oc mirror` with --dry-run

pyartcd/pyartcd/pipelines/build_microshift_bootc.py
ashwindasr: True - but I was wondering if we need to enter the function at all, sync build won't exist in dry-run mode

pyartcd/pyartcd/pipelines/build_microshift_bootc.py
ashwindasr: true - but during debugging, I had found comments like these helpful, particularly, if you don't know off the top of your head, what the value is for `source_pullspec` for example. So was hoping to keep it in for clarity, even though it might be kinda overexplaining

pyartcd/pyartcd/pipelines/build_microshift_bootc.py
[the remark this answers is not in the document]
thegreyd: Yes good point! Then a log message should suffice

pyartcd/pyartcd/pipelines/build_microshift_bootc.py
thegreyd: Hmm right, things would also break early with dry-run looking above this piece of code. Anything that depends on the build would fail. Okay this should be fine

pyartcd/pyartcd/pipelines/build_microshift_bootc.py
thegreyd: I would say a couple of things make this a bit confusing, this imo would be cleaner
```
# extract tag 
# e.g.'ose-baremetal-installer-rhel9-v4.19.0-20250210.224515' from
# quay.io/redhat-user-workloads/ocp-art-tenant/art-images:ose-baremetal-installer-rhel9-v4.19.0-20250210.224515
tag = source_pullspec.split(':')[-1]
destination_pullspec = f"{destination_repo}:{tag}"
```

ashwindasr: thanks! updated

PR #1340, opened 2025-02-13
[ART-11852] Create microshift advisory if not exist

Ximinhan:
If the release didn't create an advisory for micro shift, then when ,building micro shift, it should create one and update the config in pin micro shift build pr.

### DIFF
--- a/pyartcd/pyartcd/pipelines/build_microshift.py
+++ b/pyartcd/pyartcd/pipelines/build_microshift.py
@@ -3,14 +3,16 @@ import logging
 import os
 import re
 import traceback
+import subprocess
 import click
+from subprocess import PIPE
 from collections import namedtuple
 from pathlib import Path
 from typing import Dict, Iterable, List, Optional, Tuple
 
 from artcommonlib.arch_util import brew_arch_for_go_arch
 from artcommonlib.assembly import AssemblyTypes
-from artcommonlib.util import get_ocp_version_from_group, new_roundtrip_yaml_handler
+from artcommonlib.util import get_ocp_version_from_group, new_roundtrip_yaml_handler, get_assembly_release_date
 from artcommonlib import exectools
 from doozerlib.util import isolate_nightly_name_components
 from ghapi.all import GhApi
@@ -24,6 +26,7 @@ from pyartcd.record import parse_record_log
 from pyartcd.runtime import Runtime
 from pyartcd.util import (get_assembly_type,
                           isolate_el_version_in_release,
+                          get_release_name_for_assembly
                           load_group_config,
                           load_releases_config,
                           default_release_suffix,
@@ -49,6 +52,7 @@ class BuildMicroShiftPipeline:
         self._logger = logger or runtime.logger
         self._working_dir = self.runtime.working_dir.absolute()
         self.releases_config = None
+        self.advisory_num = None
         self.slack_client = slack_client
 
         # determines OCP version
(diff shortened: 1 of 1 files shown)


pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -84,15 +90,67 @@ async def run(self):
         if self.assembly_type is AssemblyTypes.STREAM:
             await self._rebase_and_build_for_stream()
         else:
+            # Check if microshift advisory is defined in assembly
+            if 'microshift' not in advisories:
+                self.advisory_num = await self.create_microshift_advisory()
```
thegreyd, line 95: So as I understand this will create the advisory always when this job is run? I was hoping we had a separate flag to do this

pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -2,28 +2,33 @@
...
 from pathlib import Path
 from typing import Dict, Iterable, List, Optional, Tuple
 
 from artcommonlib.arch_util import brew_arch_for_go_arch
 from artcommonlib.assembly import AssemblyTypes
-from artcommonlib.util import get_ocp_version_from_group, new_roundtrip_yaml_handler
+from artcommonlib.util import get_ocp_version_from_group, new_roundtrip_yaml_handler, get_assembly_release_date
 from artcommonlib import exectools
 from doozerlib.util import isolate_nightly_name_components
-from ghapi.all import GhApi
```
thegreyd, line 16: Why replace ghapi?

pyartcd/pyartcd/pipelines/build_microshift.py
Ximinhan: if this run for a named release and microshfit is not created then it will create it, I think flag don't have much value

pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -294,21 +352,16 @@ async def _verify_microshift_bugs(self, advisory_id):
...
+    async def _change_advisory_status(self, advisory_id):
+        # move advisory status to QE
+        e = Erratum(errata_id=advisory_id)
+        if e.errata_state == "QE":
+            self._logger.info(f"Errata {advisory_id} status already in QE")
+        else:
+            e.setState("QE")
+            e.commit()
+            self._logger.info(f"Changed {advisory_id} to QE")
+            push_cdn_stage(advisory_id)
```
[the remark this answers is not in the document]
Ximinhan: one reason is this need push_cdn_stage, another reason is to avoid cmd invoking to do the same thing

pyartcd/pyartcd/pipelines/build_microshift.py
Ximinhan: I refactor that create pr function to reduce git clone and use a more concise way to create pr

pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -84,15 +87,47 @@ async def run(self):
         if self.assembly_type is AssemblyTypes.STREAM:
             await self._rebase_and_build_for_stream()
         else:
+            # Check if microshift advisory is defined in assembly
+            if 'microshift' not in advisories:
```
vfreex, line 91: Previously we use `microshift: -1` to signal `prepare-release` job to create microshift advisory. Thinking we should still respect `-1` in this job. 
```suggestion
            if advisories.get(''microshift') <= 0:
```

Ximinhan: updated to check both

vfreex (approved):
/lgtm

PR #1346, opened 2025-02-14
[konflux] explicitly pass cyclonedx

ashwindasr:
To follow this change: https://github.com/openshift-priv/art-konflux-template/pull/46

We have the option to pass it as a param: https://github.com/konflux-ci/build-definitions/blob/main/task/buildah-remote-oci-ta/0.4/buildah-remote-oci-ta.yaml#L88


ashwindasr:
/hold

Until test is green

ashwindasr:
/unhold

Test build green

PR #1347, opened 2025-02-14
[konflux] fix bootc

ashwindasr:
To provide pull access to konflux build repo

Same logic as build sync: https://github.com/openshift-eng/art-tools/blob/29f04b71d7e8e1b5a2cbfd4901840de82917059e/doozer/doozerlib/cli/release_gen_payload.py#L883-L887


ashwindasr:
https://github.com/openshift-eng/art-tools/pull/1346 need to merge first

PR #1353, opened 2025-02-17
Fail build-sync for konflux in case of missing builds

locriandev:
We had introduced a temporary permit for missing builds when syncing Konflux images. This was needed because not all components had available builds, as most of them had been garbage-collected and been failing to rebuild since then.

Now this should be fixed at least for 4.19, so re-introducing the constraint.


joepvd (approved):
/lgtm

PR #1357, opened 2025-02-18
microshift-buildc: bootc-pullspec.txt fixes

vfreex:
The pullspecs should refer to the public repo. In case of embargo, we don't sync those files at all.

Also the local bootc-pullspec.txt files will be created in TemporaryDirectory to avoid potential race condition.

Tested with https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fbuild-microshift-bootc/57/


joepvd (approved):
/lgtm

PR #1361, opened 2025-02-18
Pylint fixes

thegreyd:
Changes
- Introduce pylintrc for the art-tools repo and make task
- Resolve (some) pylint errors


doozer/doozerlib/runtime.py
```diff
@@ -107,6 +107,24 @@ def __init__(self, **kwargs):
         self.disable_gssapi = False
         self._build_data_product_cache: Model = None
 
+        # init cli options
+        self.group = None
```
locriandev, line 111: as we are trying to remove duplications, wouldn't it make sense to move some of these fields into artcommonlib.runtime? e.g. group, images, rpms, are going to be used by elliott as well.

Makefile
```diff
@@ -9,6 +9,9 @@ venv:
 lint:
 	./.venv/bin/python -m flake8
 
+pylint:
```
locriandev, line 12: are both linters necessary? I am curious to see if flake8 can catch syntax "errors" that pylint cannot

doozer/doozerlib/build_info.py
```diff
@@ -445,7 +445,7 @@ def get_rhel_base_version(self) -> Optional[int]:
             if el_ver:
                 return el_ver
 
-        raise None
+        return None
```
locriandev, line 448: This is a behavioral change, as we would be raising an error (`raise None` results in a TypeError exception) instead of returning None, which should be handled by the caller. Unless we are handling this, I'd rather `raise ValueError` or the like.

Makefile
thegreyd: pylint is more rigorous than flake8, it runs slower but includes a lot more checks

thegreyd: Ah so you mean what's the advantage of keeping flake8 now? I would say flake8 is great for quick check as I mentioned, as pylint runs slower. pylint also shows more errors in the repo. This PR addresses some of them but until we resolve all of them we cannot make pylint a blocking CI check. So it is good to keep flake8 until then atleast.

doozer/doozerlib/build_info.py
locriandev: /lgtm
yet
/hold
until we figure this out

thegreyd: raising ValueError now

PR #1379, opened 2025-02-26
ART-12161 Handle brew/konflux sync failures separately

locriandev:
- create separate fail counters for Brew and Konflux
- explicitly state which build system failed to assemble a payload in pings
- do not ping public channels for Konflux (yet)

Ref. [ART-12161](https://issues.redhat.com/browse/ART-12161)


joepvd (approved):
/lgtm

PR #1384, opened 2025-02-27
Add configurable timeouts to pipelineruns

jupierce:
There is a master 12h spec.timeout set in the pipelinerun by doozer, but this is far more generous that we need in practice.  Allow each image metadata to specify a timeout in minutes like `konflux.build_timeout: 60` or default to 5 hours (enforced by doozer itself).  Doozer will also cancel the pipelinerun if it detects that a pod has been pending for over an hour.

### DIFF
--- a/doozer/doozerlib/backend/konflux_client.py
+++ b/doozer/doozerlib/backend/konflux_client.py
@@ -16,41 +16,9 @@ from kubernetes.dynamic import DynamicClient, exceptions, resource
 from ruamel.yaml import YAML
 
 from doozerlib import constants
-from doozerlib.image import ImageMetadata
 
 yaml = YAML(typ="safe")
 LOGGER = logging.getLogger(__name__)
-
-
-class KubeCondition:
-    def __init__(self, condition_obj: Dict):
-        self.type = condition_obj.get('type')
-        self.message = condition_obj.get('message')
-        self.reason = condition_obj.get('reason')
-        self.status = condition_obj.get('status')
-        self.last_transition_time = None
-        if condition_obj.get('lastTransitionTime'):
-            self.last_transition_time = datetime.datetime.fromisoformat(condition_obj.get('lastTransitionTime').rstrip("Z"))
-
-    def is_status_true(self) -> bool:
-        return str(self.status).lower() == 'true'
-
-    def is_status_false(self) -> bool:
-        return str(self.status).lower() == 'false'
-
-    @staticmethod
-    def find_condition(obj, condition_type: str, _default: Optional["KubeCondition"] = None) -> "KubeCondition":
-        """
-        Searches a kube object's status.conditions for a specified condition type. Returns the
-        condition entry if found. Otherwise, returns _default value.
-        """
-        try:
-            for condition in obj.status.get('conditions', []):
-                if condition['type'] == condition_type:
-                    return KubeCondition(condition)
-        except AttributeError:
-            pass
-        return _default
 
 
 class KonfluxClient:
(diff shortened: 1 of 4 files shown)


vfreex (commented):
Generally LGTM. One nit about type annotation for `wait_for_pipelinerun`.
@jupierce Once you make flake8 happy, i'll LGTM.

doozer/doozerlib/backend/konflux_client.py
```diff
@@ -524,14 +492,18 @@ def build_pipeline_url(pipelinerun):
...
                 f"pipelineruns/{pipelinerun_name}")
 
-    async def wait_for_pipelinerun(self, pipelinerun_name: str, namespace: Optional[str] = None) -> (resource.ResourceInstance, List[Dict]):
+    async def wait_for_pipelinerun(self, pipelinerun_name: str, namespace: Optional[str] = None, timeout_timedelta: Optional[datetime.timedelta] = None) -> (resource.ResourceInstance, List[Dict]):
         """
         Wait for a PipelineRun to complete.
 
         :param pipelinerun_name: The name of the PipelineRun.
         :param namespace: The namespace of the PipelineRun.
+        :param timeout_timedelta: Maximum time to wait for pipeline to run before canceling it (defaults to 1 hour)
```
ashwindasr, line 501: Can't seem to find where the default 1hr is configured?

ashwindasr (changes requested):

doozer/doozerlib/backend/konflux_client.py
```diff
@@ -524,14 +492,18 @@ def build_pipeline_url(pipelinerun):
                 f"applications/{application}/"
                 f"pipelineruns/{pipelinerun_name}")
 
-    async def wait_for_pipelinerun(self, pipelinerun_name: str, namespace: Optional[str] = None) -> (resource.ResourceInstance, List[Dict]):
+    async def wait_for_pipelinerun(self, pipelinerun_name: str, namespace: Optional[str] = None, timeout_timedelta: Optional[datetime.timedelta] = None) -> (resource.ResourceInstance, List[Dict]):
```
ashwindasr, line 495:
```suggestion
    async def wait_for_pipelinerun(self, pipelinerun_name: str, namespace: Optional[str] = None, timeout_timedelta: Optional[datetime.timedelta] = datetime.timedelta(hours=1)) -> (resource.ResourceInstance, List[Dict]):
```

doozer/doozerlib/backend/konflux_client.py
```diff
@@ -524,14 +492,18 @@ def build_pipeline_url(pipelinerun):
...
         """
         Wait for a PipelineRun to complete.
 
         :param pipelinerun_name: The name of the PipelineRun.
         :param namespace: The namespace of the PipelineRun.
+        :param timeout_timedelta: Maximum time to wait for pipeline to run before canceling it (defaults to 1 hour)
         :return: The PipelineRun ResourceInstance and a List[Dict] with a copy of an associated Pod.
         """
+        if timeout_timedelta is None:
+            timeout_timedelta = datetime.timedelta(hours=5)
```
ashwindasr, line 505: This can possibly be removed if we are setting the default value in the function definition?

doozer/doozerlib/backend/konflux_client.py
```diff
@@ -606,6 +583,11 @@ def _inner():
                                 else:
                                     creation_time = current_time
                                 age = current_time - creation_time
+
+                                if pod_phase == 'Pending' and age > datetime.timedelta(hours=1):
```
ashwindasr, line 587: Ahh I see that its here. Maybe we should move it to someplace more visible?

jupierce: Promoted to a method parameter to make it easier to trace and adjust in the future.

doozer/doozerlib/backend/konflux_client.py
ashwindasr: Well I assumed that programmers by default will look at the function definitions to understand what are the accepted values first, before they pass in the values. Since if we want to prevent errors like this, we should be adopting even more rigorous checks right

PR #1389, opened 2025-02-27
[konflux] remove unused code

ashwindasr:
SUPPORTED_ARCHES is defined here: https://github.com/openshift-eng/art-tools/blob/main/doozer/doozerlib/backend/konflux_client.py#L30, which seems to be the source of truth.


vfreex (approved):
/lgtm

PR #1394, opened 2025-02-28
artcommon: Update Konflux DB classes

vfreex:
- Adds bundle_package_name and bundle_csv_name fields to KonfluxBundleBuildRecord
- Adds KonfluxFbcBuildRecord for FBC build
- Fix type annotation for `add_build`.


locriandev (commented):
/lgtm
/hold

need to create the FBCS table first

PR #1399, opened 2025-03-03
Fix image component for Konflux builds

thegreyd:
Konflux image builds do not have `-container` suffix. So adjust for that

### DIFF
--- a/artcommon/artcommonlib/metadata.py
+++ b/artcommon/artcommonlib/metadata.py
@@ -9,7 +9,7 @@ from artcommonlib.brew import BuildStates
 from artcommonlib import logutil
 from artcommonlib.model import Missing, Model
 from artcommonlib.assembly import assembly_basis_event, assembly_metadata_config
-from artcommonlib.konflux.konflux_build_record import KonfluxBuildRecord, KonfluxBuildOutcome
+from artcommonlib.konflux.konflux_build_record import KonfluxBuildRecord, KonfluxBuildOutcome, Engine
 
 CONFIG_MODES = [
     'enabled',  # business as usual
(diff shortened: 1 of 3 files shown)


vfreex:
@thegreyd The component name should not be used in Konflux. This concept of component name is dedicated to Brew. Could you explain the issue you experienced and the reason why this change is needed?

thegreyd:
Here's where I am trying to work with brew and konflux builds in an abstract way: https://github.com/openshift-eng/art-tools/pull/1398/files#diff-905e83af7dae5d68460b5dafb78b47402f38e8ac9e9afaf18c0afa8abf180e82R46

I thought i could update and use this. But what you say makes sense, that in konflux world distgit_key will be the definitive name, right?

vfreex:
> Here's where I am trying to work with brew and konflux builds in an abstract way: https://github.com/openshift-eng/art-tools/pull/1398/files#diff-905e83af7dae5d68460b5dafb78b47402f38e8ac9e9afaf18c0afa8abf180e82R46 . So the usecase is to parse nvrs from cli and match them to the right metadata object.

Got it. That makes sense to me. But there is still one thing to concern: [The component name is set in `com.redhat.component` label](https://github.com/openshift/art-tools/blob/6f2ad7415184d70074f3c4b39d1f8f5e9697a983/doozer/doozerlib/backend/rebaser.py#L665). I am not sure if application code or external systems rely on that. So there might be a small possibility that things will break. But maybe we just go ahead and see if anyone complains? 

>  that in konflux world distgit_key will be the definitive package "name" in nvr, right?
Yeah, our konflux build uses distgit_key.

PR #1400, opened 2025-03-03
Fix upstream url for rpms

thegreyd:
Right now rpms in Konflux db have source_repo set to openshift-priv url.
It should instead use the public upstream url.
Also rely on object var which contains build commit instead of parsing specfile.

ref: https://github.com/openshift-eng/art-tools/blob/e7689d92bf2eea33c47f91aa8cf15e82a7ea3ae7/doozer/doozerlib/rpm_builder.py#L179-L180


locriandev (approved):
/lgtm

PR #1404, opened 2025-03-04
Fix lock name for Brew build-sync

locriandev:
Currently all locks for build-sync reference `build-sync-konflux`. Fixing this for Brew builds


locriandev (commented):
/approve

PR #1407, opened 2025-03-04
[konflux] disable ppc

ashwindasr:
Since PPC VMs on Konflux are not spawning correctly, disabling triggering builds temporarily until Konflux fixes are in. 

ref [thread](https://redhat-internal.slack.com/archives/C06Q309QUDV/p1740589236365859) for details


locriandev (commented):
/lgtm

PR #1410, opened 2025-03-04
[konflux] konflux_arches param in group.yaml

ashwindasr:


ashwindasr:
/hold

Until https://github.com/openshift-eng/ocp-build-data/pull/6280 is merged and cherry picked to all lower versions (4.18-4.12)

ashwindasr:
/unhold

Added [fallback](https://github.com/openshift-eng/art-tools/pull/1410/files#diff-1406051f1a04a744fb77f4c5faaac7f4b33f803c82506229b6e02bc7499e744fR785) so is not dependent on konflux_arches param to be present

doozer/doozerlib/runtime.py
```diff
@@ -774,6 +774,18 @@ def get_global_arches(self):
         """
         return list(self.arches)
 
+    def get_global_konflux_arches(self):
+        """
+        :return: Returns a list of architectures that are enabled globally in group.yml, for konflux.
+        """
+        # For now, cli override (LIMIT_ARCHES) and arches_override in group config are not supported
+        arches = list(self.group_config.konflux_arches)
```
vfreex, line 782: Suggesting to add a new key `konflux` for all konflux-specific options and move `arches` under `konflux`. e.g.

```yaml
konflux:
  arches: ['x86_64']
```
Reasons:
1. We already use this model in image metadata config, e.g. https://github.com/openshift-eng/ocp-build-data/blob/5436cfe344a48898be6a3fcb32a9bb0d046082d2/images/driver-toolkit.yml#L41. This will look more consistent.
2. In the future when we add more konflux-specific options, we can use a similar approach to merge options under `konflux` to global group options.

doozer/doozerlib/metadata.py
```diff
@@ -284,21 +284,23 @@ def hotfix_brew_tags(self):
         """
         return [self.hotfix_brew_tag()]
 
-    def get_arches(self):
+    def get_arches(self, konflux: bool = False):
```
vfreex, line 287: Guess we no longer need to explicitly pass this argument. `runtime.build_system` can be used to determine if the current build system is konflux or not.

ashwindasr: @vfreex looks like there was a small bug, we weren't initializing the build system for `KonfluxBuildCli` till now.

Pushed the fix to this PR: https://github.com/openshift-eng/art-tools/pull/1410/commits/44c51b8b5ec5f6f2bc58cf6565b5cbad65d4e423

locriandev: this is surprising, I suggest we add the `--build-system` to the Doozer base command [here](https://github.com/openshift-eng/art-tools/blob/main/pyartcd/pyartcd/pipelines/ocp4_konflux.py#L71)

ashwindasr: Good idea, updating now

ashwindasr: updated: https://github.com/openshift-eng/art-tools/pull/1410/commits/fc233d6e40a727d4056357717dc498fa047c17e4 @locriandev

ashwindasr: hmm wait that doesn't seem to be working

ashwindasr: ahh nvm, I was testing the doozer command directly. its working

locriandev (commented):
/lgtm

PR #1412, opened 2025-03-06
[konflux] cachito emulation update

ashwindasr:
In some cases, in image config, metadata.config.content.source.pkg_managers will be populated but metadata.config.cachito.packages.npm will be empty. In that case, we assume that the path is the current directory

### DIFF
--- a/doozer/doozerlib/backend/rebaser.py
+++ b/doozer/doozerlib/backend/rebaser.py
@@ -872,31 +872,57 @@ class KonfluxRebaser:
 
             # The value we will set REMOTE_SOURCES_DIR to.
             remote_source_dir_env = '/tmp/cachito-emulation'
+            pkg_managers = metadata.config.content.source.pkg_managers
 
-            for npm_entry in metadata.config.cachito.packages.npm:
-                # cachito creates an .npmrc file in the "npm" package manager paths. This
-                # used to cause NPM to use an NPM server, supplied by cachito,
-                # which only supplied the component's NPM deps.
-                # See https://github.com/containerbuildsystem/cachito/blob/fe2ef761d7d8169597166e9f92ee8ef3d62a091c/README.md .
-                # Dockerfiles may reference this file. We create it in emulation mode, but do
-                # not populate it. cachi2 works fundamentally differently
-                # by re-writing package-lock.json to reference local files.
-                # https://github.com/hermetoproject/cachi2/blob/2235854bfc5cb76ea168cb04d50be16efce9fa0a/docs/npm.md
-                if 'path' in npm_entry:
+            if "npm" in pkg_managers:
+                flag = False
+                for npm_entry in metadata.config.cachito.packages.npm:
+                    # cachito creates an .npmrc file in the "npm" package manager paths. This
+                    # used to cause NPM to use an NPM server, supplied by cachito,
+                    # which only supplied the component's NPM deps.
+                    # See https://github.com/containerbuildsystem/cachito/blob/fe2ef761d7d8169597166e9f92ee8ef3d62a091c/README.md .
+                    # Dockerfiles may reference this file. We create it in emulation mode, but do
+                    # not populate it. cachi2 works fundamentally differently
+                    # by re-writing package-lock.json to reference local files.
+                    # https://github.com/hermetoproject/cachi2/blob/2235854bfc5cb76ea168cb04d50be16efce9fa0a/docs/npm.md
+                    if 'path' in npm_entry:
+                        app_path = gomod_deps_path.joinpath('app')
+                        app_path = app_path.joinpath(npm_entry['path'])
+                        app_path.mkdir(parents=True, exist_ok=True)
+                        app_path.joinpath('.npmrc').touch(exist_ok=True)
+                        flag = True
+
+                if not flag:
+                    # In some cases, in image config, metadata.config.content.source.pkg_managers will be populated
+                    # but metadata.config.cachito.packages.npm will be emtpy. In that case, we assume that the path
+                    # is the current directory
                     app_path = gomod_deps_path.joinpath('app')
-                    app_path = app_path.joinpath(npm_entry['path'])
+                    app_path = app_path.joinpath(".")
                     app_path.mkdir(parents=True, exist_ok=True)
                     app_path.joinpath('.npmrc').touch(exist_ok=True)
 
-            for npm_entry in metadata.config.cachito.packages.yarn:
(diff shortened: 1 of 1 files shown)


doozer/doozerlib/backend/rebaser.py
```diff
@@ -872,31 +872,57 @@ def _add_build_repos(self, dfp: DockerfileParser, metadata: ImageMetadata, dest_
...
-                # cachito creates an .npmrc file in the "npm" package manager paths. This
-                # used to cause NPM to use an NPM server, supplied by cachito,
-                # which only supplied the component's NPM deps.
-                # See https://github.com/containerbuildsystem/cachito/blob/fe2ef761d7d8169597166e9f92ee8ef3d62a091c/README.md .
-                # Dockerfiles may reference this file. We create it in emulation mode, but do
-                # not populate it. cachi2 works fundamentally differently
-                # by re-writing package-lock.json to reference local files.
-                # https://github.com/hermetoproject/cachi2/blob/2235854bfc5cb76ea168cb04d50be16efce9fa0a/docs/npm.md
-                if 'path' in npm_entry:
+            pkg_managers = metadata.config.content.source.pkg_managers
```
vfreex, line 875: In Brew build, we don't really fully rely on this option. If not explicitly specified, Doozer will guess what package managers are used. See https://github.com/openshift/art-tools/blob/6f2ad7415184d70074f3c4b39d1f8f5e9697a983/doozer/doozerlib/distgit.py#L587

doozer/doozerlib/backend/rebaser.py
```diff
@@ -872,31 +872,57 @@ def _add_build_repos(self, dfp: DockerfileParser, metadata: ImageMetadata, dest_
...
+                for npm_entry in metadata.config.cachito.packages.npm:
+                    # cachito creates an .npmrc file in the "npm" package manager paths. This
+                    # used to cause NPM to use an NPM server, supplied by cachito,
+                    # which only supplied the component's NPM deps.
+                    # See https://github.com/containerbuildsystem/cachito/blob/fe2ef761d7d8169597166e9f92ee8ef3d62a091c/README.md .
+                    # Dockerfiles may reference this file. We create it in emulation mode, but do
+                    # not populate it. cachi2 works fundamentally differently
+                    # by re-writing package-lock.json to reference local files.
+                    # https://github.com/hermetoproject/cachi2/blob/2235854bfc5cb76ea168cb04d50be16efce9fa0a/docs/npm.md
+                    if 'path' in npm_entry:
```
vfreex, line 888: If `path` doesn't exist, it should default to `.`, then the .npmrc file located at the top directory of the source code. So a proper implementation could be:

```python
app_path = gomod_deps_path.joinpath('app').joinpath(npm_entry.get('path', '.'))
```

doozer/doozerlib/backend/rebaser.py
```diff
@@ -872,31 +872,57 @@ def _add_build_repos(self, dfp: DockerfileParser, metadata: ImageMetadata, dest_
...
+                        app_path = gomod_deps_path.joinpath('app')
+                        app_path = app_path.joinpath(npm_entry['path'])
+                        app_path.mkdir(parents=True, exist_ok=True)
+                        app_path.joinpath('.npmrc').touch(exist_ok=True)
+                        flag = True
+
+                if not flag:
```
vfreex, line 895: If `metadata.config.cachito.packages.npm` is empty, it means nothing to customize for the npm packager. We can add an empty entry to `metadata.config.cachito.packages.npm` to simplify the implementation.

doozer/doozerlib/backend/rebaser.py
```diff
@@ -872,31 +872,57 @@ def _add_build_repos(self, dfp: DockerfileParser, metadata: ImageMetadata, dest_
...
+                    # In some cases, in image config, metadata.config.content.source.pkg_managers will be populated
+                    # but metadata.config.cachito.packages.npm will be emtpy. In that case, we assume that the path
+                    # is the current directory
                     app_path = gomod_deps_path.joinpath('app')
-                    app_path = app_path.joinpath(npm_entry['path'])
+                    app_path = app_path.joinpath(".")
```
vfreex, line 900: joining `.` is actually a no-op.

doozer/doozerlib/backend/rebaser.py
```diff
@@ -872,31 +872,57 @@ def _add_build_repos(self, dfp: DockerfileParser, metadata: ImageMetadata, dest_
...
                     app_path.joinpath('.npmrc').touch(exist_ok=True)
 
-            for npm_entry in metadata.config.cachito.packages.yarn:
-                # cachito creates an .npmrc/.yarnrc files in the "yarn" package manager paths. This
-                # used to cause NPM/yarn to use servers supplied by cachito,
-                # cachi2 works fundamentally differently
-                # by re-writing package-lock.json to reference local files.
-                # https://github.com/hermetoproject/cachi2/blob/2235854bfc5cb76ea168cb04d50be16efce9fa0a/docs/npm.md
-                if 'path' in npm_entry:
+            if "yarn" in pkg_managers:
```
vfreex, line 904: Same here.

doozer/doozerlib/backend/rebaser.py
ashwindasr: Thanks for pointing me to `_detect_package_managers`, updated.

doozer/doozerlib/backend/rebaser.py
ashwindasr: The issue is that, in some cases, in image config, metadata.config.content.source.pkg_managers will be populated but metadata.config.cachito.packages.npm will be empty. In that case, we assume that the path is the current directory

doozer/doozerlib/backend/rebaser.py
ashwindasr: I've updated the code to use `for-else` for simplicity

doozer/doozerlib/backend/rebaser.py
ashwindasr: Thanks [updated](https://github.com/openshift-eng/art-tools/pull/1412/commits/6568247ab7d2c7df0632e7443dc7ab7d672d6924)

doozer/doozerlib/backend/rebaser.py
ashwindasr: Thanks [updated](https://github.com/openshift-eng/art-tools/pull/1412/commits/6568247ab7d2c7df0632e7443dc7ab7d672d6924)

PR #1413, opened 2025-03-06
find-builds: Check for False instead of None

thegreyd:
Because of the recent change of get_latest_build returning Model/MissingModel,
we want to check for bool(MissingModel)/False instead of None.


locriandev (approved):
/lgtm

PR #1414, opened 2025-03-06
uv run when running make tasks

thegreyd:
Let uv handle venv instead of activating/hardcoding it


rayfordj:
Not familiar with this and I might be misunderstanding the use of 'assume' in this context :grin: , but could some _check_ be in place to for the existence of, for example, `deactivate` that is often found when in a venv?

```
[[ $(type -t deactivate) == function ]] && echo 'in a venv'
```

rayfordj:
> Not familiar with this and I might be misunderstanding the use of 'assume' in this context 😁 , but could some _check_ be in place to for the existence of, for example, `deactivate` that is often found when in a venv?
> 
> ```
> [[ $(type -t deactivate) == function ]] && echo 'in a venv'
> ```

Ignore me; you've already rewritten the PR by the time I go the comment added.  :sweat_smile:

thegreyd:
@rayfordj I discovered "uv run" and used that instead :) https://docs.astral.sh/uv/reference/cli/#uv-run

locriandev (commented):
/lgtm

PR #1416, opened 2025-03-07
fix find builds issue

Ximinhan:
- fix the issue that get_latest_build return empty model instead of the default value set in args
- for el_target should honor the config in component instead of the global config
- revert pr https://github.com/openshift-eng/art-tools/pull/1413 since it didn't fix the issue and raise error incorrectly when no build found we should silently exit

### DIFF
--- a/artcommon/artcommonlib/metadata.py
+++ b/artcommon/artcommonlib/metadata.py
@@ -481,8 +481,8 @@ class MetadataBase(object):
             return default
         return build_record
 
-    async def get_latest_brew_build_async(self, **kwargs) -> Model:
-        return Model(await asyncio.to_thread(self.get_latest_brew_build, **kwargs))
+    async def get_latest_brew_build_async(self, **kwargs):
+        return await asyncio.to_thread(self.get_latest_brew_build, **kwargs)
 
     async def get_latest_build(self, **kwargs):
         """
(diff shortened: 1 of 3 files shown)


elliott/elliottlib/cli/find_builds_cli.py
```diff
@@ -478,9 +472,9 @@ async def _fetch_builds_by_kind_rpm(runtime: Runtime, tag_pv_map: Dict[str, str]
     pinned_nvrs = set()
     if member_only:  # Sweep only member rpms
         for tag in tag_pv_map:
-            tasks = [rpm.get_latest_build(default=None, el_target=tag) for rpm in runtime.rpm_metas()]
+            tasks = [rpm.get_latest_build(default=None, el_target=tag) for rpm in runtime.rpm_metas() if tag in rpm.determine_targets()]
```
Ximinhan, line 475: the tag used for el_target should come from component config

artcommon/artcommonlib/metadata.py
```diff
@@ -481,8 +481,8 @@ async def get_latest_konflux_build(self,
             return default
         return build_record
 
-    async def get_latest_brew_build_async(self, **kwargs) -> Model:
-        return Model(await asyncio.to_thread(self.get_latest_brew_build, **kwargs))
```
Ximinhan, line 485: get_latest_brew_build has a default parameter used to define the return value while the model breaks the type

PR #1422, opened 2025-03-11
images:health look back to latest 90 days

locriandev:
Look at latest 90 days, as 30 could not be enough for slowly updating images and might result in unwanted pings.

Also, no need to filter builds in `PENDING` state as `search_builds_by_field()` is doing that in a smarter way:

```
if 'outcome' not in where:
    base_clauses.append(Column('outcome', String).in_(['success', 'failure']))
```


joepvd (approved):
/lgtm

PR #1430, opened 2025-03-12
Konflux rebaser: use Konflux DB to find previous versions

locriandev:


doozer/doozerlib/backend/rebaser.py
```diff
@@ -1826,30 +1831,16 @@ def _map_image_name(name, image_map):
...
-
-            return versions
+        builds = [build async for build in self._runtime.konflux_db.search_builds_by_fields(
+            where={
+                'name': metadata.get_component_name(),
+                'outcome': str(KonfluxBuildOutcome.SUCCESS),
+                'group': self._runtime.group
+            }
+        )]
+        return set([f'{build.version.strip("v")}-{build.release.split(".")[0]}' for build in builds])
```
thegreyd, line 1846: Is there a unit test that tests this? :sweat_smile:

doozer/doozerlib/backend/rebaser.py
```diff
@@ -276,7 +281,7 @@ def _resolve_parents(self, metadata: ImageMetadata, dfp: DockerfileParser, image
         mapped_images: List[Tuple[str, bool]] = []
         for parent, original_parent in zip(parents, dfp.parent_images):
             if "member" in parent:
-                mapped_images.append(self._resolve_member_parent(parent["member"], original_parent, image_repo, uuid_tag))
+                mapped_images.append(await self._resolve_member_parent(parent["member"], original_parent, image_repo, uuid_tag))
```
thegreyd, line 284: stylistic suggestion, to split these into 2 lines for readability

doozer/doozerlib/backend/rebaser.py
```diff
@@ -1826,30 +1831,16 @@ def _map_image_name(name, image_map):
             # Re-write the CSV content.
             Path(csv_file).write_text(yaml.dump(csv_obj))
 
-    def _find_previous_versions(self, metadata: ImageMetadata, pattern_suffix='') -> Set[str]:
+    async def _find_previous_versions(self, metadata: ImageMetadata, pattern_suffix='') -> Set[str]:
         """
         Returns: Searches brew for builds of this operator in order and processes them into a set of versions.
         These version may or may not have shipped.
         """
-        # FIXME: This is based on Brew/OSBS. Needs to figure out how to do this with Konflux
```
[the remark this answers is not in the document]
locriandev: No, this is part of the `KonfluxRebaser` class

doozer/doozerlib/backend/rebaser.py
locriandev: Not against it but I'd rather not touch anything more than what's strictly needed for the fix

doozer/doozerlib/backend/rebaser.py
```diff
@@ -1826,30 +1830,16 @@ def _map_image_name(name, image_map):
...
-            for nvr in nvrs:
-                without_component = nvr[len(f'{component_name}-v'):]  # e.g. "4.10.0-202211280957.p0.ga42b581.assembly.stream"
-                version_components = without_component.split('.')[0:3]  # e.g. ['4', '10', '0-202211280957']
-                version = '.'.join(version_components)
-                versions.add(version)
-
-            return versions
+        builds = [build async for build in self._runtime.konflux_db.search_builds_by_fields(
+            where={
+                'name': metadata.get_component_name(),
```
thegreyd, line 1840: if i'm not wrong, "name" is always recorded as meta.distgit_key, so let's use that? Also the reason of restoring "-container", which would mean component and disgit_key, which right now are the same, would be different in future.

doozer/doozerlib/backend/rebaser.py
```diff
@@ -1826,30 +1830,16 @@ def _map_image_name(name, image_map):
...
-            # We need to convert these into versions like "4.11.0-202205250107"
-            versions: Set[str] = set()
-            for nvr in nvrs:
-                without_component = nvr[len(f'{component_name}-v'):]  # e.g. "4.10.0-202211280957.p0.ga42b581.assembly.stream"
-                version_components = without_component.split('.')[0:3]  # e.g. ['4', '10', '0-202211280957']
-                version = '.'.join(version_components)
-                versions.add(version)
-
-            return versions
+        builds = [build async for build in self._runtime.konflux_db.search_builds_by_fields(
```
thegreyd, line 1838: how many builds would this get .. :thinking: is there a max limit? did brew api have a max limit?

doozer/doozerlib/backend/rebaser.py
locriandev: Done with 1329a5745ab4be52096578d638a16b287280a5a9

vfreex:
Thanks @locriandev . The skips field is used by IIB to populate the upgrade graph. However with FBC, this might not be needed anymore. Do you mind I temporarily holding this PR before we decide how we should manage skips in FBC?
/hold

PR #1432, opened 2025-03-13
[temp][konflux] disable rebuilds on config changes

locriandev:
**TO BE REVERTED!!!**

Temporarily disable konflux rebuilds on config changes. Some images are continuously being rebuilt, so something might be wrong with our metadata configuration check. Disabling it until we assess the problem


vfreex (approved):
/lgtm

PR #1433, opened 2025-03-13
Fix scan-sources config changes check

locriandev:
`scan-sources` and 'images:rebase` compute image config digest at different times; the latter happens after merging the `konflux` config stanza into the main config, and this alters the result. As a consequence, we were continuously finding reasons to rebuild images because the digest computed by scan-sources did not match what we had pushed upstream. 

Moving the config merge inside the function that computes the digest will ensure consistent results across pipelines.

Test build: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-scan-konflux/12930/console

### DIFF
--- a/doozer/doozerlib/backend/rebaser.py
+++ b/doozer/doozerlib/backend/rebaser.py
@@ -95,13 +95,8 @@ class KonfluxRebaser:
             commit_message: str,
             push: bool,
     ) -> None:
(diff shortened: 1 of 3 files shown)


vfreex (commented):
/lgtm

PR #1446, opened 2025-03-19
[konflux] fix component_name

ashwindasr:
https://github.com/openshift-eng/art-tools/pull/1446#issuecomment-2738165059


thegreyd:
That's because we made a modification here: https://github.com/openshift-eng/art-tools/pull/1399/files 
I think the proper fix would be to use get_component_name() when building NVRs - can we do that?

ashwindasr:
/unhold

Updated PR

ashwindasr:
Rebase succeeded: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-konflux/6698/console

PR #1451, opened 2025-03-20
elliott release watch

thegreyd:


.pylintrc
```diff
@@ -16,7 +16,7 @@ clear-cache-post-run=no
 # In error mode, messages with a category besides ERROR or FATAL are
 # suppressed, and no reports are done by default. Error mode is compatible with
 # disabling specific errors.
-errors-only=
+#errors-only=
```
locriandev, line 19: Let's maybe take this out, since you added `--errors-only`

doozer/doozerlib/backend/konflux_client.py
```diff
@@ -712,3 +714,84 @@ def _inner():
...
+        """
+        Wait for a Release to complete.
+
+        :param release_name: The name of the Release.
+        :param namespace: The namespace of the Release.
+        :param overall_timeout_timedelta: Maximum time to wait for release to complete before exiting (defaults to 5 hour)
+        :return: The Release ResourceInstance
+        """
+        if overall_timeout_timedelta is None:
+            overall_timeout_timedelta = datetime.timedelta(hours=5)
```
locriandev, line 729: These hard-coded constants are always sneaky to find when we want to change them... maybe there's a better place for them?

doozer/doozerlib/backend/konflux_client.py
```diff
@@ -712,3 +714,84 @@ def _inner():
...
+
+        namespace = namespace or self.default_namespace
+        api = await self._get_api(API_VERSION, KIND_RELEASE)
+
+        if self.dry_run:
+            await asyncio.sleep(3)
```
locriandev, line 735: No need to wait during a dry run?

doozer/doozerlib/backend/konflux_client.py
```diff
@@ -712,3 +714,84 @@ def _inner():
...
+                        if datetime.datetime.now() > timeout_datetime:
+                            self._logger.info("Timeout reached. Exiting..")
+                            watcher.stop()
+                            return obj
+
+                    self._logger.info("No updates for Release %s during watch timeout period; requerying",
+                                      release_name)
+                except TimeoutError:
+                    self._logger.error("Timeout waiting for Release %s to complete", release_name)
+                    continue
```
locriandev, line 788: If we `continue` on TimeoutError the while loop will restart, and this is true until `datetime.datetime.now() > timeout_datetime:`. In this case we log `Timeout reached. Exiting..` and return. So wondering, what's the point of logging an error here?

elliott/elliottlib/cli/konflux_release_cli.py
```diff
@@ -0,0 +1,93 @@
...
+        'kubeconfig': konflux_kubeconfig,
+        'namespace': konflux_namespace,
+        'context': konflux_context,
+    }
+
+    pipeline = WatchReleaseCli(runtime, release=release, konflux_config=konflux_config, timeout=timeout,
+                               dry_run=dry_run)
+    release_status = await pipeline.run()
+    if release_status is True:
+        print("Release successful!")
```
locriandev, line 89: `click.echo`?

elliott/elliottlib/cli/konflux_release_cli.py
```diff
@@ -0,0 +1,93 @@
...
+                                                            context=self.konflux_config['context'],
+                                                            dry_run=self.dry_run)
+        self.konflux_client.verify_connection()
+
+    async def run(self):
+        self.runtime.initialize(no_group=True)
+        release_obj = await self.konflux_client.wait_for_release(self.release, overall_timeout_timedelta=timedelta(hours=self.timeout))
+
+        # Assume that these will be available
+        released_condition = next(c for c in release_obj['status']['conditions'] if c['type'] == "Released")
```
locriandev, line 39: Need to handle `StopIteration`?

thegreyd: I think it is okay to error out. But a better exception can be raised

elliott/elliottlib/cli/konflux_release_cli.py
thegreyd: :+1:

.pylintrc
thegreyd: That is the convention of pylintrc file - by default it was explained and commented out, similar to other config settings

doozer/doozerlib/backend/konflux_client.py
thegreyd: I just followed the pattern we set in `wait_for_pipelinerun` above this :)

doozer/doozerlib/backend/konflux_client.py
thegreyd: https://github.com/kubernetes-client/python/blob/master/examples/watch/timeout-settings.md#sever-side-timeout-kwargstimeout_seconds--n 

In case there is a server side error, it will be useful to know

locriandev (commented):
/lgtm

PR #1466, opened 2025-03-25
[ART-12287] Update rhcos job to support target job name

Ximinhan:
To support trigger different rhcos job, the pr add a job name parameter to our rhcos job, also make some changes in the pipeline

* change the triggering method, we used to trigger a build and then wait in loop to monitor its complete, change the trigger API method to directly get the build id so no need to loop and use extra logic to find the build id, and no need to wait until it complete, the triggered job is rhcos x86_64 it just a first step of a rhcos build, it complete with success doesn't mean new rhcos build is created.
* change the get rhcos stream method, we used to use a doozer command to get the group config and then parsing the string value to get the rhcos stream. I make it easier to directly look at the raw source file and explicitly get the rhcos vars
* also removed some unnecessary parameter checks, this command should be used in job invoking, still keep dry-run option but it doesn't seem have much value.

need https://github.com/openshift-eng/aos-cd-jobs/pull/4383


pyartcd/pyartcd/pipelines/build_rhcos.py
```diff
@@ -58,184 +52,72 @@ def __init__(self, runtime: Runtime, new_build: bool, ignore_running: bool, vers
...
-            # when the above check cannot see it, and so this may start a new one. interestingly,
-            # Jenkins seems to silently ignore identical builds in close succession (probably
-            # double-click protection - if the parameters differ both are started), and should this
-            # happen we will see a build start, assume we started it, and watch it to completion.
-            # this seems unlikely to cause any problems other than mild confusion.
-            self.start_build()
-            if self.dry_run:
-                print('DRY RUN - Exiting', file=sys.stderr)
-            result["action"] = "build"
-            result["builds"] = self.wait_for_builds()
```
Ximinhan, line 79: we used to start_build then wait for builds, now start_build should be able to get build id

pyartcd/pyartcd/pipelines/build_rhcos.py
```diff
@@ -58,184 +52,72 @@ def __init__(self, runtime: Runtime, new_build: bool, ignore_running: bool, vers
...
-        if match:
-            self._stream = match[1]
-        else:
-            self._stream = self.version
-        return self._stream
+        """Check if there are any existing builds for the given job with given version. Returns builds dicts."""
+        if self.ignore_running:
+            return []
+        response = self.request_session.get(
+            f"{JENKINS_BASE_URL}/job/{self.job}/api/json?tree=builds[number,description,url,result,actions[parameters[name,value]]]",
```
Ximinhan, line 81: the query api and return url, so no need to call build_url

pyartcd/pyartcd/pipelines/build_rhcos.py
```diff
@@ -58,184 +52,72 @@ def __init__(self, runtime: Runtime, new_build: bool, ignore_running: bool, vers
...
-                    builds.append(build)
-
-        return [b for b in builds if b["parameters"].get("STREAM") == self.stream]
-
-    @property
-    def stream(self):
-        if self._stream:
-            return self._stream
-
-        # doozer --quiet -g openshift-4.14 config:read-group urls.rhcos_release_base.multi --default ''
```
Ximinhan, line 141: we used to read the config in group and determining if rhcos migrate from old server to new server, now there is only one rhcos server, the stream used for rhcos job parameter should be simplify to directly get the vars from group.

pyartcd/pyartcd/pipelines/build_rhcos.py
```diff
@@ -58,184 +52,72 @@ def __init__(self, runtime: Runtime, new_build: bool, ignore_running: bool, vers
...
+        build_number = self.trigger_build(job_url, params)
+        build = self.request_session.get(f"{JENKINS_BASE_URL}/job/{self.job}/{build_number}/api/json").json()
+        return dict(url=build['url'], result=build['result'], description=build['description'])
+
+    def trigger_build(self, job_url, params):
+        """trigger build and return build bumber"""
+        response = self.request_session.post(job_url, data=params)
+        if response.status_code not in (201, 200):
+            raise Exception(f"Failed to trigger rhcos build: {response}")
+        queue_url = response.headers.get('Location')
```
Ximinhan, line 113: now we know trigger a jenkins job will return a queue, and wait until the queue start it will get a build id

pyartcd/pyartcd/pipelines/build_rhcos.py
```diff
@@ -8,45 +8,39 @@
 import json
 import time
 import sys
-import subprocess
-from subprocess import PIPE
+import yaml
```
joepvd, line 11: Please use `ruamel.yaml`

pyartcd/pyartcd/pipelines/build_rhcos.py
```diff
@@ -245,8 +127,8 @@ def wait_for_builds(self):
               help="Ignore in-progress builds instead of just exiting like usual")
 @click.option("--new-build", required=False, default=False, type=bool,
               help="Force a new build even if no changes were detected from the last build")
+@click.option("--job", required=False, type=str, default="build",
```
joepvd, line 130: Can this be a list of options? like `['build', 'build-node-image']`? Think we can list the types that we support here.

pyartcd/pyartcd/pipelines/build_rhcos.py
```diff
@@ -58,184 +52,72 @@ def __init__(self, runtime: Runtime, new_build: bool, ignore_running: bool, vers
...
+                    for action in build["actions"]
+                    if "parameters" in action
+                    for param in action["parameters"]
+        )]
+
+    def get_stream(self):
+        """Get rhcos job's stream parameter value, for 4.12+ it looks like 4.x-9.x for 4.12 is just 4.12"""
+        group_response = self.request_session.get(f"{OCP_REPO_RAW}/openshift-{self.version}/group.yml")
+        group_file = yaml.load(group_response.text, Loader=yaml.SafeLoader)
+        return f"{self.version}-{group_file['vars']['RHCOS_EL_MAJOR']}.{group_file['vars']['RHCOS_EL_MINOR']}" if group_file['vars']['MINOR'] > 12 else self.version
```
joepvd, line 98: Could you make an assertion about ``MAJOR` as well, so this does not break when we start openshift5?

pyartcd/pyartcd/pipelines/build_rhcos.py
```diff
@@ -58,184 +52,72 @@ def __init__(self, runtime: Runtime, new_build: bool, ignore_running: bool, vers
...
+                and any(
+                    param["name"] == "STREAM" and param["value"] == self._stream  # check build parameter with given stream
+                    for action in build["actions"]
+                    if "parameters" in action
+                    for param in action["parameters"]
+        )]
+
+    def get_stream(self):
+        """Get rhcos job's stream parameter value, for 4.12+ it looks like 4.x-9.x for 4.12 is just 4.12"""
+        group_response = self.request_session.get(f"{OCP_REPO_RAW}/openshift-{self.version}/group.yml")
```
joepvd, line 96: Think raw.githubuserccontent.com content will read from cache, and there is no guarantee that it is up to date. Which is fine almost always, but just when we change this, we want this to be right. 

Soon, we'll want to be able to trigger this for `4.20-9.6` and `4.20-10.0`. Maybe the rhel major/minor should also be a job input? Not sure about the best place to look for that info, and if this is the PR to think about that.

Ximinhan: we just need to el version vars, the current design is to read from config, of course it can be set from parameter, but I don't think the el version will frequently change. (if we use parameter to set the stream, then maybe rhcos vars and config in group it not needed anymore, eg after this change the multi config is not needed anymore)

pyartcd/pyartcd/pipelines/build_rhcos.py
Ximinhan: do you mean MAJOR <= 4 ?

pyartcd/pyartcd/pipelines/build_rhcos.py
Ximinhan: I put it in a list in the Jenkins job parameter https://github.com/openshift-eng/aos-cd-jobs/pull/4383/files

pyartcd/pyartcd/pipelines/build_rhcos.py
Ximinhan: if we use the parameter to set stream, then the parent job that triggers it also needs to decide the value, the logic seem still needed

joepvd: Current logic is to run `doozer`, which runs a clone and gets the current value. 
Proposed logic is to ask GitHub cache about the value. This is wrong for a while when this value gets changed. This is a type of defect that is very time consuming to debug. Let's not have this bug.

joepvd (approved):
/lgtm
Yet
/hold
Unhold at will, Ximin!

PR #1469, opened 2025-03-25
[konflux] fix yum line rebase

ashwindasr:
ref [thread](https://redhat-internal.slack.com/archives/C06Q309QUDV/p1742914239326569?thread_ts=1742477914.887549&cid=C06Q309QUDV)


ashwindasr:
fixed tests

PR #1471, opened 2025-03-25
[konflux] .dockerignore fix

ashwindasr:
If a .dockerignore file exists, we update it to allow .oit dir: https://github.com/openshift-eng/art-tools/blob/main/doozer/doozerlib/backend/rebaser.py#L257

But this causes an error with pre-fetch dependencies task, where it complains that the go vendors are not upto date. Hence run `go mod vendor` to update.


vfreex (commented):
/hold

I will be hesitate to run `go mod vendor` by ourselves. This will change the upstream code, potentially break upstream's assumption, and probably break things by using a wrong version of golang toolchain.

Let's have a conversation with the team before adopting this.

PR #1472, opened 2025-03-26
promote-assembly: sign with stage RADAS in DRY_RUN mode

vfreex:
Currently if the job is run in DRY_RUN mode, no signing requests will be sent to RADAS. With this change, artifacts will be signed with stage RADAS but resulting signature files will not be published.


joepvd (approved):
/lgtm

PR #1473, opened 2025-03-26
Use p2/p3 flags for Konflux releases

locriandev:


doozer/doozerlib/util.py
```diff
@@ -241,18 +241,32 @@ def isolate_git_commit_in_release(release: str) -> Optional[str]:
...
+    For Brew .p0 is associatedd with public builds, .p1 with private ones.
+    For Konflux .p2 is associatedd with public builds, .p3 with private ones.
+    """
+
+    pflag = isolate_pflag_in_release(release)
+    if runtime.build_system == 'brew':
+        return pflag == 'p1'
+    if runtime.build_system == 'konflux':
+        return pflag == 'p3'
+    return False
```
joepvd, line 267: Should this default to True? Better consider it private unless proven otherwise?

thegreyd:
I have some thoughts about this. When it was p0/p1 it was easy to remember and know what means what when reading code. But with 4 values now, it can cause confusion - and it can be potentially very bad if we have such a bug..

So I would propose something like this
```python
class EmbargoStatus(Enum):
    BREW_NO = "p0"
    BREW_YES = "p1"
    KONFLUX_NO = "p2"
    KONFLUX_YES = "p3"
```

doozer/doozerlib/cli/config_plashet.py
```diff
@@ -586,11 +588,12 @@ def from_tags(config: SimpleNamespace, brew_tag: Tuple[Tuple[str, str], ...], em
...
 
             def is_embargoed(an_nvre):
-                # .p1 or inclusion in the embargoed_tag_nvrs indicates this rpm is embargoed OR *was* embargoed.
+                # .p1/.p3 or inclusion in the embargoed_tag_nvrs indicates this rpm is embargoed OR *was* embargoed.
                 # We can ignore it if it has already shipped.
                 test_nvre_obj = parse_nvr(an_nvre)
                 if released_nvre_obj is None or compare_nvr(test_nvre_obj, released_nvre_obj) > 0:  # If this nvr hasn't shipped
-                    if '.p1' in an_nvre or strip_epoch(an_nvre) in embargoed_tag_nvrs:  # It's embargoed!
+                    pflag = get_visibility_suffix(runtime.build_system, BuildVisibility.PRIVATE)
+                    if f'.{pflag}' in (an_nvre or '') or strip_epoch(an_nvre) in embargoed_tag_nvrs:  # It's embargoed!
```
thegreyd, line 596: I think it would be nice to have a `is_nvr_embargoed` / `is_nvre_embargoed` convenience methods to make it easier for the caller to determine this

doozer/doozerlib/build_visibility.py
```diff
@@ -0,0 +1,80 @@
...
+
+# from doozerlib.util import isolate_pflag_in_release
+
+
+class BuildVisibility(Enum):
+    PUBLIC = 0
+    PRIVATE = 1
+
+
+VISIBILITY_SUFFIX = {
```
thegreyd, line 14: sorry i think i'm nitpicking a bit. but why a dict instead of another Enum?

```
class PFlag(Enum):
    BREW_PUBLIC = "p0"
    BREW_PRIVATE = "p1"
    KONFLUX_PUBLIC = "p2"
    KONFLUX_PRIVATE = "p3"

    # a regular class attr to save config will misbehave, so make available via a classmethod
    @classmethod
    def get_config(cls):
        config_map = {
            "p0": ("brew", BuildVisibility.PUBLIC),
            "p1": ("brew", BuildVisibility.PRIVATE),
            "p2": ("konflux", BuildVisibility.PUBLIC),
            "p3": ("konflux", BuildVisibility.PRIVATE)
        }
        return config_map

    @property
    def build_system(self):
        key = self.value
        return self.get_config()[key][0]

    @property
    def embargoed(self):
        key = self.value
        return self.get_config()[key][1] == BuildVisibility.PRIVATE
```

thegreyd:
Discussed with Daniele that these suggestions can be implemented as our need for them becomes clearer.

PR #1475, opened 2025-03-27
ART-9439 - fixes after rebase

mbiarnes:
This reverts commit 974a4bd26676390650a4b35e4b383805e93855d7.


thegreyd:
lg!
let's fix the lint errors and then it should be good

PR #1476, opened 2025-03-27
[konflux] .dockerignore fix

ashwindasr:
`hypershift` image has a `.gitignore` file that needs to be present after rebase, in konflux. Usually its removed before rebase to make sure that files/folders that doozer generates are not  excluded by git. But in order to work with cachi2, we need to temporarily rename the file, add all the .oit/* and other files that doozer generates, rename the file back to .gitignore so that konflux can read the correct file.

### DIFF
--- a/doozer/doozerlib/backend/build_repo.py
+++ b/doozer/doozerlib/backend/build_repo.py
@@ -162,10 +162,14 @@ class BuildRepo:
             raise ChildProcessError(f"Failed to get commit hash: {err}")
         return out.strip()
 
+    async def git_add(self):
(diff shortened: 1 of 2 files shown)


ashwindasr:
/hold

Until [test run](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-konflux/6451/console) finishes on jenkins

ashwindasr:
/close 

in favor of https://github.com/openshift-eng/art-tools/pull/1478

PR #1478, opened 2025-03-28
[konflux] use force for git add

ashwindasr:
`hypershift` image has a .gitignore file that needs to be present after rebase, in konflux. Usually its removed before rebase to make sure that files/folders that doozer generates are not excluded by git.

### DIFF
--- a/doozer/doozerlib/backend/build_repo.py
+++ b/doozer/doozerlib/backend/build_repo.py
@@ -165,7 +165,7 @@ class BuildRepo:
     async def commit(self, message: str, allow_empty: bool = False, force: bool = False):
         """ Commit changes in the local directory to the build source repository."""
         local_dir = str(self.local_dir)
-        await git_helper.run_git_async(["-C", local_dir, "."] + (["-f"] if force else []))
+        await git_helper.run_git_async(["-C", local_dir, "add", "."] + (["-f"] if force else []))
         commit_opts = []
         if allow_empty:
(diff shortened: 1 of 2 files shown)


ashwindasr:
/hold

Until a test run succeeds

doozer/doozerlib/backend/build_repo.py
```diff
@@ -162,10 +162,10 @@ async def _get_commit_hash(local_dir: str, strict: bool = False) -> Optional[str
             raise ChildProcessError(f"Failed to get commit hash: {err}")
         return out.strip()
 
-    async def commit(self, message: str, allow_empty: bool = False):
+    async def commit(self, message: str, allow_empty: bool = False, force: bool = False):
         """ Commit changes in the local directory to the build source repository."""
         local_dir = str(self.local_dir)
-        await git_helper.run_git_async(["-C", local_dir, "add", "."])
+        await git_helper.run_git_async(["-C", local_dir, f"{'-f' if force else ''}", "."])
```
jupierce, line 168: My brain reads this a passing an empty string parameter to a command line tool, which is a valid way for a callee to interpret the arguments, but not what is intended here. I'd just pull out the array into an args var and append `-f` if appropriate. If you want it inline, maybe something like:
```suggestion
        await git_helper.run_git_async(["-C", local_dir, "."] + (["-f"] if force else []))
```

ashwindasr:
Merging the MR since it is a gap on our side

PR #1479, opened 2025-03-31
build-fbc: Push to a dedicated quay repo

vfreex:
Due to a limitation in Konflux, FBC artifacts must be pushed to a public quay repo (instead of the private art-images repo). Making this change shouldn't cause any security concerns since FBC only includes metadata.


ashwindasr:
/hold

Would need to add image push secret to Jenkins first.

vfreex:
/hold cancel
Doozer doesn't directly push to that repo in Doozer. Instead, Konflux pipeline should be configured to use that pull secret.

PR #1480, opened 2025-03-31
[konflux] append .0 for go version

ashwindasr:
go versions defined like https://github.com/openshift/prometheus-alertmanager/blob/release-4.18/go.mod#L3 is breaking konflux prefetch task. i.e. the go version has to be defined as `1.22.0` instead of `1.22`.

Since there are 20+ images with this issue, for more than one OCP version, updating our automation to fix it.

Successfully rebased file: https://github.com/openshift-priv/image-customization-controller/blob/art-openshift-4.18-assembly-stream-dgk-ose-image-customization-controller/go.mod#L3
Successful prefetch: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-konflux/6683/

### DIFF
--- a/doozer/doozerlib/backend/rebaser.py
+++ b/doozer/doozerlib/backend/rebaser.py
@@ -427,6 +427,29 @@ class KonfluxRebaser:
             source_dockerfile_content = source_dockerfile.read()
             distgit_dockerfile.write(source_dockerfile_content)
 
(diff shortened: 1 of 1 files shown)


ashwindasr:
Created a better implementation with regex: https://github.com/openshift-eng/art-tools/pull/1523

PR #1483, opened 2025-03-31
Add awscli to pyartcd requirements

thegreyd:
pyartcd requires awscli to sync packages to mirror.
This will make it not depend on the global installation on buildvm.


joepvd (approved):
/lgtm

PR #1485, opened 2025-04-01
don't trigger rhcos build for 4.20

Ximinhan:
rhcos job don't have 4.20 right now and start from 4.19 rhcos pipeline will have changes, the old format probably won't suit, disable the trigger to avoid unnecessary failure for now


joepvd (approved):
/lgtm

PR #1488, opened 2025-04-02
[konflux] update UI url

ashwindasr:
ref slack [thread](https://redhat-internal.slack.com/archives/GDBRP5YJH/p1743616345840189)


rayfordj (commented):
/lgtm

PR #1492, opened 2025-04-03
[konflux] delete param | sast option

ashwindasr:
Changes in this PR:
- delete param option for PLR
- sast task disable option


doozer/doozerlib/backend/konflux_client.py
```diff
@@ -432,6 +447,12 @@ def _modify_param(params: List, name: str, value: Union[str, bool, list[str]]):
                 }]
             }]
 
+        if not sast:
+            params = self._delete_param(params, "sast-unicode-check")
+            params = self._delete_param(params, "sast-shell-check")
```
thegreyd, line 452: if we want to filter out these params, would recommend something like this:
```suggestion
        if not sast:
            sast_params = ("sast-unicode-check", "sast-shell-check")
            params = [p for p in params if p.get("name") not in sast_params]
```

doozer/doozerlib/backend/konflux_client.py
```diff
@@ -331,11 +331,26 @@ async def _get_pipelinerun_template(self, template_url: str):
...
+                continue
+            filtered_params.append(param)
+        return filtered_params
+
     async def _new_pipelinerun_for_image_build(self, generate_name: str, namespace: Optional[str], application_name: str, component_name: str,
                                                git_url: str, commit_sha: str, target_branch: str, output_image: str,
                                                build_platforms: Sequence[str], prefetch: Optional[list] = None, git_auth_secret: str = "pipelines-as-code-secret",
                                                additional_tags: Optional[Sequence[str]] = None, skip_checks: bool = False,
-                                               hermetic: Optional[bool] = None,
+                                               hermetic: Optional[bool] = None, sast: Optional[bool] = None,
```
thegreyd, line 353: Why not set it to False?

doozer/doozerlib/backend/konflux_client.py
```diff
@@ -447,6 +468,7 @@ async def start_pipeline_run_for_image_build(
         vm_override: dict,
         building_arches: Sequence[str],
         prefetch: Optional[list] = None,
+        sast: Optional[bool] = None,
```
thegreyd, line 471: Why not False?

doozer/doozerlib/backend/konflux_client.py
```diff
@@ -331,11 +331,26 @@ async def _get_pipelinerun_template(self, template_url: str):
                 template = jinja2.Template(template_text, autoescape=True)
                 return template
 
+    @staticmethod
+    def _delete_param(params: List, name: str, ):
```
thegreyd, line 335: I don't think this is required .. we shouldn't create a method for something small like this, which can be done via list comprehension

PR #1496, opened 2025-04-04
Sync only x86_64 for 4.12 and 4.13

adobes1:
We are producing only x86_64 for both 4.12 and 4.13. 
Fixes errors like [this](https://redhat-internal.slack.com/archives/C03JYDGPX8F/p1743723045279199?thread_ts=1743723039.998309&cid=C03JYDGPX8F)


ashwindasr (approved):

joepvd (approved):
/lgtm

PR #1502, opened 2025-04-07
[konflux] let doozer get-nightlies look at Konflux payloads

locriandev:
Tested with:
```
$ doozer --group=openshift-4.18 --assembly=stream --build-system=konflux get-nightlies
4.18.0-0.konflux-nightly-arm64-2025-04-09-095620 4.18.0-0.konflux-nightly-2025-04-09-093813
```

### DIFF
--- a/doozer/doozerlib/runtime.py
+++ b/doozer/doozerlib/runtime.py
@@ -4,7 +4,7 @@ from artcommonlib.assembly import AssemblyTypes, assembly_type, assembly_basis_e
     assembly_streams_config
 from artcommonlib.model import Model, Missing
 from artcommonlib.pushd import Dir
-from artcommonlib.util import isolate_el_version_in_brew_tag
+from artcommonlib.util import isolate_el_version_in_brew_tag, deep_merge
 from doozerlib.record_logger import RecordLogger
 from doozerlib.source_resolver import SourceResolver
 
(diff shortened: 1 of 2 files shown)


doozer/doozerlib/runtime.py
```diff
@@ -507,7 +507,10 @@ def filter_disabled(n, d):
             elif self.group_config.arches_override:  # Allow arches_override in group.yaml to temporarily override GA architectures
                 self.arches = self.group_config.arches_override
             else:
-                self.arches = self.group_config.get('arches', ['x86_64'])
+                if self.build_system == 'brew':
+                    self.arches = self.group_config.get('arches', ['x86_64'])
+                else:
+                    self.arches = self.group_config.get('konflux', {}).get('arches', ['x86_64'])
```
thegreyd, line 513: thinking if this is going to be a recurring pattern we can do something like this in runtime

```python
if self.build_system == "brew":
   group_config_without_konflux = {k:v for k, v in self.group_config if k != 'konflux' }
   # override konflux specific settings
   self.group_config = group_config_without_konflux.update(self.group_config.get('konflux', {}))
```

locriandev: Done [here](https://github.com/openshift-eng/art-tools/pull/1502/files#diff-1406051f1a04a744fb77f4c5faaac7f4b33f803c82506229b6e02bc7499e744fR254-R255)

PR #1504, opened 2025-04-07
elliott shipment init

thegreyd:
Output a shipment config, optionally based on an erratatool.yml advisory-key

```
elliott -g openshift-4.18 --assembly 4.18.3 shipment init --advisory-key microshift
```


vfreex (commented):
/lgtm

PR #1506, opened 2025-04-08
[konflux] fix sast bug

ashwindasr:
We need to remove the SAST tasks from `obj["spec"]["pipelineSpec"]["tasks"]` instead of `obj["spec"]["params"]`

### DIFF
--- a/doozer/doozerlib/backend/konflux_client.py
+++ b/doozer/doozerlib/backend/konflux_client.py
@@ -433,9 +433,15 @@ class KonfluxClient:
             }]
 
         if not sast:
-            params = [p for p in params if p.get("name") not in ("sast-unicode-check", "sast-shell-check")]
+            tasks = []
+            for task in obj["spec"]["pipelineSpec"]["tasks"]:
+                task_name = task.get("name")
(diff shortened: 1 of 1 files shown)


doozer/doozerlib/backend/konflux_client.py
```diff
@@ -433,9 +433,9 @@ def _modify_param(params: List, name: str, value: Union[str, bool, list[str]]):
             }]
 
         if not sast:
-            params = [p for p in params if p.get("name") not in ("sast-unicode-check", "sast-shell-check")]
-
-        obj["spec"]["params"] = params
+            self._logger.info("Removing sast-unicode-check and sast-shell-check tasks since sast is disabled")
+            tasks = [task for task in obj["spec"]["pipelineSpec"]["tasks"] if task.get("name") not in ("sast-unicode-check", "sast-shell-check")]
```
thegreyd, line 437: it would be good to know if something got filtered
we can compare the length of before and after and then log appropriately

vfreex (commented):
This looks fine although it might be more clear to explicitly mark those tasks as "skipped" instead of remove them:

```yaml
    when:
      - input: "true"
        operator: in
        values: ["false"]
```

PR #1508, opened 2025-04-09
Manually install golang 1.23.0

adobes1:
`check-payload` newly requires `go >= 1.23.0` which causes our art-cd builds to fail (such as [this](https://console-openshift-console.apps.artc2023.pc3z.p1.openshiftapps.com/k8s/ns/art-cd/builds/art-cd-update-802))

~~Manually install go 1.23.0 because this version of go is not yet available for ubi9. We can go back to installing it with `dnf` once it's available~~
Use the CI golang 1.23 builder image instead of ubi9 to get golang 1.23


joepvd (approved):
/lgtm

🎉

PR #1510, opened 2025-04-10
[konflux] implement gen-assembly

locriandev:
Test build: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/hack/job/dpaolell/job/gen-assembly/6/console

Needs https://github.com/openshift-eng/aos-cd-jobs/pull/4396


artcommon/artcommonlib/metadata.py
```diff
@@ -416,6 +417,9 @@ async def get_latest_konflux_build(self,
                           associated with the metadata.
         :param honor_is: If True, and an assembly component specifies 'is', that nvr will be returned.
         :param completed_before: cut off timestamp for builds completion time
+        :param engine: Even if we're looking at KonfluxDB, we might be interested in fetching records that belong
```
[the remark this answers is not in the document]
locriandev: This function will almost always fetch Konflux builds records, but at this point in time we're still building RPMs with Brew so giving it extra flexibility. This function gets normally called when `build_system == 'konflux'`, so that would be the default behavior as it sets
```
'engine': engine if engine else self.runtime.build_system,
```
and `runtime.build_system` is the primary switch. I would rather not rename this function, as its name will be an exact match once we start building RPMs with Brew.

doozer/doozerlib/cli/release_gen_assembly.py
```diff
@@ -259,17 +269,31 @@ async def _process_release(self, brew_cpu_arch, pullspec, rhcos_tag_names):
...
+            else:
+                image_envs = image_info['config']['config']['Env']
+                package_name = next((env.split('=')[1] for env in image_envs if env.startswith('__doozer_key=')))
+                build_nvr = package_name + '-' + image_labels['version'] + '-' + image_labels['release']
+                try:
+                    build_record = await self.runtime.konflux_db.get_build_record_by_nvr(
+                        nvr=build_nvr,
+                        outcome=KonfluxBuildOutcome.SUCCESS
+                    )
+                except:
```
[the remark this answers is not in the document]
locriandev: ah yeah that was a debugging facility. Removed it

doozer/doozerlib/cli/release_gen_assembly.py
```diff
@@ -351,16 +379,33 @@ def _collect_outliers(self):
...
+                                          f'at estimated brew event {self.basis_event}. No normal reason for this to '
+                                          f'happen so exiting out of caution.')
+
+                basis_event_build_dict: BrewBuildRecordInspector = BrewBuildRecordInspector(
+                    self.runtime, basis_event_dict['id'])
+                basis_event_build_nvr = basis_event_build_dict.get_nvr()
+
+            else:
+                package_name = dgk
+                build_record = await anext(self.runtime.konflux_db.search_builds_by_fields(
```
thegreyd, line 397: Can we use `meta.get_latest_konflux_build` here? is it `konflux_event` aware?

doozer/doozerlib/cli/release_gen_assembly.py
```diff
@@ -620,7 +677,7 @@ def _generate_assembly_definition(self) -> dict:
                     "assembly": {
                         'type': self.assembly_type.value,
                         'basis': {
-                            'brew_event': self.basis_event,
+                            f'{self.runtime.build_system}_event': self.basis_event,
```
thegreyd, line 680: Hmm not a big fan of `konflux_event` since this is just the epoch we're calculating via our own DB records. We can in-theory do the same thing for brew builds (it maybe less accurate?). So maybe something more generic? `cutoff_time` , also since `event` here means something very specific (monotonically increasing non-overlapping int counter in brew)

doozer/doozerlib/cli/release_gen_assembly.py
locriandev: Yes! Done it with 8f5e4b8f76fbbb13dd4f8b9473023c31f83cc1ac :+1: 

Note that konflux event is not a BigQuery concept. We store timestamps, but in order for us to reuse most of gen-assembly existing logic I had to introduce this timestamp-to-integer conversion.

doozer/doozerlib/cli/release_gen_assembly.py
locriandev: Not sure how `cutoff_time` would be appropriate. It's true that we store a Unix epoch, but I would expect a timestamp when reading `time`. In this case I just chose the easiest path to get rid of `brew` names in Konflux assemblies. I do agree that `event` might be misleading as well. `latest_build_epoch`? `cutoff_epoch_ms`? `cutoff_timestamp_ms`? Not a big name-finding guru :laughing:

thegreyd: Hmm yeah. How about `cutoff_event`? `event` is vague enough

artcommon/artcommonlib/metadata.py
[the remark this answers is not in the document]
thegreyd: imo "konflux db / konflux format" is confusing since it has nothing to do with Konflux itself, but everything to do with our internal database - which we can use to store konflux/brew builds or any more kind of data. baking in assumptions like these are bad for readability and understanding

vfreex (commented):
/lgtm

PR #1517, opened 2025-04-12
Revert "Use p2/p3 flags for Konflux releases"

joepvd:
Reverts openshift-eng/art-tools#1473


joepvd:
This has some [issues](https://redhat-internal.slack.com/archives/CJARLA942/p1744426299267849), reverting.

PR #1521, opened 2025-04-14
Bring back Slack reporting to quay-doomsday-backup

adobes1:
This PR brings back Slack reporting logic of the `quay-doomsday-backup` pipeline which was removed in https://github.com/openshift-eng/art-tools/pull/1516.

To achieve this without major changes, this PR makes the functions async again and nearly identical to how they were before https://github.com/openshift-eng/art-tools/pull/1516. The only difference are [these lines](https://github.com/adobes1/art-tools/blob/73f047be1dafca95d1a61e552cce45c3f3f6f279/pyartcd/pyartcd/pipelines/quay_doomsday_backup.py#L93-L96) which run the synchronization process for each arch in sequence - this is done to hopefully fix Quay returning 502 when calling `oc adm release mirror ...` ([example](https://console-openshift-console.apps.artc2023.pc3z.p1.openshiftapps.com/k8s/ns/art-cd/tekton.dev~v1~PipelineRun/doomsday-pipeline-run-nbpjk/logs))

Below is the diff between the [previous async version](https://github.com/openshift-eng/art-tools/blob/08c08bf59958382dbd589ddc244c5dbac1d88570/pyartcd/pyartcd/pipelines/quay_doomsday_backup.py) and the one in this PR:
```
diff --git a/pyartcd/pyartcd/pipelines/quay_doomsday_backup.py b/pyartcd/pyartcd/pipelines/quay_doomsday_backup.py
index 3e80b2de..2f7b3a2f 100644
--- a/pyartcd/pyartcd/pipelines/quay_doomsday_backup.py
+++ b/pyartcd/pyartcd/pipelines/quay_doomsday_backup.py
@@ -90,8 +90,10 @@ class QuayDoomsdaySync:
         else:
             self.runtime.logger.info("[DRY RUN] Would have messaged Slack")
 
-        tasks = [self.sync_arch(arch) for arch in self.arches]
-        results = await asyncio.gather(*tasks)
+        # Synchronize individual arches sequentially to help with quay returning 502
+        results = []
+        for arch in self.arches:
+            results.append(await self.sync_arch(arch))
 
         # Report the results to Slack
         if not self.runtime.dry_run:
```

### DIFF
--- a/pyartcd/pyartcd/pipelines/quay_doomsday_backup.py
+++ b/pyartcd/pyartcd/pipelines/quay_doomsday_backup.py
@@ -1,12 +1,13 @@
 import os
 from typing import Optional
 import click
-from time import sleep
 import shutil
+import asyncio
+from tenacity import AsyncRetrying, stop_after_attempt
 
 from pyartcd.runtime import Runtime
-from pyartcd.cli import cli, pass_runtime
-from artcommonlib.exectools import cmd_assert
+from pyartcd.cli import cli, pass_runtime, click_coroutine
+from artcommonlib.exectools import cmd_assert_async
 from doozerlib.util import mkdirs
 
 
@@ -23,10 +24,12 @@ class QuayDoomsdaySync:
         self.runtime = runtime
         self.version = version
         self.workdir = "./workspace"
+        self.slack_client = self.runtime.new_slack_client()
+        self.slack_client.bind_channel(version)
 
         self.arches = arches.split(",") if arches else ALL_ARCHES_LIST
 
-    def sync_arch(self, arch: str):
+    async def sync_arch(self, arch: str) -> bool:
         if arch not in ALL_ARCHES_LIST:
             raise Exception(f"Invalid arch: {arch}")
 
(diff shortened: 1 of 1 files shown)


pyartcd/pyartcd/pipelines/quay_doomsday_backup.py
```diff
@@ -44,39 +47,70 @@ def sync_arch(self, arch: str):
...
+            main_message_ts = slack_response["message"]["ts"]
+        else:
+            self.runtime.logger.info("[DRY RUN] Would have messaged Slack")
+
+        # Synchronize individual arches sequentially to help with quay returning 502
+        results = []
         for arch in self.arches:
-            self.runtime.logger.info("Now syncing arch %s", arch)
-            self.sync_arch(arch)
+            results.append(await self.sync_arch(arch))
```
vfreex, line 96: Here per-arch `sync_arch` calls still run sequentially. To run them in parallel, `asyncio.gather` can be used.

adobes1: This is actually intentional with the hope that it could potentially help with Quay returning 502 Bad Gateway (such as in [this PipelineRun](https://console-openshift-console.apps.artc2023.pc3z.p1.openshiftapps.com/k8s/ns/art-cd/tekton.dev~v1~PipelineRun/doomsday-pipeline-run-nbpjk/logs)). I probably could've explained this better in the PR description 😅

vfreex: ok, that makes sense. thanks

PR #1523, opened 2025-04-15
[konflux] append .0 for go version

ashwindasr:
go versions defined like https://github.com/openshift/prometheus-alertmanager/blob/release-4.18/go.mod#L3 is breaking konflux prefetch task. i.e. the go version has to be defined as `1.22.0` instead of `1.22`.

Since there are 20+ images with this issue, for more than one OCP version, updating our automation to fix it.

Successful rebase: https://github.com/openshift-priv/image-customization-controller/commit/2db4984435298ee339696f1cc122517b5b5b52e3#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L3

Custom gomod path support will be added in a follow up PR.

### DIFF
--- a/doozer/doozerlib/backend/rebaser.py
+++ b/doozer/doozerlib/backend/rebaser.py
@@ -428,6 +428,26 @@ class KonfluxRebaser:
             source_dockerfile_content = source_dockerfile.read()
             distgit_dockerfile.write(source_dockerfile_content)
 
+        gomod_path = dest_dir.joinpath('go.mod')
+        if gomod_path.exists():
+            # Read the gomod contents
+            new_lines = []
+            with open(gomod_path, "r") as file:
+                lines = file.readlines()
+                for line in lines:
+                    stripped_line = line.strip()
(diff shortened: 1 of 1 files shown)


doozer/doozerlib/backend/rebaser.py
```diff
@@ -428,6 +428,26 @@ class MergeSourceResult:
...
 
+        gomod_path = dest_dir.joinpath('go.mod')
+        if gomod_path.exists():
+            # Read the gomod contents
+            new_lines = []
+            with open(gomod_path, "r") as file:
+                lines = file.readlines()
+                for line in lines:
+                    stripped_line = line.strip()
+                    match = re.match(r"(^go \d\.\d+$)", stripped_line)
```
jupierce, line 439: I'd permit whitespace+ before and after the number, but unlikely to be important.

ashwindasr: thanks - when I checked, I didn't see a case like that. If I do, will update the regex

PR #1498, opened 2025-04-07
ART-12288: Support new rhcos node image and extensions

Ximinhan:
use new function to check latest node image by digest
use new function to get rpms from node image and extensions 
trigger rhcos node image job for 4.19

test build-sync: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fbuild-sync/17324/console
```
2025-04-07 09:55:12,560 artcommonlib INFO Getting latest RHCOS source for x86_64...
2025-04-07 09:55:12,563 art_tools.artcommonlib.exectools INFO $3271: ['oc', 'image', 'info', 'quay.io/openshift-release-dev/ocp-v4.0-art-dev:4.19-9.6-node-image', '--filter-by-os', 'amd64', '-o', 'json'] - [cwd=None]: Executing:cmd_gather: oc image info quay.io/openshift-release-dev/ocp-v4.0-art-dev:4.19-9.6-node-image --filter-by-os amd64 -o json
2025-04-07 09:55:14,749 art_tools.artcommonlib.exectools INFO $3272: ['oc', 'image', 'info', 'quay.io/openshift-release-dev/ocp-v4.0-art-dev:4.19-9.6-node-image', '--filter-by-os', 'amd64', '-o', 'json'] - [cwd=None]: Executing:cmd_gather: oc image info quay.io/openshift-release-dev/ocp-v4.0-art-dev:4.19-9.6-node-image --filter-by-os amd64 -o json
2025-04-07 09:55:15,916 art_tools.artcommonlib.exectools INFO $3273: ['oc', 'image', 'info', 'quay.io/openshift-release-dev/ocp-v4.0-art-dev:4.19-9.6-node-image-extensions', '--filter-by-os', 'amd64', '-o', 'json'] - [cwd=None]: Executing:cmd_gather: oc image info quay.io/openshift-release-dev/ocp-v4.0-art-dev:4.19-9.6-node-image-extensions --filter-by-os amd64 -o json
```

test ocp4_scan job: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4_scan/117895/consoleFull
```
2025-04-08 05:31:04,852 artcommonlib INFO Getting latest RHCOS source for aarch64...
[...]

PR #1537, opened 2025-04-22
Revert "ART-12288: Support new rhcos node image and extensions"

locriandev:
Reverts openshift-eng/art-tools#1498


vfreex:
/lgtm
/approve

@ximhan We need to temporarily revert this to unblock ec releases. Could you have a look at https://redhat-internal.slack.com/archives/GDBRP5YJH/p1745327071535269?thread_ts=1745314562.772319&cid=GDBRP5YJH when you get time?

PR #1549, opened 2025-04-28
ART-12599: Add Doomsday PipelineRun URL to the Slack message

adobes1:
Include link to the doomsday PipelineRun (such as [this](https://console-openshift-console.apps.artc2023.pc3z.p1.openshiftapps.com/k8s/ns/art-cd/tekton.dev~v1~PipelineRun/doomsday-pipeline-run-ghscf)) in the message posted in version specific Slack channel.

[Test run triggered from the cluster](https://redhat-internal.slack.com/archives/C05HB1F1THP/p1745844135783889) (modified to only post the message without syncing anything)

### DIFF
--- a/art-cluster/pipelines/config/argocd/project/art-cd/quay-doomsday-backup/base/doomsday-pipeline-task.yaml
+++ b/art-cluster/pipelines/config/argocd/project/art-cd/quay-doomsday-backup/base/doomsday-pipeline-task.yaml
@@ -10,6 +10,9 @@ spec:
     - description: 'Full version, eg: 4.15.5'
       name: version
       type: string
+    - description: 'URL of the PipelineRun'
+      name: pipelinerun_url
+      type: string
   steps:
     - env:
         - name: DOCKER_CONFIG
@@ -19,6 +22,8 @@ spec:
             secretKeyRef:
               key: api-token
               name: slack-bot-creds
+        - name: PIPELINERUN_URL
+          value: $(params.pipelinerun_url)
       image: default-route-openshift-image-registry.apps.artc2023.pc3z.p1.openshiftapps.com/art-cd/art-cd:latest
       name: run-script
       resources: {}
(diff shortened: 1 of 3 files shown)


art-cluster/pipelines/config/argocd/project/art-cd/quay-doomsday-backup/base/doomsday-pipeline-task.yaml
```diff
@@ -10,6 +10,9 @@ spec:
     - description: 'Full version, eg: 4.15.5'
       name: version
       type: string
+    - description: 'URL of the PipelineRun'
```
ashwindasr, line 13: Passing the PipelineRun URL to the job does not modify its functionality. I do not see any case where we have to pass a different pipelinerun URL than which can be automatically created. Which means that it need not be a param here.

pyartcd/pyartcd/pipelines/quay_doomsday_backup.py
```diff
@@ -21,6 +21,7 @@ class QuayDoomsdaySync:
     """
 
     def __init__(self, runtime: Runtime, version: str, arches: Optional[str]):
+        self.pipelinerun_url = os.environ.get("PIPELINERUN_URL")
```
ashwindasr, line 24: Generate the PipelineRun URL in the code itself here

ashwindasr (changes requested):

pyartcd/pyartcd/pipelines/quay_doomsday_backup.py
adobes1: I have to admit that I have no idea how to get the URL directly inside the code 😅 The reason I did it using the Pipeline params is because in there I have access to `$(context.pipelineRun.name)` which is not even available inside Task. Could you point me in the right direction?

adobes1: Nevermind, I forgot there's `oc` and `tkn` in the art-cd image 😅. We should be able to get the PipelineRun name using either of these two.

PR #1551, opened 2025-04-28
Fix modification error

ashwindasr:
The build system doesn't seem to be available always, default to brew if its missing


joepvd (approved):
/lgtm

PR #1553, opened 2025-04-28
Revert "Construct rpm from purl"

ashwindasr:
Reverts openshift-eng/art-tools#1542


ashwindasr:
/hold

Only for testing

PR #1556, opened 2025-04-29
[konflux] gen-payload for named assemblies

locriandev:
Test build: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/hack/job/dpaolell/job/build-sync-konflux/7/console

Create imagestreams:
- `4.18-konflux-art-assembly-4.18.100`
- `4.18-konflux-art-assembly-4.18.100-multi`
- `4.18-konflux-art-assembly-4.18.100-s390x`
- `4.18-konflux-art-assembly-4.18.100-ppc64le`
- `4.18-konflux-art-assembly-4.18.100-arm64`

### DIFF
--- a/artcommon/artcommonlib/assembly.py
+++ b/artcommon/artcommonlib/assembly.py
@@ -1,4 +1,5 @@
 import copy
+from datetime import datetime
 from enum import Enum
 
 import typing
@@ -273,7 +274,9 @@ def _assembly_field(field_name: str, releases_config: Model, assembly: str) -> M
     return Model(dict_to_model=config_dict)
 
 
-def assembly_basis_event(releases_config: Model, assembly: str, strict: bool = False) -> typing.Optional[int]:
+def assembly_basis_event(
+    releases_config: Model, assembly: str, strict: bool = False
+) -> typing.Optional[typing.Union[int, datetime]]:
     """
     :param releases_config: The content of releases.yml in Model form.
     :param assembly: The name of the assembly to assess
@@ -288,8 +291,11 @@ def assembly_basis_event(releases_config: Model, assembly: str, strict: bool = F
 
     _check_recursion(releases_config, assembly)
     target_assembly = releases_config.releases[assembly].assembly
+
     if target_assembly.basis.brew_event:
-        return int(target_assembly.basis.brew_event)
+        return int(target_assembly.basis.brew_event)  # Integer for Brew event
+    elif target_assembly.basis.time:
+        return target_assembly.basis.time  # Datetime for Konflux
     return assembly_basis_event(releases_config, target_assembly.basis.assembly, strict=strict)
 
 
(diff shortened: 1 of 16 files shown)


artcommon/artcommonlib/assembly.py
```diff
@@ -269,7 +270,7 @@ def _assembly_field(field_name: str, releases_config: Model, assembly: str) -> M
     return Model(dict_to_model=config_dict)
 
 
-def assembly_basis_event(releases_config: Model, assembly: str, strict: bool = False) -> typing.Optional[int]:
+def assembly_basis_event(releases_config: Model, assembly: str, strict: bool = False) -> typing.Optional[typing.Union[int, datetime]]:
```
thegreyd, line 273: hmm why not a new method for this, so we know for sure which field is referenced and what type will be returned ?

artcommon/artcommonlib/konflux/package_rpm_finder.py
```diff
@@ -86,7 +86,7 @@ def get_installed_packages_build_dicts(self, build_record: KonfluxBuildRecord) -
         installed_packages = [pkg for pkg in installed_packages if pkg not in self._not_found_packages]
         self._cache_packages(installed_packages)
         return {self._packages_build_dicts[package_nvr]['name']: self._packages_build_dicts[package_nvr]
-                for package_nvr in installed_packages}
+                for package_nvr in installed_packages if package_nvr not in self._not_found_packages}
```
thegreyd, line 89: I think the if condition is repeated here? (already checked at line 86?)

thegreyd, line 89: I think the if condition is repeated here? (already checked at line 86?)

artcommon/artcommonlib/assembly.py
locriandev: Why bother? We've always chosen to make code flexible instead of duplicating logic across build systems

doozer/doozerlib/util.py
```diff
@@ -507,6 +507,28 @@ def get_release_name_for_assembly(group_name: str, releases_config: Model, assem
     return get_release_name(assembly_type, group_name, assembly_name, patch_version)
 
 
+def get_registry_basic_auth(pullspec: str, registry_password: str, registry_username: str = None) -> str:
```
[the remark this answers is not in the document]
locriandev: This was introduced a while ago, and I think was meant by design. It is a lot easier to update a username or a password in Jenkins than a base64 encoded auth file, which tells us nothing about what we're currently using. Not against discussing this though

artcommon/artcommonlib/konflux/package_rpm_finder.py
locriandev: I think it's intended. First `installed_packages` is subtracted the packages we already know are not found. Then, we cache the rest. There can be more that are not mapped in Brew, so this condition again ensures we're not raising a KeyError by trying to index a package that could not be mapped.

PR #1557, opened 2025-04-29
konflux SA override update

ashwindasr:
Follow up to https://github.com/openshift-eng/art-tools/pull/1547

https://issues.redhat.com/browse/KONFLUX-7956 has been setup i.e. which means that we need to use the per component SA, which is updated as part of the PR


ashwindasr:
/hold

Until test build succeeds: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-konflux/8373/console

ashwindasr:
/unhold

Test build succeeded

PR #1565, opened 2025-04-30
Bump retries for multi payload creation

joepvd:
Quay throws occasionally more 502s than current retry can handle


locriandev (commented):
/lgtm

PR #1570, opened 2025-04-30
check SLSA attestation before storing build

ashwindasr:


doozer/doozerlib/backend/konflux_image_builder.py
```diff
@@ -187,7 +187,7 @@ async def build(self, metadata: ImageMetadata):
                 if self._config.dry_run:
                     logger.info("Dry run: Would have inserted build record in Konflux DB")
                 else:
-                    await self.update_konflux_db(metadata, build_repo, pipelinerun, outcome, building_arches, pod_list)
+                    _, outcome = await self.update_konflux_db(metadata, build_repo, pipelinerun, outcome, building_arches, pod_list)
```
thegreyd, line 190: I don't think we should have the gating as part of the update_konflux_db func. It would be better to do it after line 185, i.e. when build has finished.

artcommon/artcommonlib/util.py
```diff
@@ -430,3 +430,15 @@ def detect_package_managers(metadata, dest_dir: Path):
         if any(dest_dir.joinpath(file).is_file() for file in files):
             pkg_managers.append(pkg_manager)
     return pkg_managers
+
+
+@retry(reraise=True, wait=wait_fixed(10), stop=stop_after_attempt(3))
+async def get_konflux_slsa_attestation(pull_spec: str):
+    """
+    Inspect the SLSA provenance: https://konflux.pages.redhat.com/docs/users/metadata/attestations.html
+    To see if the latest build is actually hermetic
```
thegreyd, line 439: nit: think this comment can be removed, since we're just returning the attestation and func caller can decide what to do with it

artcommon/artcommonlib/util.py
```diff
@@ -430,3 +430,15 @@ def detect_package_managers(metadata, dest_dir: Path):
...
+    """
+    cmd = f"cosign download attestation {pull_spec} --registry-username {os.environ['KONFLUX_ART_IMAGES_USERNAME']} --registry-password {os.environ['KONFLUX_ART_IMAGES_PASSWORD']}"
```
thegreyd, line 441: would recommend readin env vars at the beginning of cli script and then passing them along as normal func params to this func.

doozer/doozerlib/cli/scan_sources_konflux.py
```diff
@@ -417,18 +417,10 @@ async def scan_network_mode_changes(self, image_meta: ImageMetadata):
...
-            )
-
+        attestation = await artcommonlib.util.get_konflux_slsa_attestation(build_record.image_pullspec)
         try:
             # Equivalent bash code: jq -r ' .payload | @base64d | fromjson | .predicate.invocation.parameters.hermetic'
-            payload_json = json.loads(base64.b64decode(json.loads(attestation.strip())["payload"]).decode("utf-8"))
+            payload_json = json.loads(base64.b64decode(json.loads(attestation)["payload"]).decode("utf-8"))
         except Exception as e:
             raise IOError(f"Failed to parse SLSA attestation for {build_record.image_pullspec}: {e}")
         is_hermetic = payload_json["predicate"]["invocation"]["parameters"]["hermetic"]
```
thegreyd, line 426: if this is supposed to be a bool param, then maybe we should convert it to bool (instead of str for comparison later on) - and if it is not a bool, complain!

thegreyd:
Regarding config changes and rebuilding: we have digest calculation and keys defined here: https://github.com/openshift-eng/art-tools/blob/main/doozer/doozerlib/image.py#L504 , now as konflux and brew keys are differing, it probably makes sense to have it be factored in digest calculation

doozer/doozerlib/backend/konflux_image_builder.py
ashwindasr: makes sense - updated in https://github.com/openshift-eng/art-tools/pull/1570/commits/e7f988d9c6288eea1bebb7772a03797f100cead3

artcommon/artcommonlib/util.py
ashwindasr: updated in https://github.com/openshift-eng/art-tools/pull/1570/commits/e7f988d9c6288eea1bebb7772a03797f100cead3

artcommon/artcommonlib/util.py
ashwindasr: Not sure where you mean beginning at the cli script? Do you mean as part of the `click` params?

doozer/doozerlib/cli/scan_sources_konflux.py
ashwindasr: its a bool param but its retrieved as a string, I've made an update in https://github.com/openshift-eng/art-tools/pull/1570/commits/e7f988d9c6288eea1bebb7772a03797f100cead3

```
is_hermetic = True if is_hermetic.lower() == "true" else False
```
if that works

ashwindasr:
> now as konflux and brew keys are differing, it probably makes sense to have it be factored in digest calculation
We are already merging image level konflux keys: https://github.com/openshift-eng/art-tools/blob/main/doozer/doozerlib/image.py#L509. But we haven't for group config keys. Are you referring to that?

ashwindasr:
/hold

Until a new test build is triggered

doozer/doozerlib/cli/scan_sources_konflux.py
thegreyd: yup! that definitely reads better

artcommon/artcommonlib/util.py
thegreyd: For example - https://github.com/openshift-eng/art-tools/blob/543b5a99dbb0e8daf2bc9cf92c45988cf5233735/elliott/elliottlib/cli/konflux_release_cli.py#L259

doozer/doozerlib/backend/konflux_image_builder.py
```diff
@@ -184,6 +184,16 @@ async def build(self, metadata: ImageMetadata):
                 succeeded_condition = artlib_util.KubeCondition.find_condition(pipelinerun, 'Succeeded')
                 outcome = KonfluxBuildOutcome.extract_from_pipelinerun_succeeded_condition(succeeded_condition)
 
+                # Even if the build succeeded, if the SLSA attestation cannot be retrieved, it is unreleasable.
+                if outcome is KonfluxBuildOutcome.SUCCESS:
+                    image_pullspec = next((r['value'] for r in pipelinerun.status.results if r['name'] == 'IMAGE_URL'), None)
+                    try:
+                        # Get SLA attestation from konflux. The command will error out if it cannot find it.
+                        await artlib_util.get_konflux_slsa_attestation(image_pullspec, os.environ['KONFLUX_ART_IMAGES_USERNAME'], os.environ['KONFLUX_ART_IMAGES_PASSWORD'])
+                    except Exception as e:
```
thegreyd, line 193: Hm I don't think it's a good practice to rely on a generic exception being raised. Instead if you have an http error code or message which indicates that the attestation is missing and then return None from the func, that will be a better indicator

thegreyd: ah since it's a cmd, then rc != 0 indicates that there was a failure, which should be sufficient

doozer/doozerlib/backend/konflux_image_builder.py
```diff
@@ -184,6 +184,16 @@ async def build(self, metadata: ImageMetadata):
...
+                        logger.error(f"Failed to get SLA attestation from konflux for image {image_pullspec}: {e}")
```
thegreyd, line 194:
```suggestion
                        logger.error(f"Failed to get SLA attestation from konflux for image {image_pullspec}, marking build as {KonfluxBuildOutcome.FAILURE}. Error: {e}")
```

doozer/doozerlib/backend/konflux_image_builder.py
ashwindasr: `cmd_gather_async` inside `get_konflux_slsa_attestation`  will raise a `ChildProcessError` if the command fails. So I can update `Exception` to that, but does it matter here since we are logging the error and explaining the reason?

thegreyd: Not if you have check=False, then you can rely on rc

thegreyd: exceptions can be raised for many reasons, for example if there is a command syntax error or even a python error in method. so imo we should look for the exact fail reason (as exact as possible)

artcommon/artcommonlib/util.py
ashwindasr: thanks - updated in https://github.com/openshift-eng/art-tools/pull/1570/commits/83dc5bb61da5145df894dfc287fca0dbe191aef9

so the goal is try to get the variable in the beginning and fail fast if it doesn't

doozer/doozerlib/backend/konflux_image_builder.py
ashwindasr: ack - updated in https://github.com/openshift-eng/art-tools/pull/1570/commits/0d3670c9dd295b604d2ac718e698dcf95f0b59ad

PR #1571, opened 2025-05-01
Add ruff for formatting

thegreyd:
https://docs.astral.sh/ruff/ is a fast drop-in replacement for
- flake8 and more lint checks (`ruff check`)
- black & isort (`ruff format`)

This PR introduces `ruff format` for auto-formatting. No functional code changes are expected

Future PR will cover linting w. `ruff check`

Formatting style:
- ruff's format style is aligned to black https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html
- I've added ruff.toml to configure
  - max-line-length to 120
  - quote-style to "preserve"  

black/ruff don't allow us to change how it wraps multiple lines into 1 if it is possible. This results in quite a lot of undesirable changes. To get around this the official guidance is to add a trailing comma. So ruff formats as follows

```diff
-a = [
-  1,
-  2
-]
+a = [1, 2]

b = [
  3,
  4,
]
```

https://black.vercel.app/ to see how formatting works.

To add trailing comma in our repo at scale, I used a tweaked version of https://github.com/asottile/add-trailing-comma so that it only operates on multi-line literals.

```bash
$ uv pip install "git+https://github.com/thegreyd/add-trailing-comma.git@dev"
$ git ls-files -z '*.py' | xargs -0 uv run -m add_trailing_comma

$ make format  # run ruff format
```

Follow up to avoid ruining git blame: https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html#avoiding-ruining-git-blame


doozer/doozerlib/build_info.py
```diff
@@ -579,7 +605,14 @@ def is_under_embargo(self) -> bool:
         if cn not in self._cache:
             with self.runtime.shared_build_status_detector() as bs_detector:
                 # determine if the image build is embargoed (or otherwise "private")
-                self._cache[cn] = len(bs_detector.find_embargoed_builds([self._brew_build_obj], [self.get_image_meta().candidate_brew_tag()])) > 0
+                self._cache[cn] = (
+                    len(
```
locriandev, line 609: Debatable formatting choice :sweat_smile: Do you think we have room for customization here?

doozer/doozerlib/backend/konflux_olm_bundler.py
```diff
@@ -343,10 +386,7 @@ def _create_dockerfile(self, metadata: ImageMetadata, operator_dir: Path, bundle
         bundle_df.labels['com.redhat.component'] = metadata.get_olm_bundle_brew_component_name()
         bundle_df.labels['com.redhat.delivery.appregistry'] = False
         bundle_df.labels['name'] = metadata.get_olm_bundle_image_name()
-        bundle_df.labels['version'] = '{}.{}'.format(
-            operator_df.labels['version'],
-            operator_df.labels['release']
-        )
+        bundle_df.labels['version'] = '{}.{}'.format(operator_df.labels['version'], operator_df.labels['release'])
```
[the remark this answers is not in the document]
thegreyd: Yes! we have a lot of instances of .format from before f-string .. don't want to tackle them in this PR though

doozer/doozerlib/build_info.py
thegreyd: Yeah this is the case of the formatter trying to fix someting that was a poor readability decision in the first place - nested calls. imo it is okay to have these monstrosities, so that when we see one, we fix it by using vars like a sane person on a good day :D

locriandev (commented):
/lgtm

PR #1573, opened 2025-05-02
ART-12761: docs workflow update

ashwindasr:
For https://github.com/openshift-eng/ocp-build-data/pull/6706

In this PR:
- Store the template thats common to all OCP versions in openshift-eng/ocp-build-data#main
- The config can be overridden in the group level, if necessary
- As a first step, use the updated template. Follow up features include (updating the description to include the advisory id, adding the SHAs for the release digest)

### DIFF
--- a/elliott/elliottlib/cli/attach_cve_flaws_cli.py
+++ b/elliott/elliottlib/cli/attach_cve_flaws_cli.py
@@ -154,7 +154,7 @@ def _update_advisory(runtime, advisory, flaw_bugs, bug_tracker, noop):
 
     # Get the advisory name, eg. image|rpm|metadata|microshift
     releases_config = runtime.get_releases_config()
-    advisories = releases_config[runtime.assembly]['assembly']['group']['advisories']
+    advisories = releases_config["releases"][runtime.assembly]['assembly']['group']['advisories']
     art_advisory_key = next((k for k, v in advisories.items() if v == advisory_id), None)
 
     if not art_advisory_key:
--- a/elliott/elliottlib/cli/attach_cve_flaws_cli.py
+++ b/elliott/elliottlib/cli/attach_cve_flaws_cli.py
@@ -155,11 +155,7 @@ def _update_advisory(runtime, advisory, flaw_bugs, bug_tracker, noop):
     # Get the advisory name, eg. image|rpm|metadata|microshift
     releases_config = runtime.get_releases_config()
     advisories = releases_config[runtime.assembly]['assembly']['group']['advisories']
-    art_advisory_key = None
-    for _name, _id in advisories.items():
-        if _id == advisory_id:
-            art_advisory_key = _name
-            break
+    art_advisory_key = next((k for k, v in advisories.items() if v == advisory_id), None)
 
     if not art_advisory_key:
         raise ValueError(f'ART advisory key not found for advisory: {advisory_id} in list {advisories.items()}')
--- a/elliott/elliottlib/cli/attach_cve_flaws_cli.py
+++ b/elliott/elliottlib/cli/attach_cve_flaws_cli.py
@@ -67,7 +67,7 @@ async def attach_cve_flaws_cli(
 
     # Get the advisory kind-id mapping
     releases_config = runtime.get_releases_config()
-    advisories_by_kind = releases_config["releases"][runtime.assembly]['assembly']['group']['advisories']
+    advisories_by_kind = releases_config['releases'][runtime.assembly]['assembly']['group']['advisories']
 
     exit_code = 0
     flaw_bug_tracker = runtime.get_bug_tracker('bugzilla')
--- a/elliott/elliottlib/cli/attach_cve_flaws_cli.py
+++ b/elliott/elliottlib/cli/attach_cve_flaws_cli.py
@@ -64,6 +64,11 @@ async def attach_cve_flaws_cli(
         advisories = [find_default_advisory(runtime, default_advisory_type)]
     else:
         advisories = [advisory_id]
+
+    # Get the advisory kind-id mapping
+    releases_config = runtime.get_releases_config()
+    advisories_by_kind = releases_config["releases"][runtime.assembly]['assembly']['group']['advisories']
+
     exit_code = 0
     flaw_bug_tracker = runtime.get_bug_tracker('bugzilla')
     errata_config = runtime.get_errata_config()
@@ -81,7 +86,7 @@ async def attach_cve_flaws_cli(
 
         try:
             if flaw_bugs:
-                _update_advisory(runtime, advisory, flaw_bugs, flaw_bug_tracker, noop)
+                _update_advisory(runtime, advisory, advisories_by_kind, flaw_bugs, flaw_bug_tracker, noop)
                 # Associate builds with CVEs
                 LOGGER.info('Associating CVEs with builds')
                 await associate_builds_with_cves(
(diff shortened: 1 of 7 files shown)


ashwindasr:
/hold

Until peer review and test advisory created.

elliott/elliottlib/cli/create_cli.py
```diff
@@ -133,25 +136,42 @@ async def create_cli(
...
-    if "boilerplates" not in et_data:
-        raise ValueError("`boilerplates` is required in erratatool.yml")
+    common_advisory_template = yaml.safe_load(
+        requests.get(
+            "https://raw.githubusercontent.com/openshift-eng/ocp-build-data/refs/heads/docs-update/release_notes.yml"
+        )
+    )
+
+    common_boilerplate = common_advisory_template.get("boilerplates", {}).get(art_advisory_key)
+    group_boilerplate = et_data.get("boilerplates", {}).get(art_advisory_key)
```
joepvd, line 146: Think we need an override, but not sure if `group` is the right granularity. Think we run the risk of keeping that in group.yml. Is it possible/desirable to have only a assembly based override to not get tempted for durable fixes in the wrong place?

ashwindasr: By group, I was referring to the erratatool.yml file found in the "group" level, i.e. OCP version: https://github.com/openshift-eng/ocp-build-data/blob/openshift-4.19/erratatool.yml

elliott/elliottlib/cli/create_cli.py
```diff
@@ -133,25 +136,42 @@ async def create_cli(
     if sum(map(bool, [date, batch_id])) != 1:
         raise click.BadParameter("Need either --date or --batch-id")
 
-    if "boilerplates" not in et_data:
-        raise ValueError("`boilerplates` is required in erratatool.yml")
+    common_advisory_template = yaml.safe_load(
+        requests.get(
+            "https://raw.githubusercontent.com/openshift-eng/ocp-build-data/refs/heads/docs-update/release_notes.yml"
```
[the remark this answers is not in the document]
ashwindasr: From what I knew, content would be updated every 5 mins, but looking further, that doesn't always seem to be the case. Using the GH api could be a good alternative, I'll look into that.

elliott/elliottlib/cli/create_cli.py
```diff
@@ -13,6 +16,50 @@
...
 LOGGER = logutil.get_logger(__name__)
+GIT_REPO_OWNER = "openshift-eng"
+GIT_REPO_NAME = "ocp-build-data"
+GIT_REPO_BRANCH = "main"
+COMMON_ADVISORY_TEMPLATE_FILE = "config/advisory_templates.yml"
+
+
+def get_common_advisory_template():
+    gh_api = GhApi()
+    content = gh_api.repos.get_content(
```
thegreyd, line 27: we don't need to use gh api, since we already have build data git repo cloned and ready to use. You can use a method like get_file_from_branch https://github.com/openshift-eng/art-tools/blob/cf5fcf85d7ecb0f06369925a554d4922f6530b18/elliott/elliottlib/runtime.py#L439

elliott/elliottlib/runtime.py
```diff
@@ -94,6 +94,17 @@ def __init__(self, **kwargs):
     def get_major_minor(self):
         return self.group_config.vars.MAJOR, self.group_config.vars.MINOR
 
+    def get_major_minor_patch(self):
+        pattern = r"\d+.\d+.\d+"
+        match = re.search(pattern, self.assembly)
+        if match:
+            self._logger.debug(f"Found major.minor.patch for assembly {self.assembly}")
```
thegreyd, line 101: how about also logging the found value here?

elliott/elliottlib/cli/create_cli.py
```diff
@@ -128,31 +175,31 @@ async def create_cli(
         $ elliott -g openshift-4.14 create --art-advisory-key image --date 2018-Mar-05 --yes
     """
     runtime.initialize()
-
     et_data = runtime.get_errata_config()
 
-    if sum(map(bool, [date, batch_id])) != 1:
```
thegreyd, line 134: why remove this?

thegreyd: ah i see it got moved.. but why move?

elliott/elliottlib/cli/create_cli.py
```diff
@@ -13,6 +16,50 @@
...
+        ref=GIT_REPO_BRANCH,
+        token=os.environ["GITHUB_TOKEN"],
+        path=COMMON_ADVISORY_TEMPLATE_FILE,
+    )
+    decoded_content = base64.b64decode(content.content).decode("utf-8")
+    return yaml.safe_load(decoded_content)
+
+
+def get_advisory_boilerplate(et_data, date, batch_id, art_advisory_key, errata_type):
+    common_advisory_template = get_common_advisory_template()
```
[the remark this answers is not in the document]
ashwindasr: ack - updated in https://github.com/openshift-eng/art-tools/pull/1573/commits/588bcdd0118448bef4189750bf0d00936f20122d

elliott/elliottlib/cli/create_cli.py
ashwindasr: apologies - added in https://github.com/openshift-eng/art-tools/pull/1573/commits/588bcdd0118448bef4189750bf0d00936f20122d

elliott/elliottlib/runtime.py
ashwindasr: ack added in https://github.com/openshift-eng/art-tools/pull/1573/commits/588bcdd0118448bef4189750bf0d00936f20122d

elliott/elliottlib/cli/create_cli.py
ashwindasr: ack added in https://github.com/openshift-eng/art-tools/pull/1573/commits/588bcdd0118448bef4189750bf0d00936f20122d

elliott/elliottlib/cli/attach_cve_flaws_cli.py
```diff
@@ -150,7 +151,23 @@ def get_flaws(flaw_bug_tracker: BugTracker, tracker_bugs: Iterable[Bug], brew_ap
 def _update_advisory(runtime, advisory, flaw_bugs, bug_tracker, noop):
     advisory_id = advisory.errata_id
     errata_config = runtime.get_errata_config()
-    cve_boilerplate = errata_config['boilerplates']['cve']
+
+    # Get the advisory name, eg. image|rpm|metadata|microshift
+    releases_config = runtime.get_releases_config()
+    advisories = releases_config[runtime.assembly]['assembly']['group']['advisories']
```
thegreyd, line 157: i think there should be a better method which returns assembly config as Model then you can just do
`assembly_config.group.advisories`

elliott/elliottlib/cli/attach_cve_flaws_cli.py
```diff
@@ -150,7 +151,23 @@ def get_flaws(flaw_bug_tracker: BugTracker, tracker_bugs: Iterable[Bug], brew_ap
...
+    art_advisory_key = None
+    for _name, _id in advisories.items():
+        if _id == advisory_id:
+            art_advisory_key = _name
+            break
```
thegreyd, line 162:
```suggestion
    art_advisory_key = next((name for name, id in advisories.items() if id == advisory_id), None)
```

thegreyd, line 162:
```suggestion
    art_advisory_key = next((name for name, id in advisories.items() if id == advisory_id), None)
```

thegreyd: i typed it out, do test before committing :)

elliott/elliottlib/cli/attach_cve_flaws_cli.py
thegreyd: have a suggestion, instead of doing this here, it would be better if you passed `advisory_by_kind: dict` to this func. this advisory_by_kind can be prepared initially (see line 60-70)

elliott/elliottlib/util.py
```diff
@@ -25,6 +26,8 @@
 now = datetime.datetime.now()
 YMD = '%Y-%b-%d'
 LOGGER = get_logger(__name__)
+GIT_REPO_BRANCH = "main"
```
thegreyd, line 29: i don't think having this as a constant adds value .. , it is okay to hardcode it since it's an exceptional behavior

ashwindasr: thanks - updated in https://github.com/openshift-eng/art-tools/pull/1573/commits/544a92c5f8588db85ac8199fc945d3c181eb2b3c

elliott/elliottlib/cli/attach_cve_flaws_cli.py
ashwindasr: TIL!

updated in https://github.com/openshift-eng/art-tools/pull/1573/commits/544a92c5f8588db85ac8199fc945d3c181eb2b3c

elliott/elliottlib/cli/attach_cve_flaws_cli.py
ashwindasr: well if the goal is fail fast, I've made the update here: https://github.com/openshift-eng/art-tools/pull/1573/commits/e938741b82d243e6468f17776d8990c9a754d43a

I checked other places in the code base and this is how we were retrieving the value
```
releases_config["releases"][runtime.assembly]['assembly']['group']['advisories']
```
and I wanted it to fail fast too. if we change it to a model, we have to write additional lines to check if its a missing Model or not.

ashwindasr:
/unhold

create cli tested locally with microshift-bootc, works as expected. update advisory can only be checked once we encounter a case where an RHBA advisory becomes RHSA.

PR #1578, opened 2025-05-06
fix nvr issue

Ximinhan:
rstrip will cut the nvr, should use removesuffix


joepvd:
/hold

The `src.rpm` may be used in other parts, let's investigate

PR #1579, opened 2025-05-06
Replace flake8 with ruff lint

thegreyd:
Summary of changes
- remove flake8 as dependency
- migrate flake8 rules to ruff.toml
- remove tox.ini files (which have been used to contain flake8 rules)
- `make lint` uses `ruff check`
- progressively remove flake8 excludes and fix them


ruff.toml
```diff
@@ -37,7 +37,14 @@ target-version = "py311"
...
 # Enable Pyflakes (`F`) and a subset of the pycodestyle (`E`)  codes by default.
 select = ["E4", "E7", "E9", "F"]
-ignore = []
+ignore = [
+    # imported but unused
+    # __init__ often imports to make available
+    # but does not use
+    "F401",
+    # bare except
+    "E722",
```
thegreyd, line 46: These 2 are where the most lint warnings are, so keeping these intact for now, and hoping to address the "unused imports" in a future PR.

doozer/doozerlib/backend/konflux_fbc.py
```diff
@@ -9,11 +9,12 @@
 
 from artcommonlib import util as artlib_util
 from artcommonlib.konflux.konflux_build_record import (
+    Engine,
```
locriandev, line 12: very good catch, we were importing the package indirectly :)

locriandev (commented):
/lgtm

PR #1582, opened 2025-05-06
Update ruff rules

thegreyd:


ruff.toml
```diff
@@ -35,15 +35,32 @@ indent-width = 4
...
 [lint]
-# Enable Pyflakes (`F`) and a subset of the pycodestyle (`E`)  codes by default.
-select = ["E4", "E7", "E9", "F"]
+select = [
+    # pycodestyle
+    "E",
+    # Pyflakes
+    "F",
+    # isort
+    "I",
```
joepvd, line 44: Think this should be under a new group `extend-select`? (If I read that single comment some place correctly, and that comment was accurate)

thegreyd: https://docs.astral.sh/ruff/settings/#lint_extend-select 
extend-select from what i understand is `default ruff rules + our_list`

here when we define select, we are overriding default ruff rules and saying this is what we want

PR #1591, opened 2025-05-08
Enable defaults for shipment config init

thegreyd:
This is to have default RP's set per application in shipment-data repo

Test:

https://gitlab.cee.redhat.com/hybrid-platforms/art/ocp-shipment-data/-/merge_requests/20

```
$ elliott -g openshift-4.18 --assembly 4.18.3 --shipment-path https://gitlab.cee.red
hat.com/sidsharm/ocp-shipment-data@add_config shipment init --advisory-key image
...
shipment:
  metadata:
    product: ocp
    application: openshift-4-18
    group: openshift-4.18
    assembly: 4.18.3
    fbc: false
  environments:
    stage:
      releasePlan: ocp-art-advisory-stage-4-18
    prod:
      releasePlan: ocp-art-advisory-prod-4-18
  snapshot:
    name: test-snapshot
    spec:
      nvrs: []
...
```

### DIFF
--- a/elliott/elliottlib/cli/shipment_cli.py
+++ b/elliott/elliottlib/cli/shipment_cli.py
@@ -2,7 +2,6 @@ import sys
 
 import click
 from artcommonlib import logutil
-from artcommonlib.model import Model
 from doozerlib.backend.konflux_image_builder import KonfluxImageBuilder
 from ruamel.yaml import YAML
 
--- a/elliott/elliottlib/cli/shipment_cli.py
+++ b/elliott/elliottlib/cli/shipment_cli.py
@@ -58,7 +58,7 @@ class InitShipmentCli:
         # if stage/prod rpa are not given in cli, try to load them from shipment repo config
         # where defaults are set per application
         if not (self.stage_rpa and self.prod_rpa):
-            shipment_config = self.runtime.shipment_gitdata.load_yaml_file('config.yaml') or {}
+            shipment_config = self.runtime.shipment_gitdata.load_yaml_file('config.yaml', strict=False) or {}
             app_env_config = shipment_config.get("applications", {}).get(self.application, {}).get("environments", {})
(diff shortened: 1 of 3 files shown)


elliott/elliottlib/cli/shipment_cli.py
```diff
@@ -44,15 +45,23 @@ def __init__(
         advisory_key: str,
     ):
         self.runtime = runtime
-        self.application = application
-        self.snapshot = snapshot
+        self.application = application or KonfluxImageBuilder.get_application_name(self.runtime.group)
```
ashwindasr, line 48: there are no dots in the application name.
i.e we have to do:
```suggestion
        self.application = application or KonfluxImageBuilder.get_application_name(self.runtime.group).replace(".","-")
```

elliott/elliottlib/runtime.py
```diff
@@ -384,20 +384,6 @@ def resolve_shipment_metadata(self, strict=True):
             if '@' in self.shipment_path:
                 shipment_path, commitish = self.shipment_path.split('@')
 
-            # TODO: find a better solution
```
ashwindasr, line 387: So we don't need the `GITLAB_TOKEN` anymore?

thegreyd: That's right it is not really needed here. It is needed when the repo is private / or we want to make pushes to it, both aren't applicable here.

elliott/elliottlib/cli/shipment_cli.py
```diff
@@ -2,6 +2,7 @@
 
 import click
 from artcommonlib import logutil
+from artcommonlib.model import Model
```
thegreyd, line 5:
```suggestion
```

elliott/elliottlib/runtime.py
thegreyd: Not sure why my comment never got posted .. 
but yes it is not needed - it is needed when the repo is private or we want to push to it - neither of those is applicable here

ashwindasr (approved):

PR #1595, opened 2025-05-08
[konflux] slsa attestation raise exception

ashwindasr:
Looks like there is a 1-3 second delay from when the PLR completes and the attestation is created.

`@retry(reraise=True, wait=wait_fixed(10), stop=stop_after_attempt(3))` can retry it after 10 seconds, but the subcommand needs to raise an exception first


ashwindasr:
/hold

Until test build succeeds

ashwindasr:
/unhold

Test build succeeded

PR #1602, opened 2025-05-13
Temporarily run FIPS scan in dry-run mode

adobes1:
Temporarily run FIPS scan in dry-run mode until retry logic (and message improvements) are implemented to reduce Slack spam caused by false negatives


joepvd (approved):
/lgtm

PR #1603, opened 2025-05-13
ART-12770: Improve FIPS messages posted on Slack

adobes1:
Modify the opening message of the FIPS thread on Slack to include number of failed builds, list of affected versions, and link to the FIPS PipelineRun. 
[Test example](https://redhat-internal.slack.com/archives/C05HB1F1THP/p1747216469176879)

### DIFF
--- a/pyartcd/pyartcd/pipelines/scan_fips.py
+++ b/pyartcd/pyartcd/pipelines/scan_fips.py
@@ -21,7 +21,6 @@ from pyartcd.runtime import Runtime
 JIRA_PROJECT = "OCPBUGS"
 JIRA_DOMAIN = "https://issues.redhat.com/"
 
-INITIAL_SLACK_MSG = ":warning: FIPS scan has failed for some builds. Please verify (Triage <https://art-docs.engineering.redhat.com/sop/triage-fips/|docs>)"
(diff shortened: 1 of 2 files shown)


pyartcd/pyartcd/pipelines/scan_fips.py
```diff
@@ -40,6 +41,48 @@ def __init__(self, runtime: Runtime, data_path: str, all_images: bool, nvrs: Opt
...
         self.slack_client.bind_channel("#art-release")
 
+    async def get_pipelinerun_name(self) -> Optional[str]:
+        """
+        Returns the name of the currently running fips-pipeline PipelineRun
+        """
+        if self.taskrun_namespace is None or self.taskrun_name is None:
+            return None
+
+        cmd = [
```
ashwindasr, line 51: Why do we need the TaskRun? Can't we craft the PipelineRun URL and share that back? 

i.e. share https://console-openshift-console.apps.artc2023.pc3z.p1.openshiftapps.com/k8s/ns/art-cd/tekton.dev~v1~PipelineRun/fips-pipeline-run-g6lz4/ in the Slack message for example

ashwindasr (changes requested):

PR #1613, opened 2025-05-15
[Konflux] Make EC happy for mixed Konflux/Brew FBC release

vfreex:
https://github.com/openshift-eng/art-tools/pull/1541 only works for pure Konflux bundle build. In order to make EC happy for FBC release with mixed Konflux and Brew content, we need to have ImageDigestMirrorSet to mirror unreleased images from Brew registry. Luckly, we have Brew pullspecs in our BigQuery database.


doozer/doozerlib/backend/konflux_fbc.py
```diff
@@ -310,6 +311,16 @@ async def rebase(self, metadata: ImageMetadata, bundle_build: KonfluxBundleBuild
...
                 self._record_logger.add_record("rebase_fbc_konflux", **record)
 
+    async def _get_referenced_images(self, metadata: ImageMetadata, bundle_build: KonfluxBundleBuildRecord):
+        assert bundle_build.operator_nvr, "operator_nvr is empty; doozer bug?"
+        nvrs = {bundle_build.operator_nvr}
+        if bundle_build.operand_nvrs:
+            nvrs |= set(bundle_build.operand_nvrs)
+        assert metadata.runtime.konflux_db is not None, "konflux_db is not initialized. Doozer bug?"
+        konflux_db: KonfluxDb = metadata.runtime.konflux_db
+        ref_builds = await konflux_db.get_build_records_by_nvrs(list(nvrs))
```
thegreyd, line 321: I think this assumes that the db is bound to the right table? maybe assert here?
or call the .bind again and revert after operation?

vfreex: Good point! I will fix this in the next push.

vfreex: I updated to use assert to ensure the right table is bound. Binding again might cause race condition without locking if it is not the right table.

PR #1619, opened 2025-05-15
ART-12761: update docs workflow

ashwindasr:
Changes in this PR:
- Uncomment `solution=cve_boilerplate['solution'],`
- Create variable for `security_impact`

### DIFF
--- a/elliott/elliottlib/util.py
+++ b/elliott/elliottlib/util.py
@@ -606,11 +606,4 @@ def get_advisory_boilerplate(runtime, et_data, art_advisory_key, errata_type):
     if art_advisory_key not in boilerplate:
         raise ValueError(f"Boilerplate {art_advisory_key} not found in erratatool.yml")
 
-    # Get the boilerplate for a type of errata and advisory type
-    try:
-        advisory_boilerplate = boilerplate[art_advisory_key][errata_type]
-    except KeyError:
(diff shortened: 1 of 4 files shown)


elliott/elliottlib/util.py
```diff
@@ -606,11 +606,4 @@ def get_advisory_boilerplate(runtime, et_data, art_advisory_key, errata_type):
...
 
-    # Get the boilerplate for a type of errata and advisory type
-    try:
-        advisory_boilerplate = boilerplate[art_advisory_key][errata_type]
-    except KeyError:
-        # For backwards compatibility with older versions of erratatool.yml, i.e. rhsa, rhba keys does not exist
-        advisory_boilerplate = boilerplate[art_advisory_key]
-
-    return advisory_boilerplate
+    return boilerplate[art_advisory_key][errata_type]
```
thegreyd, line 609: why remove this? some advisories will not have bugs attached like advance/prerelease, so they will never be rhsa, so instead of this distinction, a default makes sense

ashwindasr: disussion here: https://redhat-internal.slack.com/archives/GDBRP5YJH/p1747332636707259?thread_ts=1747330383.412199&cid=GDBRP5YJH

I've kept back the fallback for now: https://github.com/openshift-eng/art-tools/pull/1619/commits/278a22923449876f052cb6ca42280caeda8aed7d

PR #1634, opened 2025-05-21
Report on Konflux errors for 4.20

joepvd:


ashwindasr (changes requested):

pyartcd/pyartcd/pipelines/images_health.py
```diff
@@ -38,6 +39,13 @@ async def run(self):
         if any([self.send_to_release_channel, self.send_to_forum_ocp_art]):
             await asyncio.gather(*[self._send_notifications(engine) for engine in ['brew', 'konflux']])
 
+    async def send_for_engine(self, engine):
```
ashwindasr, line 42:
```suggestion
    def send_for_engine(self, engine):
```
Should not be async here, since we are not using `await`

ashwindasr (changes requested):

PR #1636, opened 2025-05-22
[Konflux] Add request timeouts to all API calls

vfreex:
In case of a network outage, KonfluxClient may hang indefinitely without raising any exception. Adding request timeouts should resolve this issue.

See
https://redhat-internal.slack.com/archives/GDBRP5YJH/p1747817430982149 and https://github.com/kubernetes-client/python/blob/master/examples/watch/timeout-settings.md.


joepvd (approved):
/lgtm

PR #1648, opened 2025-05-28
gen-assembly: fix yaml dump

locriandev:
- `print(yaml.dump)` erroneously prints the return value of `dump`, which broke artcd at the time it loads back the report
- `yaml.width = 4096` prevents Ruamel YAML from splitting simple `key: value` entries on multiple lines, if the string is long

### DIFF
--- a/doozer/doozerlib/cli/release_gen_assembly.py
+++ b/doozer/doozerlib/cli/release_gen_assembly.py
@@ -11,13 +11,12 @@ from artcommonlib import exectools, rhcos
 from artcommonlib.arch_util import go_arch_for_brew_arch, go_suffix_for_arch
 from artcommonlib.assembly import AssemblyTypes
 from artcommonlib.constants import RHCOS_RELEASES_STREAM_URL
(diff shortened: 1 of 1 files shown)


doozer/doozerlib/cli/release_gen_assembly.py
```diff
@@ -11,13 +11,12 @@
 from artcommonlib.arch_util import go_arch_for_brew_arch, go_suffix_for_arch
 from artcommonlib.assembly import AssemblyTypes
 from artcommonlib.constants import RHCOS_RELEASES_STREAM_URL
-from artcommonlib.konflux.konflux_build_record import Engine, KonfluxBuildOutcome, KonfluxBuildRecord
+from artcommonlib.konflux.konflux_build_record import  KonfluxBuildOutcome, KonfluxBuildRecord
```
thegreyd, line 14:
```suggestion
from artcommonlib.konflux.konflux_build_record import KonfluxBuildOutcome, KonfluxBuildRecord
```

ashwindasr (approved):

PR #1609, opened 2025-05-14
Sync konflux builds to default imagestream for 4.20

ashwindasr:
For 4.20, sync konflux builds to the default image stream `4.20-art-latest`

PR #1655, opened 2025-05-29
imagestream override konflux

ashwindasr:
For 4.20 for example, we are syncing konflux builds to formerly brew imagestreams. So we need to look at the usual nightly URLs instead.

Follows: https://github.com/openshift-eng/art-tools/pull/1609


thegreyd:
/lgtm
should be good after format fix

PR #1657, opened 2025-05-30
Handle exceptions raised when failing to get Konflux SLSA attestation

locriandev:
`scan_network_mode_changes()` is breaking many ocp4-scan-konflux builds. Handling ChildProcessErrors to keep the pipeline intact


vfreex (approved):
/lgtm

PR #1658, opened 2025-05-30
Adjust build-sync for Konflux

locriandev:
- `assembly.basis.time` is now a string, and needs to be converted back into a `datetime` object
- for 4.20 we're syncing out Konflux builds to standard nightlies, so we need to remove the `-konflux` suffix in the pullspec

### DIFF
--- a/doozer/doozerlib/cli/release_gen_payload.py
+++ b/doozer/doozerlib/cli/release_gen_payload.py
@@ -2286,7 +2286,12 @@ class PayloadGenerator:
         if major_minor != runtime.get_minor_version():
             return terminal_issue(f"Specified nightly {nightly} does not match group major.minor")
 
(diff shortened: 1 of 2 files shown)


doozer/doozerlib/cli/release_gen_payload.py
```diff
@@ -2286,7 +2286,12 @@ def terminal_issue(msg: str) -> List[AssemblyIssue]:
         if major_minor != runtime.get_minor_version():
             return terminal_issue(f"Specified nightly {nightly} does not match group major.minor")
 
-        release_suffix = f"{'konflux-' if runtime.build_system == 'konflux' else ''}release"
+        # For 4.20, remove the -konflux suffix as Konflux builds are being mirrored to standard imagestreams
+        if runtime.build_system == 'brew' or major_minor == '4.20':
```
thegreyd, line 2290: I think there's a constant list available to use instead of the hardcoded value?

PR #1659, opened 2025-05-30
Add retry option to FIPS scan pipeline

adobes1:
Add `--retry` option to FIPS scan pipeline to retry the scan for failing NVRs. This should hopefully decrease the number of false positives


joepvd:
I do not see a strict need to add `--retry` to the invocation. I see this as an implementation detail tyhat does not need exposure as a cli option. If it does, it needs to get a default value. 

Also, getting the feeling that this is more complicated than need be. Could you run `git grep @retry` and see if that approach is applicable?

PR #1664, opened 2025-06-02
Fix prepare release bugs

ashwindasr:
Make the pipeline work with konflux / brew


pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -314,7 +314,10 @@ async def run(self):
                 self.pre_release = True
 
                 # Remove all builds from the metadata advisory
-                await self.remove_builds_all(advisories["metadata"])
+                try:
```
thegreyd, line 317: instead of try-catch, would like a `if 'metadata' in advisories": drop builds`

PR #1666, opened 2025-06-03
Parallelize plashets building processes

locriandev:
Refactoring out our `plashets` module to let us make use of `asyncio` and concurrency. 

As a result of this changes, 4.19 test plashets were rebuilt in just 10 minutes ([test build](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fbuild-plashets/10/))

### DIFF
--- a/pyartcd/pyartcd/plashets.py
+++ b/pyartcd/pyartcd/plashets.py
@@ -11,11 +11,9 @@ from artcommonlib import exectools
 from pyartcd import constants, util
 from pyartcd.constants import PLASHET_REMOTES
 
-working_dir = "plashet-working"
+WORKING_DIR = "plashet-working"
 
-logger = logging.getLogger('pyartcd')
-
-previous_packages = [
+PREVIOUS_PACKAGES = [
     "conmon",
     "cri-o",
     "cri-tools",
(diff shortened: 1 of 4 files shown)


pyartcd/tests/pipelines/test_plashets.py
```diff
@@ -8,14 +8,14 @@ class TestBuildCompose(unittest.IsolatedAsyncioTestCase):
     @patch('pyartcd.jenkins.init_jenkins')
```
lgarciaaco, line 8: While these tests are focused on the BuildPlashetsPipeline, it would be nice to cover the actual logic of `PlashetBuilder` itself.

pyartcd/pyartcd/plashets.py
```diff
@@ -43,360 +41,369 @@
...
+        }
+        """
+
+        await self.load_and_configure(repos)
+        await asyncio.gather(
+            *[self.build_plashet(repo_type, config) for repo_type, config in self._plashet_config.items()]
+        )
+        return self._plashets_built
+
+    async def build_plashet_from_tags(
```
lgarciaaco, line 283: You're mixing os.PathLike, Path, and str a bit interchangeably. In methods like `build_plashet_from_tags`, it might be clearer to just stick with Path to reduce casting clutter and make intent clearer — unless there's a specific reason for juggling multiple types

lgarciaaco (commented):
Looks good overall. The only suggestion is to consider adding some unit tests for `PlashetBuilder`—even basic coverage could help catch regressions.

locriandev:
Closing as this might be problematic with errata tool signing process.

PR #1670, opened 2025-06-05
ART-13075: Add check for self-referential upgrades

adobes1:
Add check to ocp-build-data-validator to ensure that the `upgrades` field in `releases.yml` does not contain references to itself.

If there's self-referential entry, displays the following error:
```
validator.exceptions.ValidationFailed: releases.yml validation failed
Returned error: The following releases contain references to themselves in their respective 'upgrades': 4.18.14, 4.18.13, rc.1, ec.1
```

### DIFF
--- a/ocp-build-data-validator/validator/releases.py
+++ b/ocp-build-data-validator/validator/releases.py
@@ -1,0 +1,30 @@
+from typing import List, Optional
+
+from . import support
+
+
+def validate(data: dict) -> Optional[str]:
+    self_referential_upgrades = get_self_referential_upgrades(data)
+    if self_referential_upgrades:
+        return f"The following releases contain references to themselves in their respective 'upgrades': {', '.join(self_referential_upgrades)}"
+
+
+def get_self_referential_upgrades(releases_config: dict) -> List[str]:
+    """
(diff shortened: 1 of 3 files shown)


ocp-build-data-validator/validator/releases.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+    self_referential_releases = []
+    for release, release_config in releases_config['releases'].items():
+        full_release = release
+
+        if release.startswith('ec') or release.startswith('rc'):
+            # Construct full release name
+            group_config = support.load_group_config_for('releases.yml')
+            major, minor = group_config['vars']['MAJOR'], group_config['vars']['MINOR']
+            full_release = f"{major}.{minor}.0-{release}"
```
ashwindasr, line 24:
```suggestion
            full_release = f"{group_config['vars']['MAJOR']}.{group_config['vars']['MINOR']}.0-{release}"
```
since the `major`, `minor` variables are used immediately after, and is not being used anywhere else, we can directly put in inside the formatted string.

adobes1: Thank you! Fixed

PR #1704, opened 2025-06-16
[Konflux] Add required labels for OLM bundles

vfreex:
Labels like `distribution-scope` and `url` are required by Conforma (Enterprise Policy).


joepvd (approved):
/lgtm

PR #1707, opened 2025-06-17
Fix YAML parser error message compatibility in format tests

lgarciaaco:
## Summary
  - Update format test to handle different error messages from various YAML parser versions
  - Replace exact string match with multiple possible error message patterns

  ## Problem
  The test was failing on systems with different YAML parser versions due to hardcoded error message
  expectations.

  ## Solution
  Modified the test assertion to check for any of the common YAML parsing error messages:
  - "did not find expected key"
  - "expected <block end>, but found"
  - "while parsing a block mapping"

### DIFF
--- a/ocp-build-data-validator/tests/test_format.py
+++ b/ocp-build-data-validator/tests/test_format.py
@@ -12,7 +12,13 @@ class TestFormat(unittest.TestCase):
         """
         (parsed, err) = format.validate(invalid_yaml)
         self.assertIsNone(parsed)
-        self.assertIn("did not find expected key", err)
+        # Check for different possible error messages from different YAML parser versions
+        self.assertTrue(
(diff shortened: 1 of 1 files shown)


joepvd (approved):
/lgtm

ocp-build-data-validator/tests/test_format.py
```diff
@@ -12,7 +12,13 @@ def test_invalid_yaml(self):
         """
         (parsed, err) = format.validate(invalid_yaml)
         self.assertIsNone(parsed)
-        self.assertIn("did not find expected key", err)
+        # Check for different possible error messages from different YAML parser versions
```
locriandev, line 15: I am wondering, does the parser not raise an Exception when encountering invalid input? We could use `assertRaises` in that case

PR #1713, opened 2025-06-18
Fix lockfile pipeline continuity by downloading upstream files when digest matches

lgarciaaco:
## Problem

So basically when we find the digest file upstream and it matches, we'd skip generating the lockfile but then not download the files locally. This meant the next time
  the pipeline ran, the files would be missing upstream and we would generate the files again. This would end up in a pattern where we generate the files every second run.

  The flow was:
  1. Find matching digest upstream → skip generation
  2. Don't write anything locally
  3. Next pipeline run → files are gone upstream
  4. Not needed run 🙃

 ## Solution

  Now when we find a matching digest upstream, we actually download both the digest and lockfile to make sure they're availible for the next run:

  - Added `_get_lockfile_from_target_branch()` to grab the lockfile from upstream
  - Added `_get_target_branch()` helper since we were duplicating that logic everywhere
  - Fixed `generate_lockfile()` to actually download both files when digest matches
  - Added proper error handling and logging

### DIFF
--- a/doozer/doozerlib/lockfile.py
+++ b/doozer/doozerlib/lockfile.py
@@ -247,6 +247,24 @@ class RPMLockfileGenerator:
         self.builder = RpmInfoCollector(repos, self.logger)
         self.runtime = runtime
 
+    def _get_target_branch(self, distgit_key: str) -> Optional[str]:
+        """
+        Construct target branch name using the same format as rebaser.
+
+        Args:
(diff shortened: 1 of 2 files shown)


lgarciaaco:
[test job](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-konflux/11967/console)

PR #1718, opened 2025-06-18
apply .0 if go version >= 1.22

ashwindasr:
Do not need to check for `self._runtime.group_config.konflux.cachi2.gomod_version_patch` since we need to add the .0 if its missing

### DIFF
--- a/doozer/doozerlib/backend/rebaser.py
+++ b/doozer/doozerlib/backend/rebaser.py
@@ -14,6 +14,7 @@ from typing import Dict, List, Optional, Set, Tuple, cast
 import aiofiles
 import bashlex
 import bashlex.errors
+import semver
 import yaml
 from artcommonlib import exectools, release_util
 from artcommonlib.konflux.konflux_build_record import Engine, KonfluxBuildRecord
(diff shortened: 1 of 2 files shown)


doozer/doozerlib/backend/rebaser.py
```diff
@@ -463,32 +463,30 @@ class MergeSourceResult:
...
+            with open(gomod_path, "r") as file:
+                lines = file.readlines()
+                for line in lines:
+                    stripped_line = line.strip()
+                    match = re.match(r"(^go \d\.\d+$)", stripped_line)
+                    if match:
+                        # Append a .0 to the go mod version, if it exists
+                        # Replace the line 'go 1.22' with 'go 1.22.0' for example
+                        go_version_string = match.group(1)  # eg. 'go 1.23'
+                        go_version_number = float(go_version_string.split(" ")[-1])  # eg. 1.23
```
vfreex, line 479: Converting version numbers to float looks error-prone to me. Is it possible to convert that to a tuple of integer or use a library like `semver`?

ashwindasr:
/hold

Until test build succeeds

ashwindasr:
/unhold

Test builds succeeded

doozer/doozerlib/backend/rebaser.py
ashwindasr: thanks yuxiang, added in https://github.com/openshift-eng/art-tools/pull/1718/commits/e2e51ff923bd035ec2622cff344db4164cfa9d89

PR #1722, opened 2025-06-19
[Konflux] Fix deadlock in Rebaser

vfreex:
`rebase_event` is of type `multiprocessing.Event`. `wait` method shouldn't be called from a coroutine because it will essentially block the event loop. Polling for `is_set()` should be the way to go.

See the issue: https://redhat-internal.slack.com/archives/GDBRP5YJH/p1750284749153469?thread_ts=1750075062.715439&cid=GDBRP5YJH


locriandev (commented):
/lgtm

PR #1727, opened 2025-06-20
Improve(exectools): resolve event loop binding bug in limit_concurrency decorator

lgarciaaco:
The limit_concurrency decorator had a flaw where semaphores were created at
import time. This fix creates per-event-loop semaphores using
WeakKeyDictionary to prevent memory leaks.

Additional improvements:
- Add proper type hints with TypeVar for better type safety
- Add parameter validation and async function checking
- Enhanced documentation with comprehensive docstring
- Import weakref and additional typing utilities for robust implementation

### DIFF
--- a/artcommon/artcommonlib/exectools.py
+++ b/artcommon/artcommonlib/exectools.py
@@ -11,13 +11,14 @@ import sys
 import threading
 import time
 import traceback
+import weakref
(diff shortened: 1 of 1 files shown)


locriandev (commented):
/lgtm

PR #1745, opened 2025-06-25
fix(image): add group config fallback to is_lockfile_force_enabled

lgarciaaco:
## Summary
Add missing group configuration fallback to `is_lockfile_force_enabled()` method to maintain consistency with sibling configuration hierarchy methods in the ImageMetadata class.

## Problem
**Before:** The `is_lockfile_force_enabled()` method only checked image metadata configuration (`self.config.konflux.cachi2.lockfile.force`) and immediately returned `False` if not set, skipping group configuration check entirely.

**After:** The method now follows the same configuration hierarchy pattern as `is_cachi2_enabled()` and `is_lockfile_generation_enabled()`:
1. Check image metadata configuration first (highest priority)
2. If not set, check group configuration (`self.runtime.group_config.konflux.cachi2.lockfile.force`)
3. Default to `False` if neither is set

## Changes
- **doozer/doozerlib/image.py**: Add group configuration fallback logic and update docstring
- **doozer/tests/test_image.py**: Add comprehensive test coverage for group config scenarios

### DIFF
--- a/doozer/doozerlib/image.py
+++ b/doozer/doozerlib/image.py
@@ -726,10 +726,11 @@ class ImageMetadata(Metadata):
         """
         Determines whether lockfile force generation is enabled for the current image configuration.
 
-        This method only checks the image metadata configuration for the force parameter:
-        - Image metadata configuration (`self.config.konflux.cachi2.lockfile.force`)
+        The method checks configuration in the following order:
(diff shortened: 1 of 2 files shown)


joepvd (approved):
/lgtm

PR #1763, opened 2025-07-02
[Konflux] ART-13198: Snapshot creation for bundle and fbc builds

vfreex:
This PR makes it possible to create Konflux Snapshot(s) for a list of mixed container images, OLM bundles, and FBC fragments using `elliott snapshot new` command.

Multiple snapshot instances will be created if passed builds are from multiple Konflux applications, one for each application.

This PR also removes `--for-fbc` option because we can determine whether a build is FBC or not according to its NVR.


elliott/elliottlib/cli/snapshot_cli.py
```diff
@@ -237,7 +263,6 @@ def snapshot_cli():
     metavar='PATH',
     help='Path to the pull secret file to use. For example, if the images are in quay.io/org/repo then provide the pull secret to read from that repo.',
 )
-@click.option('--for-fbc', is_flag=True, help='To indicate that the given builds are fbc builds.')
```
thegreyd, line 240: thanks for removing this. Can you also remove it from konflux_release_cli.py as well? I added it there so i can pass it on here

elliott/tests/test_snapshot_cli.py
```diff
@@ -180,14 +188,14 @@ async def test_run_happy_path(self, mock_runtime, mock_konflux_client_init, mock
...
-                        "com.redhat.component": "test-component",
+                        "com.redhat.component": "test-component-container",
                         "version": f"v{self.major}.{self.minor}.0",
                         "release": "assembly.test.el8",
                     },
                 },
             },
         }
-        expected_nvrs = [f"test-component-v{self.major}.{self.minor}.0-assembly.test.el8"]
+        expected_nvrs = [f"test-component-container-v{self.major}.{self.minor}.0-assembly.test.el8"]
```
thegreyd, line 198: Not specific to this line, but i don't see a test for creating multiple snapshots. can you add that?

elliott/elliottlib/cli/snapshot_cli.py
```diff
@@ -46,25 +59,27 @@ async def _get(nvrs) -> list[KonfluxRecord]:
...
-        bundle_records = await _get(bundle_nvrs)
-        records.extend(bundle_records)
-    if non_bundle_nvrs:
-        if for_fbc:
-            runtime.konflux_db.bind(KonfluxFbcBuildRecord)
+    # Group NVRs by their type to fetch them in parallel
+    type_nvrs = defaultdict(list)
+    for nvr in nvrs:
+        nvr_dict = parse_nvr(nvr)
+        if nvr_dict['name'].endswith('-bundle-container'):
```
Ximinhan, line 66: The bundle name could also use `-metadata-container`

vfreex: Good point @Ximinhan ! I will check and add it in the next push.

PR #1766, opened 2025-07-02
Support build-data in tools shipment schema

thegreyd:


elliott/elliottlib/shipment_model.py
```diff
@@ -137,6 +137,7 @@ class Tools(StrictBaseModel):
     """Tools to use when releasing shipment to an environment"""
 
     art_tools: Optional[str] = Field(None, alias="art-tools")
+    build_data: Optional[str] = Field(None, alias="build-data")
```
locriandev, line 140: Is this equivalent to what we currently name `data-path`? If so, I'd stick to the usual nomenclature

thegreyd: This is a slightly different convention, modeled after `art_tools_commit` - `username@branch`. I will make a formal proposal to change --data-path conventions, as they are not ideal. Also since the term --data-path is confusing now that we have 2 data repos: build-data and shipment-data.

prepare-release-konflux pipeline uses a similar convention right now which combines --data-path and data_gitref attributes in a build_repo_url:

BUILD_REPO_URL
(Optional) Override build-data repo. Defaults to group branch - to use a different branch/commit use repo@branch. e.g. https://github.com/thegreyd/ocp-build-data@prep-shipment-4.19.ec5

PR #1768, opened 2025-07-03
fix: add architecture support to RPM resolution in repodata

lgarciaaco:
The get_rpms method in repodata.py wasn't considering architecture when selecting RPMs, which could lead to wrong packages being picked for lockfile generation. This adds an optional arch parameter that:

- Prefers exact architecture matches first
- Falls back to noarch packages if available
- Only uses first match as last resort
- Updates lockfile.py to pass the architecture parameter
- Includes comprehensive tests for all the new behavior

This should fix issues where lockfiles might include wrong-architecture RPMs.


rayfordj (commented):
/lgtm

PR #1775, opened 2025-07-07
feat(konflux): add configurable component names for build records

lgarciaaco:
### BREAKING: Can't be merged without Schema changes
Allow Konflux component names to be specified in metadata configuration instead of always generating them automatically. Adds component field support to all Konflux build record types and updates snapshot logic to use stored component names when available.

This provides more control over component naming in Konflux builds while maintaining backward compatibility with generated names as fallback.


elliott/elliottlib/cli/snapshot_cli.py
```diff
@@ -154,7 +154,16 @@ async def new_snapshot(self, build_records) -> dict:
         await self.konflux_client._get(API_VERSION, KIND_APPLICATION, application_name)
 
         async def _comp(record):
-            comp_name = KonfluxImageBuilder.get_component_name(application_name, record.name)
+            # Use stored component name from build record if available, otherwise generate default
+            if hasattr(record, 'component') and record.component:
+                comp_name = record.component
+            elif hasattr(record, 'bundle_component') and record.bundle_component:
```
joepvd, line 160: hmm i think these component types could be defined for a single image, making `elif` wrong.

You might also want to look into `Missing` handling elsewhere.

artcommon/artcommonlib/konflux/konflux_build_record.py
```diff
@@ -85,6 +85,7 @@ def __init__(
         pipeline_commit: str = '',
         schema_level: int = 0,
         ingestion_time: datetime = None,
+        component: str = '',
```
thegreyd, line 88: I think we should have a field called "konflux_component" in all 3 build records. 2 reasons
- `component` is an overloaded term, so this makes it clear
- the field name is standardized across tables, to keep it similar to fields like name, version, release which each build has.

doozer/doozerlib/backend/konflux_image_builder.py
```diff
@@ -555,6 +555,7 @@ async def update_konflux_db(
             'build_id': f'{pipelinerun_name}-{pipelinerun_uid}',
             'build_pipeline_url': build_pipeline_url,
             'pipeline_commit': 'n/a',  # TODO: populate this
+            'konflux_component': metadata.konflux.get('component', ''),
```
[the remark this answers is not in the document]
thegreyd: also you would have to modify these get_component_name type methods to return metadata override if there is one defined.

thegreyd (changes requested):

PR #1782, opened 2025-07-09
ART-3761 Release date in assembly definition

thegreyd:
https://issues.redhat.com/browse/ART-3761

The goal is to move all release date interactions into one single place i.e. in gen-assembly to the assembly definition. And then everything else depends on that (prepare-release, build-microshift etc..)

This then would also be used to gate shipment pushes.

Test:

https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fgen-assembly/621/console

```
2025-07-09 02:13:58,256 artcommonlib INFO Release date not provided. Fetching from release schedule...
2025-07-09 02:13:58,379 artcommonlib INFO Using release date: 2025-07-15 00:00:00
2025-07-09 02:13:58,485 artcommonlib INFO Using in-flight release: 4.18.20
```


doozer/doozerlib/cli/release_gen_assembly.py
```diff
@@ -306,6 +319,27 @@ def _validate_params(self):
...
+        if not self.release_date:
+            self.logger.info("Release date not provided. Fetching from release schedule...")
+            self.release_date = get_assembly_release_date(
+                self.gen_assembly_name, self.assembly_type, self.runtime.group
+            )
+        self.logger.info("Using release date: %s", self.release_date)
+
+        determined_in_flight = get_in_flight(self.runtime.group, self.release_date)
+        if self.in_flight:
+            if self.in_flight != determined_in_flight:
```
Ximinhan, line 335: if self.in_flight and self.in_flight != determined_in_flight

If humans deliberately set it, we should just honor the human input, then no need to get_in_flight, during ec/rc the get_in_flight may get none and input value in_flight exist, the warning is not correct

thegreyd: yes I will add a None check. I can make it an info message

PR #1791, opened 2025-07-14
ART-13350 prepare-release-konflux creates jira ticket

locriandev:
Test build https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fprepare-release-konflux/101 run from a different branch where bundle builds were bypassed, as we currently have a rebase failure that would prevent the job from succeeding.

[ART-13418](https://issues.redhat.com/browse/ART-13418) was created as a clone of the new template ticket

### DIFF
--- a/pyartcd/pyartcd/pipelines/prepare_release_konflux.py
+++ b/pyartcd/pyartcd/pipelines/prepare_release_konflux.py
@@ -67,7 +67,7 @@ class PrepareReleaseKonfluxPipeline:
         self.logger = logging.getLogger(__name__)
         self.runtime = runtime
         self.assembly = assembly
-        self.release_date = date
+        self.date = date
         self.group = group
 
         self._slack_client = slack_client
@@ -172,6 +172,10 @@ class PrepareReleaseKonfluxPipeline:
     def release_name(self) -> str:
         return get_release_name_for_assembly(self.group, self.releases_config, self.assembly)
 
+    @cached_property
+    def release_date(self):
+        return self.date or self.group_config.get("release_date")
+
     @property
     def assembly_group_config(self) -> dict:
         return assembly_config_struct(Model(self.releases_config), self.assembly, "group", {})
@@ -185,7 +189,6 @@ class PrepareReleaseKonfluxPipeline:
         self.setup_working_dir()
         await self.setup_repos()
         await self.validate_assembly()
-        await self.get_release_date()
 
         self.jira_client = JIRAClient.from_url(self.runtime.config["jira"]["url"], token_auth=self.jira_token)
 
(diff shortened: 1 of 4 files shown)


pyartcd/pyartcd/pipelines/prepare_release_konflux.py
```diff
@@ -230,6 +264,18 @@ async def validate_assembly(self):
                 f"Product mismatch: {group_product} != {self.product}. This pipeline only supports {self.product}."
             )
 
+    async def get_release_date(self):
```
Ximinhan, line 267: From what I know, the release date will be set in the assembly config, and we should rely on that

locriandev: gen-assembly is not currently adding the release date to the assembly config ([example](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fgen-assembly/634/console)). So prepare-release can currently only fetch the release date from the release schedule

Ximinhan: yes, https://github.com/openshift-eng/art-tools/pull/1782 plan to add that, then we can use `group_config.get("release_date", "")`, I mean let's not overlap the work

pyartcd/pyartcd/pipelines/prepare_release_konflux.py
```diff
@@ -958,19 +1092,11 @@ async def prepare_release(
     runtime: Runtime,
     group: str,
     assembly: str,
+    date: str,
     build_repo_url: Optional[str],
     shipment_repo_url: Optional[str],
 ):
-    job_url = os.getenv('BUILD_URL')
```
thegreyd, line 964: i'm not sure about this.. 
I find it better to get all env vars in the click command and then pass it to pipeline. It makes mocking and testing easier, the external inputs are gathered together and passed ..

pyartcd/pyartcd/pipelines/prepare_release_konflux.py
thegreyd: yes, i agree with ximin. Would propose:
- use date if --date is set
- if not rely on group_config.get("release_date", "")
- complain if not found

thegreyd (changes requested):

pyartcd/pyartcd/pipelines/prepare_release_konflux.py
locriandev: I didn't have to change tests, looks like no overhead? OTOH this makes code cleaner imo but it's easy to revert if you have strong opinions

PR #1792, opened 2025-07-14
Konflux: Use multiple components for FBC

vfreex:
Currently every FBC component lives in its own Konflux application due to Konflux's own limitation. This complicates konflux-release-data configuration and shipment pipeline implementation. Once multi-component FBC support is deployed, we can merge this PR to move to multi-component FBC.


thegreyd:
/unhold
In coordination with Konflux team, we started using `ocp-art` branch instead of `production` where the fix for this is deployed. Unholding it so we can trigger builds and run tests

PR #1798, opened 2025-07-16
ART-13327: Adds a allowlist for security level that can be accessed by ART bot

san7ket:


elliott/elliottlib/cli/find_bugs_sweep_cli.py
```diff
@@ -23,6 +23,14 @@
 type_bug_set = Set[Bug]
 yaml = new_roundtrip_yaml_handler()
 
+# Allowlist for JIRA security levels that are permitted
+JIRA_SECURITY_ALLOWLIST = {
+    "Red Hat Employee",
+    "Restricted", 
+    "Red Hat Partner",
+    "Red Hat Engineering Authorized"
```
joepvd, line 31: `Embargoed security issue` or similar should be part of this list.

elliott/elliottlib/cli/find_bugs_sweep_cli.py
```diff
@@ -516,14 +524,34 @@ def categorize_bugs_by_type(
 
     not_found = set(tracker_bugs) - found
     if not_found:
-        still_not_found = not_found
+        # Filter out JIRA bugs with security levels not in the allowlist
+        filtered_not_found = set()
+        security_filtered_bugs = set()
+        
+        for bug in not_found:
```
joepvd, line 531: This may work, but I would have imagined this to be implemented in the `bzutil.JIRABugTracker._query()` method. This is the global entrypoint for all queries. Did you look into formulating this exclusion in terms of jql?

san7ket: No I didn't, but I agree this will filter the bugs at the start so, we would need to crunch through less bugs. I will look into this.

vfreex (commented):
/lgtm

PR #1812, opened 2025-07-17
feat: combine build and config RPMs for lockfile generation

lgarciaaco:
Add support for including RPMs from both build data and explicit configuration when generating lockfiles. The new get_lockfile_rpms_to_install() method returns the union of RPMs from fetch_rpms_from_build() and meta.config.konflux.cachi2.lockfile.rpms, giving more flexibility in defining which packages should be included in the lockfile.

### DIFF
--- a/ocp-build-data-validator/tests/test_schema/test_image_schema.py
+++ b/ocp-build-data-validator/tests/test_schema/test_image_schema.py
@@ -135,16 +135,6 @@ class TestImageSchema(unittest.TestCase):
         }
         self.assertIn("123 is not of type 'string'", image_schema.validate('filename', invalid_data))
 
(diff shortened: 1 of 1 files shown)


joepvd:
Awesome! Could you set this new config option in https://github.com/openshift-eng/art-tools/blob/main/ocp-build-data-validator/validator/json_schemas/image_content.base.schema.json, so the validator will not fail when we set it?

lgarciaaco:
@joepvd https://github.com/openshift-eng/art-tools/pull/1812/commits/88d25c57e26a8eb1ba062210121e3f0736bb6bc0

vfreex (commented):
/lgtm

PR #1566, opened 2025-04-30
ART-12529: append suffix to CVEs if it is needed

mbiarnes:
There are three CVE tracker bugs created for 4.11

- https://issues.redhat.com/browse/OCPBUGS-55507
- https://issues.redhat.com/browse/OCPBUGS-55232
- https://issues.redhat.com/browse/OCPBUGS-55508

which have or don't have in their summary [openshift-{maj}.{min}] or simply don't have a suffix.

This was the elliott command applied:
_elliott --group=openshift-4.11 --assembly=stream --data-path=https://github.com/openshift/ocp-build-data
find-bugs:sweep --cve-only_

When executing this command applying changes of this PR you will see that all CVE trackers will have as ssuffix
**[openshift-4.11]**

PR #1830, opened 2025-07-24
[Elliott] Accept target version summary prefixes

vfreex:
https://github.com/openshift-eng/art-tools/pull/1566 and https://github.com/openshift-eng/art-tools/pull/1749 added a validation to find-bugs. In case a CVE tracker doesn't have a target version suffix, e.g. `[openshift-x.y]` in its summary, it is considered invalid.

Recently, we found some CVE trackers have `[openshift-x.y]` summary prefixes instead of suffixes. Guess those prefixes are valid either.

This PR makes a change in Elliott to accept those kind of prefixes.


joepvd (approved):
/lgtm

PR #1841, opened 2025-07-25
allow konflux re-release for non-prod

ashwindasr:
Right now, re-releases fail: https://gitlab.cee.redhat.com/hybrid-platforms/art/ocp-shipment-data/-/jobs/37856445

### DIFF
--- a/elliott/elliottlib/cli/konflux_release_cli.py
+++ b/elliott/elliottlib/cli/konflux_release_cli.py
@@ -110,8 +110,8 @@ class CreateReleaseCli:
         release_name = f"ose-{major}-{minor}-{self.release_env}-{get_utc_now_formatted_str()}"
 
         env_config: ShipmentEnv = getattr(config.shipment.environments, self.release_env)
-        if env_config.shipped():
-            LOGGER.warning(
+        if env_config.shipped() and self.release_env == "prod":
+            LOGGER.error(
                 f"existing release metadata is not empty for {self.release_env}: "
                 f"{env_config.model_dump()}. If you want to proceed remove the release metadata from the shipment config and try again."
             )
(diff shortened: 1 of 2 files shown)


elliott/elliottlib/cli/konflux_release_cli.py
```diff
@@ -110,7 +110,7 @@ async def run(self):
         release_name = f"ose-{major}-{minor}-{self.release_env}-{get_utc_now_formatted_str()}"
 
         env_config: ShipmentEnv = getattr(config.shipment.environments, self.release_env)
-        if env_config.shipped():
+        if env_config.shipped() and self.release_env == "prod":
```
thegreyd, line 113: nice! can you raise error instead of warn when env is prod? that would be a good safeguard

elliott/elliottlib/cli/konflux_release_cli.py
```diff
@@ -110,8 +110,8 @@ async def run(self):
         release_name = f"ose-{major}-{minor}-{self.release_env}-{get_utc_now_formatted_str()}"
 
         env_config: ShipmentEnv = getattr(config.shipment.environments, self.release_env)
-        if env_config.shipped():
-            LOGGER.warning(
+        if env_config.shipped() and self.release_env == "prod":
+            LOGGER.error(
```
thegreyd, line 114: mean something like

```suggestion
            raise ValueError(
```

elliott/elliottlib/cli/konflux_release_cli.py
ashwindasr: updated

elliott/elliottlib/cli/konflux_release_cli.py
ashwindasr: ahh gotcha

PR #1884, opened 2025-08-08
[sast] lower priority of requested OSH scans to 9

sfowl:
OCP scans are often requested on OSH in large batches, and can take several hours to process. This can cause delays for other OSH users, for example those looking to ship async advisories. Let's lower the priority from 10 (default) -> 9, so that OCP scans don't block other scans requestors for long periods.

CC @hpvyas - My understanding is that OSH scans for OCP are not regularly reviewed as part of the release process, so I don't expect this will have any adverse impact.


sfowl:
@ashwindasr @thegreyd could you please take a look at this?

PR #1915, opened 2025-08-19
ART-13995 Let ocp4-scan-konflux rebase into -priv

locriandev:
Ref. [ART-13995](https://issues.redhat.com/browse/ART-13995)


vfreex (commented):
/lgtm

PR #1918, opened 2025-08-20
[Konflux] Additional repo files shouldn't be deleted

vfreex:
Some images (e.g. openshift-base-rhel9) add additional repo files in their Dockerfiles. `rm -f /etc/yum.repos.d/*` unexpectedly removes those repo files.

We need to make sure that only internal build repos are removed. This PR adds `art-` prefix to internal repos and narrow the repo files to be deleted to `/etc/yum.repos.d/art-*`.

Apparently, we will need a mass rebuild to bring repo files back for all images.


doozer/doozerlib/backend/rebaser.py
```diff
@@ -1166,7 +1166,7 @@ def _add_build_repos(self, dfp: DockerfileParser, metadata: ImageMetadata, dest_
             konflux_lines += [
                 "USER 0",
                 "RUN mkdir -p /tmp/art/yum_temp; mv /etc/yum.repos.d/*.repo /tmp/art/yum_temp/ || true",
-                f"COPY .oit/{self.repo_type}.repo /etc/yum.repos.d/",
+                f"COPY .oit/art-{self.repo_type}.repo /etc/yum.repos.d/",
```
ashwindasr, line 1169: @vfreex We are writing files to `.oit/{self.repo_type}.repo`, in a different part of the codebase right? Dont that need to be updated as well?

vfreex: Makes sense. I fixed that part and will test it in ocp4-konflux job.

PR #1919, opened 2025-08-20
[Konflux] fbc:merge: Search for all stage FBC fragments

vfreex:
When no input FBC fragments are provided, search for all related tags (starting with `ocp__{MAJOR}.{MINOR}__*`) in the stage as input.

This simplifies the usage of this command, and will make sure all operators will be included in the resulting index even if we don't have a rebuild in a z-stream release.


vfreex:
@thegreyd Thanks. Fixed.

PR #1935, opened 2025-08-25
ART-10665 Enforce Konflux DB initialization for Konflux commands

locriandev:
Ref. [ART-10665](https://issues.redhat.com/browse/ART-10665)


thegreyd:
This would result in errors, there are places that we run commands with `--build-system=konflux` where we do not really need to read from DB, like shipment-ci. So I don't agree with this change.. we should not force that credentials be provided at init (if i'm fully understanding the change), and instead let commands choose to provide them / or fail if they expect builds and they can't be fetched.

PR #1938, opened 2025-08-26
removed slack warning

mbiarnes:


joepvd:
/hold
Let's not remove the alert, but remove the cause why it triggers.

PR #1941, opened 2025-08-27
Handle istag with null annotations when checking number of accepted multi nightlies

adobes1:
When pruning multi-arch nightlies, we look at existing tags to ensure at least 2 accepted nightlies remain. The `is_accepted` function crashes when processing istag that has `annotations: null`

### DIFF
--- a/doozer/doozerlib/cli/release_gen_payload.py
+++ b/doozer/doozerlib/cli/release_gen_payload.py
@@ -1679,7 +1679,8 @@ class GenPayloadCli:
                 # nightly to a new one.
                 def is_accepted(tag):
                     # Returns true if the imagestream tag has been accepted.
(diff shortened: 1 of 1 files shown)


doozer/doozerlib/cli/release_gen_payload.py
```diff
@@ -1679,7 +1679,8 @@ def add_multi_nightly_release(obj: oc.APIObject):
                 # nightly to a new one.
                 def is_accepted(tag):
                     # Returns true if the imagestream tag has been accepted.
-                    return tag.get('annotations', dict()).get('release.openshift.io/phase', 'Unknown') == 'Accepted'
+                    annotations = tag.get('annotations') or {}  # Handle `annotations: null`
```
thegreyd, line 1682:
```suggestion
                    annotations = tag.get('annotations', {})

thegreyd, line 1682: How is this different than above ?

PR #1943, opened 2025-08-28
Skip non-nightly tags when searching for last nightly

adobes1:
When deciding whether to update release controller with new multi-arch nightly, we look at the name of the last tag/nightly in the ImageStream and extract its date from its name.
This can be problematic when the last tag is not a nightly tag (as seen in [this run of build-sync-konflux](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fbuild-sync-konflux/6076/consoleFull)).

To prevent this, check first if the last tag really is a nightly


joepvd (approved):
/lgtm

PR #1948, opened 2025-08-28
Add mirroring to quay.io/openshift/ci (QCI) to images that should be used in CI

jupierce:
Test Platform is moving to an architecture where images are no longer used from their app.ci internal registry and instead always pulled from quay.io . The images we mirror to app.ci, therefore, specifically those that get added to the 4.x imagestream, must also show up in quay.io/openshift/ci . When CI promotes an image from an upstream build, this happens automatically. However, for images that only ART builds, we need to upload to QCI directly. An example is the 'base-rhel9' image, which gets added to the 4.x imagestreams and whose sha256 digest CI will try to resolve in quay.io/openshift/ci . 

We must have write credentials for QCI for this to work. `openshift+art_to_ci_promotion_robot` was created for this purpose, but I'm not certain it is in our credential store.

### DIFF
--- a/doozer/doozerlib/cli/images_streams.py
+++ b/doozer/doozerlib/cli/images_streams.py
@@ -5,6 +5,7 @@ import os
 import random
 import re
 import time
+from datetime import datetime
 from typing import Dict, Optional, Set, Tuple
 
 import click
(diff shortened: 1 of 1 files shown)


doozer/doozerlib/cli/images_streams.py
```diff
@@ -108,6 +109,46 @@ def images_streams_mirror(runtime, streams, only_if_missing, live_test_mode, for
...
     runtime.assert_mutation_is_permitted()
 
+    def mirror_image(cmd_start: str, upstream_dest: str):
+        full_cmd_1 = f'{cmd_start} {upstream_dest}'
+        if dry_run:
+            print(f'For {upstream_entry_name}, would have run: {full_cmd_1}')
+        else:
+            exectools.cmd_assert(full_cmd_1, retries=3, realtime=True)
+
+        if upstream_dest.startswith('registry.ci.openshift.org/ocp'):
```
joepvd, line 119: Probably also want to include `registry.ci.openshift.org/release` for the build root images.

deepsm007: Can we also pull the build root images from QCI as well?

jupierce:
/unhold
@locriandev has gotten the secrets injected for this job.
https://github.com/openshift-eng/art-tools/pull/1950
https://github.com/openshift-eng/aos-cd-jobs/pull/4465

PR #1951, opened 2025-08-29
ART-13996 Exclude mismatched siblings from payload update for pre-releases

locriandev:
Ref. [ART-13996](https://issues.redhat.com/browse/ART-13996)

### DIFF
--- a/doozer/doozerlib/cli/release_gen_payload.py
+++ b/doozer/doozerlib/cli/release_gen_payload.py
@@ -419,6 +419,9 @@ class GenPayloadCli:
         self.payload_permitted = False
         # Allows embargoed builds to be released
         self.embargo_permit_ack = embargo_permit_ack
+        # Track mismatched siblings.
+        # This will be used to prevent syncing out mismatched siblings for development releases per ART-13996
+        self.mismatched_siblings = []
 
     @start_as_current_span_async(TRACER, "releases:gen-payload")
     async def run(self):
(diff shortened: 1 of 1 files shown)


doozer/doozerlib/cli/release_gen_payload.py
```diff
@@ -562,12 +565,20 @@ def detect_mismatched_siblings(self, assembly_inspector: AssemblyInspector):
         group_images: List = list(assembly_inspector.get_group_release_images().values())
         issues = []
         for mismatched, sibling in self.payload_generator.find_mismatched_siblings(group_images):
+            component = mismatched.get_image_meta().distgit_key
+
+            if self.runtime.group_config.software_lifecycle.phase == 'pre-release':
```
jupierce, line 570: I think I would tweak this to have our ocp-build-config have `MISMATCHED_SIBLINGS` in our new pre-release permits. The semantics would be: if MISMATCHED_SIBLINGS is permitted, we will not reject the payload BUT we will also not sync the mismatched set.
The semantics would be the same on GA z-streams as well, if the permit was ever added there for some reason. This would eliminate the need to have these special checks for `phase` == `pre-release` . 

If we really want the ability to force mismatch sets to be synced, it could be another permit like `MISMATCHED_SIBLINGS_SYNC`. But I'm not sure it is worth implementing this.

jupierce:
A nit in the log message, but otherwise
/lgtm

PR #1952, opened 2025-08-29
ART-14030 find-bugs:golang konflux nightly support, bug fixes, reporting and logging enhancements

thegreyd:
https://issues.redhat.com/browse/ART-14030

Summary
- Support looking up go versions of konflux nvrs
- When analyzing ART built rpms: consider only those nvrs that belong to rhel targets that are currently building.
- Add `--csv-output` flag to dump report as csv at the end
- Add `--jql-filter` param to enable using an existing jql filter e.g. "All Open Black CVEs"
- Better logging
- Better reporting on expected and actual values
- Only look at `component = Release` bugs


joepvd (approved):
/lgtm

PR #1963, opened 2025-09-04
WIP: Use new-style plashet configs if available

vfreex:
Use new-style plashet configs if available https://github.com/openshift-eng/ocp-build-data/pull/7492/files.

If not existing, fall back to the old hardcoded configs in pyartcd.


vfreex:
PR rebased

PR #1978, opened 2025-09-10
konflux build priority

ashwindasr:


pyartcd/pyartcd/pipelines/ocp4_konflux.py
```diff
@@ -97,6 +99,12 @@ def __init__(
         self.plr_template = plr_template
         self.lock_identifier = lock_identifier
         self.skip_plashets = skip_plashets
+        self.build_priority = build_priority
+        self.use_mass_rebuild_locks = use_mass_rebuild_locks
+
+        # Fail fast if both build_priority and use_mass_rebuild_locks are set
+        if build_priority and use_mass_rebuild_locks:
+            raise ValueError("Cannot specify both --build-priority and --use-mass-rebuild-locks. These options are mutually exclusive.")
```
jupierce, line 107: Not sure I agree here. Even if we have mass rebuild locks, being able to trigger a single image build at a higher priority has value.

PR #1993, opened 2025-09-12
Allow commands to run without kubeconfig

thegreyd:
Previously konflux recomended users to login in via cli using a kubeconfig w oidc/toolchain
Recently that got changed, so stop enforcing that a kubeconfig is provided when running commands.

https://konflux.pages.redhat.com/docs/users/getting-started/getting-access-new.html#logging-in-using-oc-login-web
https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1757488512749149


vfreex (commented):
/lgtm

PR #2021, opened 2025-09-19
fail build if source image signature is missing

ashwindasr:
The konflux build pipeline has a step called build source image. The source image is built and signed as part of that step, but last week there was build that was not signed due to transient error on the konflux's side, which caused us a significant delay to promote the payload.

To prevent that, check to see if the source image is signed or not. If its signed, mark the build as failed, to trigger a rerun.

Test build: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-konflux/17240

Seeing
```
2025-09-19 16:49:56,691 art_tools.artcommonlib.exectools INFO Executing:cmd_gather_async: cosign download signature quay.io/redhat-user-workloads/ocp-art-tenant/art-images:sha256-c581e65aafeb86d8cd3d9fecd2858118a669c5e16ed723d52c13e09f01ba33ad.src@sha256:e939712b91cd9f51128bd09a9d5a76749baeefbb7527647b55b3e4ed08874e3a
```

### DIFF
--- a/artcommon/artcommonlib/util.py
+++ b/artcommon/artcommonlib/util.py
@@ -1,10 +1,13 @@
 import asyncio
+import base64
+import json
 import logging
 import os
 import re
 import sys
 from datetime import date, datetime, timedelta, timezone
 from functools import lru_cache
+from json import JSONDecodeError
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


joepvd (approved):
/lgtm

PR #2040, opened 2025-09-25
group option for bundle and fbc

ashwindasr:
OADP is being onboarded to ART and they live on seperate branches in ocp-build-data: https://github.com/openshift-eng/ocp-build-data/tree/oadp-1.3

Allow the bundle and FBC build jobs to pass in a group option to let the new operators build, since they are not related to openshift

Alongside: https://github.com/openshift-eng/aos-cd-jobs/pull/4480


vfreex (approved):
/lgtm

PR #2056, opened 2025-10-02
use safeformatter for docs replacement

ashwindasr:
similar to https://github.com/openshift-eng/art-tools/pull/1985

use SafeFormatter to let Elliott ignore those missing variables during variable substitution and not panic.


locriandev (commented):
/lgtm

PR #2061, opened 2025-10-06
support for oadp / logging

ashwindasr:
Update codepath to support non-OCP components

### DIFF
--- a/doozer/doozerlib/backend/konflux_fbc.py
+++ b/doozer/doozerlib/backend/konflux_fbc.py
@@ -737,7 +737,7 @@ class KonfluxFbcRebaser:
             replaces = None
             if 'oadp-' in olm_bundle_name:
                 # Find the current head - the entry that is not replaced by any other entry
-                bundle_with_replaces = [it for it in channel['entries'] if 'oadp-' in it['name']]
+                bundle_with_replaces = [it for it in channel['entries']]
                 replaced_names = {it.get('replaces') for it in bundle_with_replaces if it.get('replaces')}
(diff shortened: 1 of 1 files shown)


doozer/doozerlib/backend/konflux_fbc.py
```diff
@@ -732,6 +732,18 @@ def _update_channel(channel: Dict):
                 skips = set(bundle_with_skips.pop('skips'))
                 skips = (skips | {bundle_with_skips['name']}) - {olm_bundle_name}
 
+            # For an operator bundle that uses replaces -- such as OADP
+            # Update "replaces" in the channel
+            replaces = None
+            if 'oadp-' in olm_bundle_name:
+                # Find the current head - the entry that is not replaced by any other entry
+                bundle_with_replaces = [it for it in channel['entries'] if 'oadp-' in it['name']]
```
rayfordj, line 740:
```suggestion
                bundle_with_replaces = [it for it in channel['entries']]
```
I would need to test to confirm this, but since we're already in the `if ... in olm_bundle_name:`, it leads me to think we could safely remove `if ... in it['name']` here because it will always match when the outer-if == True.  I didn't catch this when I originally worked up this change.  The expected benefit to removing it here is that any additional operator that makes use of replaces: will only need to be added to the [outer-if](https://github.com/openshift-eng/art-tools/pull/2061/files#diff-30b674d7a7ea084ce01ae24b093315b2fcce1f81367c503bd2de55d8e77eada1R738).

_Based only on visual recognition and without actually validating this, I'm fine with it as it is for now because we know it works for OADP's use of `replaces:`._

PR #2062, opened 2025-10-07
Improve cachi2 state logging

locriandev:


lgarciaaco:
what about log only once at the end?, something like:
```  def is_cachi2_enabled(self):
      """
      Determine if cachi2 is enabled or not
      image config override > group config override > fallback to cachito config
      """

      cachi2_config_override = self.config.konflux.cachi2.enabled
      cachi2_group_override = self.runtime.group_config.konflux.cachi2.enabled

      if cachi2_config_override not in [Missing, None]:
          cachi2_enabled = cachi2_config_override
          source = "metadata config"
      elif cachi2_group_override not in [Missing, None]:
          cachi2_enabled = cachi2_group_override
          source = "group config"
      else:
          cachi2_enabled = artlib_util.is_cachito_enabled(
              metadata=self, group_config=self.runtime.group_config, logger=self.logger
          )
          source = "cachito fallback"

      self.logger.info(f"cachi2 enabled: {cachi2_enabled} (from {source})")
      return cachi2_enabled
```
    
Also, `is_lockfile_generation_enabled` and `is_lockfile_force_enabled` ... and `is_lockfile_parent_inspect_enabled` follow the same pattern, maybe makes sense to change there for consistency.

/lgtm

PR #2690, opened 2026-03-31
feat(lockfile): implement RPM version fallback recovery for hermetic …

lgarciaaco:
…builds

Cross-architecture RPM version mismatches in lockfile generation were causing build failures when converting images to hermetic mode. This implements intelligent fallback recovery that automatically selects the lowest common version when mismatches occur, allowing builds to continue successfully.

The fallback system handles Case 1 scenarios where the target version already exists in all architecture collections, gracefully degrading to the original error behavior for Case 2 scenarios requiring repository resolution. Only mismatched packages are modified while maintaining architecture consistency.

rh-pre-commit.version: 2.3.2

### DIFF
--- a/doozer/doozerlib/lockfile.py
+++ b/doozer/doozerlib/lockfile.py
@@ -4,7 +4,7 @@ from dataclasses import dataclass
 from functools import total_ordering
 from logging import Logger
 from pathlib import Path
-from typing import Dict, Optional, Tuple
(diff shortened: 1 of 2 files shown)


thegreyd:
Have a couple of questions regarding this here: https://redhat-internal.slack.com/archives/C07RAB68T5G/p1775233733280559?thread_ts=1774546650.764209&cid=C07RAB68T5G

PR #2728, opened 2026-04-10
Use image digest when using cosign download

thegreyd:
Seen at https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fgolang-builder/381/console

This command is run for both the image variants.

```
cosign download sbom quay.io/redhat-user-workloads/ocp-art-tenant/art-images:golang-builder-v1.20.12-20260410.110116 --platform linux/amd64
```

uuid tag for golang builder image can collide for rhel8 and rhel9 builders. So instead use image digest

for more context: this is because our images use openshift-golang-builder.yml in all branches. In future including rhel info in the image name would be the way to go.


ashwindasr (commented):
/unhold

PR #2754, opened 2026-04-13
Log git stderr when build_repo push fails during rebase

fbladilo:
## Summary
- The `push` method in `build_repo.py` used `run_git_async` which discards stderr, so push failures only showed "exited with code 1" with no details
- Switch to `gather_git_async` to capture stdout/stderr and log them on failure
- Makes it possible to diagnose issues like missing branches in openshift-priv or permission errors

## Test plan
- Trigger a build where `git push` to openshift-priv fails (e.g. missing branch)
- Verify Jenkins log now shows the actual git error message instead of just the exit code


fbladilo:
Test : https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Flayered-products/2203/

PR #2724, opened 2026-04-09
ART-14323: Add enterprise-contract verification after Konflux image builds

ashwindasr:
## Summary
- After a successful Konflux image build + SLSA attestation validation, runs an enterprise-contract (EC) policy check
- Ensures an `IntegrationTestScenario` (ITS) exists for the application, then creates an EC PipelineRun to verify the built image against the stage EC policy
- EC policy is selected based on assembly type and image type:
  - **Base images** (`base_only=true`): use [`registry-ocp-art-base-prod`](https://gitlab.cee.redhat.com/releng/konflux-release-data/-/blob/main/config/kflux-ocp-p01.7ayg.p1/product/EnterpriseContractPolicy/registry-ocp-art-base-prod.yaml) (no preGA variant)
  - **FBC builds**: use `rhtap-releng-tenant/fbc-ocp-art-stage` (no preGA variant)
  - **Bundle builds**: same logic as regular images — `registry-ocp-art-stage` (default) or `registry-ocp-art-ec-stage` (preGA/PREVIEW assemblies)
  - **Regular images**: **PREVIEW** (preGA) assemblies use [`registry-ocp-art-ec-stage`](https://gitlab.cee.redhat.com/releng/konflux-release-data/-/blob/main/config/kflux-ocp-p01.7ayg.p1/product/EnterpriseContractPolicy/registry-ocp-art-ec-stage.yaml) (allows unsigned RPMs), all others use `registry-ocp-art-stage`
- EC verification is **blocking** — build fails if EC check fails
- Adds `--skip-ec-verify` flag to `beta:images:konflux:build` to opt out
- Currently scoped to **OCP only**; layered products (logging, oadp, mta, etc.) to be expanded later
- Golang builder builds skip EC verification (`--skip-ec-verify`) since they use non-standard assembly conventions

## Changes
- `doozer/doozerlib/constants.py`: EC pipeline URL, policy constants, preGA policy constant, **new** `KONFLUX_BASE_IMAGE_EC_POLICY_CONFIGURATION` and `KONFLUX_FBC_EC_POLICY_CONFIGURATION` constants
- `doozer/doozerlib/backend/konflux_client.py`: New `IntegrationTestScenario` and EC `PipelineRun` manifest builders + orchestration methods; **new** shared `verify_enterprise_contract()` helper and `ECVerificationResult` dataclass to avoid duplicating ITS+PLR+wait logic across builders
[...]

PR #2756, opened 2026-04-13
Add group.yml toggle for EC verification failure gating

joepvd:
## Summary
- Adds `konflux.ec_failure_is_error` boolean toggle to group.yml (defaults to `false`)
- When `false`: EC verification runs and results are recorded (`ec_status`, `ec_pipeline_url`) but failures do **not** break the build
- When `true`: EC failure fails the build and skips retries (original behavior from #2724)
- Updated schema, all three builders (image, FBC, bundle), and tests

## Motivation
We want to collect EC verification results across builds before enforcing failures. This toggle lets us observe without risk, then flip to `true` per-group when ready.

## Test plan
- [x] `test_ec_failure_does_not_fail_build_by_default` — EC fails, build succeeds
- [x] `test_no_retry_when_ec_fails` — with `ec_failure_is_error: true`, EC failure breaks the build
- [x] All existing EC verification tests pass
- [x] Lint clean

### DIFF
--- a/doozer/doozerlib/backend/konflux_fbc.py
+++ b/doozer/doozerlib/backend/konflux_fbc.py
@@ -1892,8 +1892,19 @@ class KonfluxFbcBuilder:
                         ec_status = ec_result.ec_status
                         ec_pipeline_url = ec_result.ec_pipeline_url
                         if ec_result.ec_failed:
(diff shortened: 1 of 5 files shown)


joepvd:
/hold
Let's only merge if we must

PR #2758, opened 2026-04-13
Enable EC/ITS verification for layered products after Konflux builds

ashwindasr:
## Summary

- Expand EC (Enterprise Contract) verification via ITS (IntegrationTestScenario) from OCP-only to layered products (logging, mta, rhmtc, oadp)
- Replace `group_name.startswith("openshift-")` gates with product-based lookup against `PRODUCT_EC_POLICY_MAP` in all three builders (image, OLM bundle, FBC)
- Each product resolves to its own `EnterpriseContractPolicy` from `konflux-release-data` (e.g. `registry-art-logging-stage` for logging, `fbc-art-oadp-stage` for OADP FBC). Unknown products gracefully skip EC with an info log.

## Changes

- `doozer/doozerlib/constants.py`: Add `PRODUCT_EC_POLICY_MAP`, `PRODUCT_FBC_EC_POLICY_MAP`, `get_ec_policy_for_product()`, `get_fbc_ec_policy_for_product()`
- `doozer/doozerlib/backend/konflux_image_builder.py`: Replace `is_ocp_group` with product-based check; remove unused `ec_policy_configuration` config field
- `doozer/doozerlib/backend/konflux_olm_bundler.py`: Replace `is_ocp_group` with product-based check
- `doozer/doozerlib/backend/konflux_fbc.py`: Use `get_fbc_ec_policy_for_product(self.product)` instead of hardcoded OCP FBC policy
- `doozer/tests/backend/test_konflux_ec_verification.py`: Add tests for layered product EC (logging, oadp), unknown product skip, and policy helper functions

## Test plan

- [x] All 18 EC verification tests pass (8 gating + 10 policy helper)
- [x] All 24 related backend tests pass (konflux_client + image_builder)
- [x] `make lint` passes
- [ ] Verify in staging that layered product builds (e.g. logging-6.x) trigger EC with correct policy

### DIFF
--- a/doozer/doozerlib/backend/konflux_fbc.py
+++ b/doozer/doozerlib/backend/konflux_fbc.py
@@ -1862,8 +1862,39 @@ class KonfluxFbcBuilder:
                 succeeded_condition = pipelinerun_info.find_condition('Succeeded')
                 outcome = KonfluxBuildOutcome.extract_from_pipelinerun_succeeded_condition(succeeded_condition)
 
+                ec_failed = False
                 ec_status = KonfluxECStatus.NOT_APPLICABLE
                 ec_pipeline_url = ''
+
+                has_fbc_ec_policy = self.product in constants.PRODUCT_FBC_EC_POLICY_MAP
+                if outcome is KonfluxBuildOutcome.SUCCESS and has_fbc_ec_policy:
+                    results = pipelinerun_dict.get('status', {}).get('results', [])
+                    image_pullspec = next((r['value'] for r in results if r['name'] == 'IMAGE_URL'), None)
+                    image_digest = next((r['value'] for r in results if r['name'] == 'IMAGE_DIGEST'), None)
+
+                    if image_pullspec and image_digest:
+                        app_name = self.get_application_name(self.group)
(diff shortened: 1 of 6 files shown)


ashwindasr:
/hold

until test builds are green

fbladilo:
**Fallback behavior in policy helpers is unreachable dead code**

The helper functions `get_ec_policy_for_product()` and `get_fbc_ec_policy_for_product()` fall back to the OCP default policy for unknown products:

```python
def get_ec_policy_for_product(product: str) -> str:
    policy_name = PRODUCT_EC_POLICY_MAP.get(product)
    if policy_name:
        return f"{KONFLUX_EC_POLICY_NAMESPACE}/{policy_name}"
    return KONFLUX_DEFAULT_EC_POLICY_CONFIGURATION  # fallback
```

But all callers guard with `product in constants.PRODUCT_EC_POLICY_MAP` before invoking the helper, so the fallback path can never be reached in practice. This makes the behavior a bit misleading — the helpers suggest they gracefully handle unknown products, but that case is already filtered out upstream.

Consider either raising a `ValueError` for unknown products (since callers already ensure the product is in the map), or adding a brief comment noting the fallback is purely defensive.

PR #2765, opened 2026-04-14
Skip redundant assembly consistency checks in multi-model build sync

joepvd:
## Summary
- In the multi-model (`build-sync-multi`) path, `release:gen-payload` was running the full suite of assembly consistency checks (`generate_assembly_report`) even though:
  1. The model nightly (x86_64) already passed these checks during regular `build-sync`
  2. `emergency_ignore_issues` is hardcoded to `True` in multi-model mode, so the checks can never block anything
- These checks are expensive (Brew/Konflux RPM queries, hermetic checks, sibling matching, etc.) and can flake, causing unnecessary failures
- This change skips `generate_assembly_report()` in multi-model mode and writes a minimal report instead. The essential `generate_payload_entries()` and `sync_payloads()` flow is unchanged.

## Test plan
- [x] Existing `test_gen_payload.py` tests pass (50/50)
- [ ] Verify a `build-sync-multi` run completes successfully and produces the multi-arch nightly

### DIFF
--- a/doozer/doozerlib/cli/release_gen_payload.py
+++ b/doozer/doozerlib/cli/release_gen_payload.py
@@ -550,7 +550,21 @@ class GenPayloadCli:
         self.payload_entries_for_arch, self.private_payload_entries_for_arch = await self.generate_payload_entries(
             assembly_inspector
         )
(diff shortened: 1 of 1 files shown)


shruti-rh (approved):
lgtm

PR #2771, opened 2026-04-15
Temporarily disable post-build SLSA attestation validation

ashwindasr:
## Summary
- Post-build SLSA attestation validation via `cosign download attestation` is consistently failing in the build environment, despite the attestation being available (verified locally)
- This is likely an auth/environment issue in the build pod, not a timing issue
- Temporarily comments out the attestation validation block to unblock builds while the root cause is investigated
- Adds retry logic (5 attempts, 60s backoff) for when the validation is re-enabled
- Improves error handling: parse errors now fail fast instead of retrying, and `ChildProcessError` logging now includes cosign stderr for debugging

## Test plan
- [x] Unit tests pass (`doozer/tests/backend/test_konflux_image_builder.py`, `doozer/tests/cli/test_scan_sources_konflux.py`)
- [ ] Verify builds complete without being marked as failure due to attestation issues
- [ ] Investigate cosign auth in build environment and re-enable validation

### DIFF
--- a/artcommon/artcommonlib/util.py
+++ b/artcommon/artcommonlib/util.py
@@ -686,10 +686,6 @@ async def fetch_slsa_attestation(
         LOGGER.warning(f'Failed to fetch SLSA attestation for {build_name}')
         return None
 
-    except (JSONDecodeError, Exception) as e:
-        LOGGER.warning('Failed to parse SLSA attestation for %s: %s', build_name, e)
-        return None
-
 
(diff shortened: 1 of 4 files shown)


ashwindasr:
/hold

might not be needed https://redhat-internal.slack.com/archives/C07RAB68T5G/p1776227291237049?thread_ts=1776198592.444959&cid=C07RAB68T5G

PR #2774, opened 2026-04-15
Use auth file from jenkins cred instead of auth.json

thegreyd:
Changed manifest_tool to require an explicit auth_file parameter instead of discovering auth from the environment or ${XDG_RUNTIME_DIR}/containers/auth.json.

Updated the only real call sites to pass the auth file explicitly:
- doozer release_gen_payload manifest-list pushes now pass os.getenv("QUAY_AUTH_FILE")
- pyartcd promote manifest-list push now passes QUAY_AUTH_FILE

Fixed jenkins wiring: https://github.com/openshift-eng/aos-cd-jobs/pull/4654

Also updated tests in artcommon, pyartcd, and doozer to cover the explicit auth-file flow and removal of the old fallback behavior.

rh-pre-commit.version: 2.3.2

### DIFF
--- a/artcommon/artcommonlib/exectools.py
+++ b/artcommon/artcommonlib/exectools.py
@@ -17,7 +17,6 @@ from datetime import datetime
 from fcntl import F_GETFL, F_SETFL, fcntl
 from inspect import getframeinfo, stack
 from multiprocessing.pool import MapResult, ThreadPool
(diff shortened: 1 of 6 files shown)


shruti-rh (approved):
lgtm

locriandev (approved):

PR #2780, opened 2026-04-16
Pass QUAY_AUTH_FILE to oc image info for RHCOS pullspec lookups

locriandev:
The elliott find-builds --only-rhcos command calls oc image info without registry credentials, causing auth failures when the default credentials on the Jenkins node are stale. Read QUAY_AUTH_FILE from the environment (already provided by Jenkins withCredentials) and pass it through to get_build_id_from_rhcos_pullspec(), matching the existing pattern in elliottlib/util.py.


lgarciaaco (commented):
/lgtm

PR #2790, opened 2026-04-16
fix(lockfile): preserve NVR-pinned RPMs in hermetic lockfile generation

rayfordj:
PR #2746 introduced a highest-EVR-per-package-name strategy in _fetch_rpms_info_per_arch that silently discarded NVR-pinned RPMs (e.g. rust-toolset-1.84.1-1.el9) when a newer version existed, causing hermetic builds to fail for images like logging vector that pin specific toolchain versions.

Add a second tracking dict (pinned_by_nvr) alongside best_by_name so that explicitly pinned NVR entries are always included in the lockfile output. Plain package names still resolve to the highest EVR across all repos (preserving the runc/systemd-udev fix from #2746). The two dicts are merged with (name, evr) deduplication to avoid duplicates when the pinned version happens to also be the latest.

Also convert Repodata._detect_nvr_vs_name to @staticmethod so lockfile.py can call it without a Repodata instance.

Signed-off-by: Rayford Johnson <rayfordj@users.noreply.github.com>

rh-pre-commit.version: 2.3.2

### DIFF
--- a/doozer/doozerlib/lockfile.py
+++ b/doozer/doozerlib/lockfile.py
@@ -276,9 +276,12 @@ class RpmInfoCollector:
         """
         Resolve RPM metadata for a specific architecture from the given repodata names.
 
-        For each package, the highest version across ALL repos is selected (matching dnf
-        behaviour). When the same EVR exists in multiple repos, the repo iteration order
-        from ``sort_repos_for_lockfile_resolution`` acts as tiebreaker (RHEL upstream repos
+        For plain package names the highest version across ALL repos is selected (matching
+        dnf behaviour).  For NVR-pinned items the exact requested version is preserved in
+        the output so that hermetic builds can install a specific older version.
+
+        When the same EVR exists in multiple repos, the repo iteration order from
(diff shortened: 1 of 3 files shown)


rayfordj:
/hold

Unblock Logging's vector builds.

[build%252Flayered-products/2460/](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Flayered-products/2460/)

[logging-6-5/pipelineruns/logging-6-5-vector-g5dnm/](https://konflux-ui.apps.kflux-ocp-p01.7ayg.p1.openshiftapps.com/ns/art-logging-tenant/applications/logging-6-5/pipelineruns/logging-6-5-vector-g5dnm/)


:information_source:  I've since come to discover that other layered-products (such as MTC) which use modules are also affected by pull/2746.  This PR does **not** address modules _(at the moment)._

rayfordj:
/unhold
/approve

Will address `dnf modules` in another PR.  Unblocking Logging with this one merging since they have release planned soon.

PR #2797, opened 2026-04-17
Fix NVR detection misparsing package names with digit-letter suffixes

fbladilo:
## Summary

- Package names like `xorg-x11-fonts-ISO8859-1-100dpi` were incorrectly identified as NVRs because `rsplit("-", 2)` produced `version="1"` and `release="100dpi"`, both passing the loose digit-based validation in `_detect_nvr_vs_name()`
- Reject release fields matching `^\d+[a-zA-Z]+$` (digits-then-letters only) since real RPM releases always contain separators (dots, underscores) or are purely numeric
- Add fallback in `get_rpms()`: if the NVR-extracted name matches nothing in the repo, retry with the original string as a plain package name

This change is backwards compatible and should not alter current behavior. All existing test cases pass unchanged, and new tests cover the fix.

## Context

Observed in [build/layered-products #2525](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Flayered-products/2525/console) where `xorg-x11-fonts-ISO8859-1-100dpi` was silently dropped from the cachi2 lockfile, causing a hermetic build failure when `dnf builddep` could not resolve the `graphviz` dependency.

## Test plan

- [ ] All existing `test_repodata.py` tests pass (38/38)
- [ ] New test `test_get_rpms_nvr_false_positive_fallback` verifies the `get_rpms()` safety net
- [ ] New test cases in `test_detect_nvr_vs_name_package_names` and `test_detect_nvr_vs_name_edge_cases` cover the regex fix
- [ ] Full doozer test suite passes (873 passed)
- [ ] Lint clean

### DIFF
--- a/doozer/doozerlib/repodata.py
+++ b/doozer/doozerlib/repodata.py
@@ -209,6 +209,16 @@ class Repodata:
                 rpm for rpm in self.primary_rpms if rpm.name == rpm_name and (rpm.arch == arch or rpm.arch == 'noarch')
             ]
 
+            if not matching_rpms and is_nvr and rpm_name != item:
+                # NVR detection may have false-positived on a package name that
+                # contains digit-heavy segments (e.g. xorg-x11-fonts-ISO8859-1-100dpi).
+                # Retry with the original string as a plain package name.
+                matching_rpms = [
+                    rpm for rpm in self.primary_rpms if rpm.name == item and (rpm.arch == arch or rpm.arch == 'noarch')
+                ]
+                if matching_rpms:
+                    is_nvr = False
+
             if not matching_rpms:
                 not_found.append(item)
                 continue
@@ -251,7 +261,11 @@ class Repodata:
                 if re.search(r'[\d.]', version):
                     # Validate release: should contain at least one digit (typical RPM release)
                     if re.search(r'\d', release):
-                        return True, extracted_name
+                        # Reject releases that are just digits followed by letters (e.g. "100dpi")
+                        # — these are part of package names like xorg-x11-fonts-ISO8859-1-100dpi,
+                        # not real RPM releases which contain separators like dots or underscores
+                        if not re.fullmatch(r'\d+[a-zA-Z]+', release):
+                            return True, extracted_name
         except Exception:
             # If parsing fails, treat as package name
             pass
--- a/doozer/tests/test_repodata.py
+++ b/doozer/tests/test_repodata.py
@@ -666,6 +666,27 @@ data:
         self.assertEqual(found, [])
         self.assertEqual(not_found, ["foo-invalid-nvr-format"])
 
+    def test_get_rpms_nvr_false_positive_fallback(self):
+        # When NVR detection false-positives on a package name like
+        # xorg-x11-fonts-ISO8859-1-100dpi, get_rpms should fall back
+        # to treating the full string as a plain package name.
+        font_rpm = Rpm(
+            name="xorg-x11-fonts-ISO8859-1-100dpi",
+            epoch=0,
+            version="7.5",
+            release="21.el8",
+            arch="noarch",
+            checksum="fff",
+            size=1000,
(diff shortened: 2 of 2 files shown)


fbladilo:
**Verified with Jenkins test build [#2528](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Flayered-products/2528/) (`ART_TOOLS_COMMIT=fbladilo@fix-nvr-detection-digit-letter-release`, `GROUP=mtc-1.8`, `IMAGE_LIST=openshift-migration-openvpn`).**

Results:
- `xorg-x11-fonts-ISO8859-1-100dpi` is no longer in the "Could not find" warning
- The generated `rpms.lock.yaml` now correctly includes the package:
```yaml
- name: xorg-x11-fonts-ISO8859-1-100dpi
  evr: 0:7.5-19.el8
  checksum: sha256:460d123daa11f00f22164e841d49886b8852fef5a7ebf08d6c4f2b719fe8b43f
  repoid: rhel-8-for-x86_64-appstream-eus-rpms
  size: 1109848
  sourcerpm: xorg-x11-fonts-7.5-19.el8.src.rpm
  url: https://rhsm-pulp.corp.redhat.com/content/dist/rhel8/8/x86_64/appstream/os/Packages/x/xorg-x11-fonts-ISO8859-1-100dpi-7.5-19.el8.noarch.rpm
```
- Diffing the lockfile against the previous build (2525) shows this is the **only** change — no drift or side effects on any other package.

doozer/doozerlib/repodata.py
```diff
@@ -251,7 +261,11 @@ def _detect_nvr_vs_name(item: str) -> Tuple[bool, str]:
                 if re.search(r'[\d.]', version):
                     # Validate release: should contain at least one digit (typical RPM release)
                     if re.search(r'\d', release):
-                        return True, extracted_name
+                        # Reject releases that are just digits followed by letters (e.g. "100dpi")
+                        # — these are part of package names like xorg-x11-fonts-ISO8859-1-100dpi,
+                        # not real RPM releases which contain separators like dots or underscores
+                        if not re.fullmatch(r'\d+[a-zA-Z]+', release):
```
thegreyd, line 267: why not query brew, since all rpm nvrs should be built in brew?

fbladilo: I'm trying to avoid anything brew in purpose to avoid creating future problems

lgarciaaco:
@fbladilo can you try to include this case in the first loop iteration, maybe something like this work
```
            matching_rpms = [
                rpm for rpm in self.primary_rpms if (rpm.name == rpm_name or rpm.name == item) and (rpm.arch == arch or rpm.arch == 'noarch')
            ]
```

fbladilo:
> @fbladilo can you try to include this case in the first loop iteration, maybe something like this work
> ```
> matching_rpms = [
>     rpm for rpm in self.primary_rpms if (rpm.name == rpm_name or rpm.name == item) and (rpm.arch == arch or rpm.arch == 'noarch')
> ]
> ```

I considered combining both lookups into a single pass, but the issue is that just adding `or rpm.name == item` to the filter isn't enough — we also need to flip `is_nvr = False` when the match comes from `item` rather than `rpm_name`.

Without that flip, the downstream code still takes the NVR path (tries to match an exact version-release), which fails because there's no real version/release to match — and the package still ends up in `not_found`.

Once you add the logic to detect which name actually matched and conditionally reset `is_nvr`, it's no simpler than the current two-phase approach. So I kept them separate to make the intent explicit: try the NVR-parsed name first, and only if that yields nothing, fall back to treating the full string as a plain package name.

doozer/doozerlib/repodata.py
thegreyd: That's fair, but if there are more false positives - we can revisit this or some other robust approach.

thegreyd:
/approve
/hold 
please unhold when this is ready to merge

PR #2804, opened 2026-04-21
ART-14961: add RegistryConfig for explicit credential control

locriandev:
Builds on #2794 — adds kubeconfig support to `RegistryConfig`, scopes adoption to `ocp4-konflux` only, and consolidates Konflux registry constants.

### RegistryConfig: kubeconfig support
- Added `kubeconfig` parameter to `RegistryConfig` — accepts a file path (e.g. from `os.environ.get('KUBECONFIG')`)
- On `__enter__`, runs `oc --kubeconfig <path> registry login --to=<tempfile>` to obtain CI registry credentials
- The temp file is loaded as an additional source, then immediately deleted after credential extraction
- This eliminates the need for callers to manually run `oc registry login` and manage intermediate auth files
- 9 new tests covering kubeconfig flows (combined with source files, credentials, error handling, cleanup)

### ocp4-konflux pipeline cleanup
- Replaced manual `oc.registry_login` / `oc.qci_registry_login` boilerplate with `RegistryConfig(kubeconfig=...)`
- QCI credentials now use explicit `RegistryCredential` instead of `oc.qci_registry_login()`
- Stopped overwriting `REGISTRY_AUTH_FILE` and `QUAY_AUTH_FILE` env vars — `os.environ` is read-only input; only `DOCKER_CONFIG` is set (required by cosign)
- Auth file path stored as `self._registry_auth_file` for use by `mirror_streams_to_ci()`

### Scoped to ocp4-konflux only
- Reverted all changes to `build_sync`, `build_sync_multi`, `build_microshift`, and `ocp4_scan_konflux` — those pipelines still use `QUAY_AUTH_FILE` env var flows consumed by doozer's `release:gen-payload`

### Konflux registry constants consolidated
- Renamed `KONFLUX_ART_IMAGES` → `KONFLUX_DEFAULT_IMAGE_REPO`, `KONFLUX_ART_IMAGES_SHARE` → `KONFLUX_DEFAULT_IMAGE_SHARE_REPO`, `KONFLUX_ART_FBC` → `KONFLUX_DEFAULT_FBC_REPO` to align with the pre-existing `KONFLUX_DEFAULT_IMAGE_REPO` in `doozerlib/constants.py`
- `doozerlib/constants.py` now imports from `artcommonlib` instead of duplicating the strings
- Replaced hardcoded `quay.io/redhat-user-workloads/ocp-art-tenant/art-images` with the constant in `artcommonlib/util.py` and `pyartcd/pipelines/update_golang.py`

## Test plan
- [x] All 85 `test_registry_config.py` tests pass (including 9 new kubeconfig tests)
- [x] All 446 pyartcd tests pass
- [x] All 1173 artcommon + doozer tests pass
- [x] Lint clean across all changed files
- [x] Tested with ocp4-konflux pipeline run ([build #32017](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-konflux/32017/))

### DIFF
--- a/artcommon/artcommonlib/registry_config.py
+++ b/artcommon/artcommonlib/registry_config.py
@@ -3,7 +3,8 @@ Container registry credential manager for oc / Podman-style tools.
 
 RegistryConfig creates a temporary Docker config.json file containing only
 the specified registry credentials, built by cherry-picking entries from
-source auth files and/or providing explicit username:password pairs.
+source auth files, kubeconfig-based ``oc registry login``, and/or explicit
+username:password pairs.
 
 Usage examples:
 
@@ -11,6 +12,17 @@ Usage examples:
     with RegistryConfig(
         source_files=['/path/to/auth.json'],
         registries=['registry.redhat.io', 'quay.io/openshift-release-dev'],
+    ) as auth_file:
+        cmd.extend(['--registry-config', auth_file])
+
+    # Use kubeconfig to obtain CI registry credentials
+    with RegistryConfig(
+        kubeconfig=os.environ.get('KUBECONFIG'),
+        source_files=['/path/to/quay_auth.json'],
+        registries=[
+            'quay.io/openshift-release-dev',
+            'registry.ci.openshift.org',
+        ],
     ) as auth_file:
         cmd.extend(['--registry-config', auth_file])
 
@@ -48,6 +60,8 @@ import tempfile
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Optional
+
+from artcommonlib import exectools
 
 logger = logging.getLogger(__name__)
 
@@ -119,15 +133,20 @@ class RegistryConfig:
     Context manager that builds a temporary Docker auth config.json.
 
     Constructs a minimal auth file containing only the requested registries,
-    sourced from existing auth files and/or explicit credentials.
+    sourced from existing auth files, kubeconfig-based ``oc registry login``,
+    and/or explicit credentials.
 
     When a registry appears in both source files and credentials, the
     explicit credential takes precedence.  When a registry appears in
     multiple source files, the first file wins.
 
     Arg(s):
+        kubeconfig (str | None): Path to a kubeconfig file. When provided,
+            ``oc --kubeconfig=<path> registry login`` is run during ``__enter__``
+            to obtain CI registry credentials, which are added as an additional
+            source.
         source_files (list[str] | None): Paths to existing auth.json files to extract from.
-        registries (list[str] | None): Registry paths to cherry-pick from source_files.
+        registries (list[str] | None): Registry paths to cherry-pick from source_files / kubeconfig.
         credentials (list[RegistryCredential] | None): Explicit username/password credentials.
     """
 
@@ -136,22 +155,25 @@ class RegistryConfig:
         source_files: Optional[list[str]] = None,
         registries: Optional[list[str]] = None,
         credentials: Optional[list[RegistryCredential]] = None,
+        kubeconfig: Optional[str] = None,
     ) -> None:
         """
         Initialize the RegistryConfig context manager.
 
         Arg(s):
             source_files (list[str] | None): Paths to existing auth.json files.
-            registries (list[str] | None): Registry paths to extract from source_files.
+            registries (list[str] | None): Registry paths to extract from source_files / kubeconfig.
             credentials (list[RegistryCredential] | None): Explicit credentials.
+            kubeconfig (str | None): Path to a kubeconfig file for ``oc registry login``.
 
         Raises:
             ValueError: If no registries or credentials are provided, or if
-                        registries are specified without source_files.
+                        registries are specified without source_files or kubeconfig.
         """
         self._source_files = list(source_files) if source_files else []
         self._registries = [_normalize_registry_path(r) for r in registries] if registries else []
         self._credentials = list(credentials) if credentials else []
+        self._kubeconfig = kubeconfig
         self._temp_auth_file: Optional[Path] = None
         self._merged_auths: dict[str, dict[str, str]] = {}
 
@@ -159,9 +181,9 @@ class RegistryConfig:
         if not self._registries and not self._credentials:
             raise ValueError("RegistryConfig requires at least one registry or credential")
 
-        # Can't extract registries without source files
-        if self._registries and not self._source_files:
-            raise ValueError("source_files must be provided when registries are specified")
+        # Can't extract registries without a credential source
+        if self._registries and not self._source_files and not self._kubeconfig:
+            raise ValueError("source_files or kubeconfig must be provided when registries are specified")
 
     def __enter__(self) -> str:
         """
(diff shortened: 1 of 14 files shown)


locriandev:
> In `@pyartcd/pyartcd/oc.py`:
> - Around line 173-178: The helpers are appending "--registry-config=..." into
> args even though common_oc_wrapper already injects that flag when
> registry_config is passed; remove the manual appending of
> f"--registry-config={registry_config}" from the args construction in the
> functions that call common_oc_wrapper (e.g., where
> common_oc_wrapper("extract_tools", "adm", args, ...) is invoked) and apply the
> same removal to the other helper that builds similar args (the block around the
> call to extract_baremetal_installer/common_oc_wrapper later), leaving
> registry_config only in the common_oc_wrapper call so the flag is not
> duplicated.

Done with adc72e3399cd03a4442a15b79a1af4ba58eaf6b6

locriandev:
> > doozer/doozerlib/backend/konflux_olm_bundler.py (1)> `763-776`: _⚠️ Potential issue_ | _🔴 Critical_
> > > **Only mirror the bundle after the final outcome is `SUCCESS`.**
> > > Lines 774-776 push to `art-images-share` before the code checks `outcome`, and before EC verification has a chance to flip the build to failure. That means a failed bundle can be published to the shared repo if Tekton still produced `IMAGE_URL`/`IMAGE_DIGEST`.

art-images-share is fine - we will be moving to stage releases soon as part of https://redhat.atlassian.net/browse/ART-17450

locriandev:
> Missing validation for QCI_USER and QCI_PASSWORD environment variables.
>
>Direct access to os.environ['QCI_USER'] and os.environ['QCI_PASSWORD'] will raise a KeyError if these variables are not set. This is inconsistent with the explicit validation done for QUAY_AUTH_FILE (lines 799-803).

Done with ad1d20241d933fe9c27012210601a8cafe84beee

artcommon/artcommonlib/registry_config.py
```diff
@@ -0,0 +1,452 @@
...
+        to obtain CI credentials, load source files, match registries, apply
+        explicit credentials, and write the merged auth file.
+        """
+        merged: dict[str, dict[str, str]] = {}
+
+        # Registry aliases: registries that share authentication realms
+        # When the requested registry is not found, try these fallbacks
+        registry_aliases = {
+            'brew.registry.redhat.io': 'registry.redhat.io',
+        }
```
thegreyd, line 254: This looks like something that would be a constant / class var?

artcommon/artcommonlib/registry_config.py
```diff
@@ -0,0 +1,452 @@
...
+            source_auths = self._load_source_files()
+        finally:
+            if oc_login_file:
+                self._source_files.remove(oc_login_file)
+                try:
+                    os.unlink(oc_login_file)
+                    logger.debug("Removed oc login temp file %s", oc_login_file)
+                except OSError as err:
+                    logger.warning("Could not remove %s: %s", oc_login_file, err)
+                    raise
```
thegreyd: If `_load_source_files()` raises and `os.unlink()` also fails here, this `raise` replaces the original exception with the cleanup `OSError`. The caller loses the real diagnostic (e.g. "invalid JSON in source file"). Remove the `raise` — the `finally` block will automatically re-propagate the original exception when the cleanup completes without raising. The warning log is sufficient for a temp file cleanup failure.

(written by claude ^)

artcommon/artcommonlib/registry_config.py
```diff
@@ -0,0 +1,452 @@
...
+        with os.fdopen(fd, "w", encoding="utf-8") as f:
+            json.dump({"auths": {}}, f)
+
+        cmd = f"oc --kubeconfig {self._kubeconfig} registry login --to={temp_path}"
+        rc, _, stderr = exectools.cmd_gather(cmd)
+        if rc != 0:
+            os.unlink(temp_path)
+            raise RuntimeError(f"Failed to run 'oc registry login' with kubeconfig '{self._kubeconfig}': {stderr}")
+
+        self._source_files.append(temp_path)
```
thegreyd: Nit: `_run_oc_registry_login` appends to `self._source_files` as a side effect, and then `_resolve_and_write` is responsible for undoing that mutation and cleaning up the temp file in a `finally` block. It's a somewhat fragile caller-callee contract — the docstring even notes "the caller is responsible for removing the temp file."

It might be worth considering whether this method could return the credential data directly (e.g. the parsed auths dict) rather than modifying instance state. That way the temp file lifecycle could be fully contained within the method that creates it, and `_resolve_and_write` wouldn't need the cleanup block at all.

pyartcd/pyartcd/pipelines/ocp4_konflux.py
```diff
@@ -778,11 +781,90 @@ async def sync_build(build):
...
+                REGISTRY_REDHAT_IO,  # For: RHEL base images
+                REGISTRY_BREW,  # For: parent images (uses registry.redhat.io creds via alias)
+                REGISTRY_CI_OPENSHIFT,  # For: CI operations
+            ],
+            credentials=credentials,
+        ) as global_auth_file:
+            self._registry_auth_file = global_auth_file
+
+            # cosign requires DOCKER_CONFIG pointing to a dir containing config.json
+            docker_config_dir = tempfile.mkdtemp(prefix='docker_config_')
```
thegreyd, line 841: The `run()` method has about 40 lines of boilerplate for setting up `DOCKER_CONFIG` — creating a temp directory, copying the auth file into it as `config.json`, saving/restoring the env var, and cleaning up with nested `try/finally`. Any other pipeline that needs cosign compatibility would likely end up duplicating this same pattern.

It could be worth considering whether `RegistryConfig` itself could handle this — e.g. an option to also produce a `DOCKER_CONFIG`-compatible directory alongside the auth file, so the caller doesn't have to manage that lifecycle separately. Not necessarily in this PR, but might be a nice simplification down the line.

artcommon/artcommonlib/registry_config.py
locriandev: Done with 63fe13617d7b94a4158f341b3b8069b9120eb810

artcommon/artcommonlib/registry_config.py
locriandev: done with ec8d85eee4b98f85a71f3a1f144a518de5e6586a

thegreyd:
Worth noting: `release_gen_payload.py` still reads `QUAY_AUTH_FILE` directly for `manifest_tool` calls (see #2776), so in an `ocp4-konflux` run there would be two parallel auth paths. That's fine given the scoping here, but might be worth unifying down the line when other pipelines adopt `RegistryConfig`.

artcommon/artcommonlib/registry_config.py
locriandev: Done with 324350d2c55ae901b19aa338fe46035669e4e10c

pyartcd/pyartcd/pipelines/ocp4_konflux.py
locriandev: I think this can be handled in a separate PR

thegreyd:
awesome change! Thanks for quickly addressing feedback
tests are failing but otherwise lgtm :+1:

PR #2808, opened 2026-04-22
ART-18057: make LP shipment MRs ready and trigger stage pipeline

fgallott:
## Summary                                         
  - Layered Product shipment MRs (created by `release-from-fbc`) now automatically transition from draft to ready and trigger the stage pipeline, matching the existing OCP (`prepare-release-konflux`) and microshift-bootc behavior.                                                                                                                                                                                        
  - Previously, LP shipment MRs stayed in draft state, requiring manual intervention to mark ready and trigger the pipeline.
                                                                                                                                                                                                                    
  ## Changes                                         
  - Added `set_shipment_mr_ready()` method to `ReleaseFromFbcPipeline` that removes the "Draft:" prefix, waits 30s for GitLab to update, then triggers the CI stage pipeline                                        
  - Called from `run()` after `create_shipment_mr()` succeeds, inside the existing try/except block so failures don't crash the pipeline                                                                            
  - Uses `self.logger` for output (no Slack client in this pipeline)                                                                                                                                                
                                                                                                                                                                                                                    
  ## Test plan                                                                                                                                                                                                      
  - [x] 4 new unit tests: happy path, dry run, pipeline trigger failure, set_mr_ready returns None                                                                                                                  
  - [x] All 48 tests pass                                                                                                                                                                                           
  - [x] Linting clean                                
                                                                                                                                                                                                                    
  Resolves: [ART-18057](https://redhat.atlassian.net/browse/ART-18057)

### DIFF
--- a/pyartcd/pyartcd/pipelines/release_from_fbc.py
+++ b/pyartcd/pyartcd/pipelines/release_from_fbc.py
@@ -639,6 +639,26 @@ class ReleaseFromFbcPipeline:
         self.shipment_mr_url = mr_url
         return mr_url
 
+    async def set_shipment_mr_ready(self):
+        """
+        Mark the shipment MR as ready by removing the Draft prefix from the title.
+        This should be called at the end of the pipeline when all work is complete.
+        """
+        mr = await self._gitlab.set_mr_ready(self.shipment_mr_url)
+
+        if mr and not self.dry_run:
+            self.logger.info("Waiting for 30 seconds to ensure MR is updated...")
+            await asyncio.sleep(30)
+
+            try:
+                pipeline_url = await self._gitlab.trigger_ci_pipeline(mr)
+                if pipeline_url:
+                    self.logger.info(f"CI pipeline triggered: {pipeline_url}")
+                else:
+                    self.logger.warning(f"Failed to trigger CI pipeline for MR branch {mr.source_branch}")
+            except Exception as e:
+                self.logger.warning(f"Failed to trigger CI MR pipeline for branch {mr.source_branch}: {e}")
+
     async def update_shipment_data(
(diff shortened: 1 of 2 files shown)


pyartcd/pyartcd/pipelines/release_from_fbc.py
```diff
@@ -639,6 +639,26 @@ async def create_shipment_mr(self, shipments_by_kind: Dict[str, ShipmentConfig],
...
+    async def set_shipment_mr_ready(self):
+        """
+        Mark the shipment MR as ready by removing the Draft prefix from the title.
+        This should be called at the end of the pipeline when all work is complete.
+        """
+        mr = await self._gitlab.set_mr_ready(self.shipment_mr_url)
+
+        if mr and not self.dry_run:
+            self.logger.info("Waiting for 30 seconds to ensure MR is updated...")
+            await asyncio.sleep(30)
```
locriandev, line 651: Is it worth polling on the MR title to check for the presence of `Draft`, instead of using an arbitrary delay? 30 seconds seem more than enough, but you never know...

fgallott: I made it consistent with what we do for ocp shipment MRs (https://github.com/openshift-eng/art-tools/blob/main/pyartcd/pyartcd/pipelines/prepare_release_konflux.py#L1198-L1201)

fgallott: I've in mind a refactor to centralize this functionality for all pipelines, and possibly improve it, but for now I'd rather keep it consistent with the rest

locriandev (commented):
/lgtm

PR #2810, opened 2026-04-22
Add --registry-config option to elliott for explicit registry auth

locriandev:
The ocp4-konflux pipeline's RegistryConfig sets DOCKER_CONFIG but not REGISTRY_AUTH_FILE, causing oc to fall back to an inaccessible system path. Instead of relying on environment variables, elliott now accepts an explicit --registry-config CLI option that is threaded through Runtime into get_nvrs_from_release() and all its callers. The ocp4_konflux pipeline passes its merged auth file via this option when invoking elliott's find-bugs:golang command.


joepvd (approved):
/lgtm

PR #2811, opened 2026-04-22
Add --registry-config option to doozer for explicit registry auth

locriandev:
Doozer's cosign and oc commands were reading QUAY_AUTH_FILE directly, which points to the original Jenkins file rather than the merged auth file crafted by RegistryConfig. Add --registry-config as a global CLI option that defaults to QUAY_AUTH_FILE when not specified. The ocp4_konflux pipeline now passes the merged auth file to doozer via this option, ensuring cosign download attestation has the right credentials.


joepvd (approved):
/lgtm

PR #2815, opened 2026-04-23
Handle undetermined upstream RHEL version gracefully instead of crashing

locriandev:
Images like openshift-enterprise-deployer use image stream tags (e.g. :cli, :base) that don't encode RHEL version info. When no parent image yields a RHEL version, log a warning and assume it matches ART's intended version instead of raising IOError that blocks the entire build.


joepvd (approved):
/lgtm

PR #2831, opened 2026-04-24
Skip package name comparison for test assembly builds

adobes1:
Add an optional `ignore_name` parameter to `compare_nvr()` to skip package name comparison during NVR checks.

This change is needed to avoid issues with test assembly containing NVRs without `-container` suffix (remnant of Konflux test builds which did not add `-container` to the NVR).

[Test build](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-konflux/32507/)


joepvd (approved):
/lgtm

PR #2845, opened 2026-04-28
ART-14348: Support including 4.22 update sources in 5.0 release metadata

thegreyd:
Equivalent of #2601 

Move Cincinnati version ancestry logic from doozerlib to artcommonlib to support the OCP 4.22 -> 5.0 major version transition. The previous code used `minor - 1` to determine the previous version channel, which would query `candidate-5.-1` instead of `candidate-4.22` for OCP 5.0 releases.

The new calc_upgrade_sources_async fetches build-suggestions FIRST to extract the previous minor version from minor_min (e.g., 4.22.0-rc.0 -> 4.22), then queries the correct Cincinnati channel.

New artcommonlib.ocp_version_ancestry module:
- SuggestionsSpec/BuildSuggestions: Pydantic models with semver validation
- get_build_suggestions_async: fetch/validate build-suggestions YAML
- get_previous_minor_version_from_suggestions: extract previous version
- get_cincinnati_channels: moved from doozerlib with proper int typing
- get_channel_versions_async: httpx rewrite of get_channel_versions
- calc_upgrade_sources_async: replaces get_release_calc_previous
- extract_version_fields: moved to artcommonlib.util (without golang hack)

Removed from doozerlib:
- get_channel_versions, get_build_suggestions, get_release_calc_previous
- get_cincinnati_channels (moved to artcommonlib)
- graph_content_stable/candidate CLI options and params: these were already broken (the channel name check `channel == 'stable'` never matched actual channel names like `candidate-4.22`) and have no valid use cases. To manually specify upgrade sources, use the existing `--previous` option on `release:gen-assembly`.
- Unused release:calc-upgrade-tests CLI command

Test

```
$ uv run doozer release:calc-previous --version 5.0.0-rc.0 --arch x86_64
5.0.0-ec.0,4.22.0-rc.1,4.22.0-rc.0

$ uv run doozer release:calc-previous --version 4.22.0-rc.2 --arch x86_64
4.22.0-rc.1,4.22.0-rc.0,4.22.0-ec.5,4.22.0-ec.4,4.22.0-ec.3,4.22.0-ec.2,4.22.0-ec.1,4.22.0-ec.0,4.21.12,4.21.11,4.21.10,4.21.9,4.21.8,4.21.7,4.21.6


```


artcommon/artcommonlib/ocp_version_ancestry.py
```diff
@@ -0,0 +1,418 @@
...
+        ...,
+        description="Minimum version from previous minor to include (e.g., '4.22.0-rc.0')",
+    )
+    minor_max: Optional[str] = Field(
+        None,
+        description="Maximum version from previous minor, typically 'X.Y.9999' as placeholder. "
+        "If omitted, all versions >= minor_min with the same major.minor are included.",
+    )
+    minor_block_list: list[str] = Field(
+        ...,
```
fgallott: this can be 

```
minor_block_list: list[str] = Field(                                                                                                                           
          default_factory=list,
```

artcommon/artcommonlib/ocp_version_ancestry.py
```diff
@@ -0,0 +1,418 @@
...
+        description="Minimum version from current minor to include (e.g., '5.0.0-ec.0')",
+    )
+    z_max: Optional[str] = Field(
+        None,
+        description="Maximum version from current minor, typically 'X.Y.9999' as placeholder. "
+        "If omitted, all versions >= z_min with the same major.minor are included.",
+    )
+    z_block_list: list[str] = Field(
+        ...,
```
fgallott: same

artcommon/artcommonlib/ocp_version_ancestry.py
```diff
@@ -0,0 +1,418 @@
...
+        description="Explicitly blocked versions from current minor (empty list if none)",
+    )
+
+    @field_validator('minor_min', 'minor_max', 'z_min', 'z_max')
```
fgallott, line 75: what about something like:

```
      @staticmethod                                                                                                                                                  
      def _parse_semver(v: str) -> None:                          
          try:                                                                                                                                                       
              semver.VersionInfo.parse(v)                         
          except ValueError as e:                                                                                                                                    
              raise ValueError(f"Invalid semver format '{v}': {e}") from e

      @field_validator("minor_min", "minor_max", "z_min", "z_max")                                                                                                   
      @classmethod
      def validate_semver(cls, v: str | None) -> str | None:                                                                                                         
          """Validate that version strings are valid semver."""   
          if v is not None:                                                                                                                                          
              cls._parse_semver(v)
          return v                                                                                                                                                   
                                                                  
      @field_validator("minor_block_list", "z_block_list")
      @classmethod
      def validate_semver_list(cls, versions: list[str]) -> list[str]:
          """Validate that all versions in block lists are valid semver."""                                                                                          
          for v in versions:
              cls._parse_semver(v)                                                                                                                                   
          return versions     
```

artcommon/artcommonlib/util.py
```diff
@@ -39,6 +39,22 @@
 KONFLUX_LOGGER = logutil.get_logger(__name__)
 
 
+def extract_version_fields(version, at_least=0):
```
fgallott, line 42: I see that you are not removing the corresponding one in doozer, is that intended?

thegreyd: first a disclaimer that this PR is basically all yuxiang's work :) so I can only guess on his motivations. But I think it's intentional. doozerlib.util.extract_version_fields() still has existing doozer callers. Consolidating all of that would be better as a separate cleanup PR

ashwindasr (approved):

PR #2847, opened 2026-04-28
NO-ISSUE: doozer/doozerlib/cli/images_streams: Explain how sync-ci-images OCPBUGS are closed

wking:
ART wants to create these OCPBUGS to get engineer attention about preserving CI fidelity (we want CI builds to be similar to the shipped ART builds).  These bugs are not linked from errata since 01692b6598 ( #508).  And there's no expectation for QE load either, because the running CI signal that the image-sync improves is already covering the CI images being updated (or we wouldn't care about aligning the CI and ART images).  This commit updates the message to explain that no QE time is needed, and that ART will run a bulk close, and that that bulk close is expected during the RC phase.  Hopefully that keeps component teams from wasting post-merge time on these tickets.

### DIFF
--- a/doozer/doozerlib/cli/images_streams.py
+++ b/doozer/doozerlib/cli/images_streams.py
@@ -973,6 +973,12 @@ in #forum-ocp-art to discuss the discrepancy.
 
 Closing this issue without addressing the difference will cause the issue to
 be reopened automatically.
+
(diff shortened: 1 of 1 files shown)


joepvd (approved):
/lgtm

PR #2849, opened 2026-04-29
doozer: refresh reposync metadata after syncing

thegreyd:
## Summary
- use an isolated dnf cache for `beta:reposync` so stale cached repodata does not leak across runs
- force metadata refresh during sync and rebuild local repodata with `createrepo_c --update --keep-all-metadata`
- add focused coverage for the reposync CLI flow and repodata loader refactor

## Test plan
- [x] `uv run pytest --verbose --color=yes doozer/tests/test_repodata.py doozer/tests/cli/test_beta_reposync.py`
- [x] `uv run ruff check doozer/doozerlib/cli/__main__.py doozer/doozerlib/repodata.py doozer/tests/test_repodata.py doozer/tests/cli/test_beta_reposync.py`
- [x] `uv run ruff format --check doozer/doozerlib/cli/__main__.py doozer/doozerlib/repodata.py doozer/tests/test_repodata.py doozer/tests/cli/test_beta_reposync.py`

Test dry-run: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fsync-for-ci/298470/console

### DIFF
--- a/doozer/doozerlib/cli/__main__.py
+++ b/doozer/doozerlib/cli/__main__.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import os
+import shlex
 import shutil
 import sys
 import tempfile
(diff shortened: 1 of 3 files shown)


joepvd:
/hold
Unsure if this is needed. The case that triggered this investigation is explained through "config drift in repos". 

`createrepo_c` is rather time consuming and cpu intensive, so I would like to avoid if we can. If we introduce it, we need the `--no-database` option, as generating the unused sqlite db is even more expensive than "just generate the xmls".

I also do not completely understand why regenerating repo data locally would be needed, as it is also downloaded through `--download-metadata` of `dnf reposync`.

PR #2860, opened 2026-04-30
ART-18307: build-microshift-bootc: two-phase build (open then hermetic)

adobes1:
## Summary
- Hermetic Konflux builds require RPM lockfiles, which are populated in part from a prior successful build's `installed_rpms` in BigQuery. Since microshift-bootc is typically built once per assembly, there is no prior build to draw from.
- The pipeline now automatically runs an open build first (Phase 1) to produce a successful build record, then a hermetic build (Phase 2) whose lockfile generator uses that record's `installed_rpms`.
- Phase 1 is auto-skipped on retries if an open build already exists for the assembly. A `--force-open-build` flag allows forcing a redo when the open build is stale.

## Test plan
- [x] Unit tests pass (`pytest pyartcd/tests/pipelines/test_build_microshift_bootc.py`)
- [ ] Test with a dry-run build on a test assembly
- [ ] Verify Phase 1 skip works on retry (open build already exists)
- [ ] Verify `--force-open-build` forces Phase 1 redo

Companion aos-cd-jobs PR to follow (adds `FORCE_OPEN_BUILD` Jenkins parameter). The art-tools change works standalone — the new flag just defaults to False.

### DIFF
--- a/pyartcd/pyartcd/pipelines/build_microshift_bootc.py
+++ b/pyartcd/pyartcd/pipelines/build_microshift_bootc.py
@@ -66,6 +66,8 @@ yaml = new_roundtrip_yaml_handler()
 class BuildMicroShiftBootcPipeline:
     """Rebase and build MicroShift for an assembly"""
 
+    BOOTC_IMAGE_NAME = "microshift-bootc"
+
     def __init__(
         self,
         runtime: Runtime,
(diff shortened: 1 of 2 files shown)


adobes1:
Used for [4.20.20](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fbuild-microshift/3798/) microshift-bootc release

PR #2864, opened 2026-05-04
fix(lockfile): warn instead of failing on cross-arch RPM version mismatch

lgarciaaco:
## Summary

Lockfile generation no longer fails when the latest RPM EVR sets differ between architectures; the condition is reported as a warning so rebase/lockfile workflows can finish.

## Problem
**Before:** A cross-arch RPM "latest EVR set" mismatch raised `ValueError`, so lockfile generation (and dependent rebases) failed.

**After:** The same condition emits a warning with package/arch details; lockfile generation continues.

## Implementation Details
- `_validate_cross_arch_version_sets` in `doozer/doozerlib/lockfile.py` logs `RPM version set mismatches: …` at warning level instead of raising.
- `doozer/tests/test_lockfile_validation.py` asserts the warning via `caplog` where failures were previously expected.


locriandev (approved):
/lgtm

PR #2875, opened 2026-05-07
feat(konflux): include patch in golang-builder Konflux component names

lgarciaaco:
## Summary

Konflux base-image release snapshots now name golang-builder components with the full Go patch level (for example `golang-builder-v1.22.12-rhel8` instead of `golang-builder-v1.22-rhel8`).

## Problem

**Before:** `get_golang_builder_component_name` only used major.minor from the NVR version field.

**After:** Component names always use `vMAJOR.MINOR.PATCH`, using a single normalization helper shared with `go:report` version extraction.

## Implementation

- Added `rpm_version_to_golang_v_semver` in `artcommonlib.rpm_utils` (first three dot-separated segments, zero-pad to three, optional leading `v` stripped for parsing).
- `KonfluxImageBuilder.get_golang_builder_component_name` uses that helper after `parse_nvr`.
- `elliott go:report` (`go_version_from_nvr_string`) uses the same helper for the semantic triple (still appends `.el*` when `--ignore-rhel` is not set).
- Tests in `artcommon/tests/test_rpm_utils.py` and updated expectations in `TestKonfluxImageBuilderGolangComponent`.

### DIFF
--- a/artcommon/artcommonlib/rpm_utils.py
+++ b/artcommon/artcommonlib/rpm_utils.py
@@ -68,6 +68,34 @@ def parse_nvr(nvre: str):
     return result
 
 
+def rpm_version_to_golang_v_semver(version: str) -> str:
+    """
+    Normalize an RPM NVR *version* field to ``v`` + exactly three dot-separated segments (Go-style semver core).
+
+    Used for golang-builder labels (Konflux component names, etc.). First three segments of the version are kept;
+    shorter versions are zero-padded (``1.24`` -> ``v1.24.0``); four or more segments truncate (``1.24.0.1`` ->
+    ``v1.24.0``). Optional leading ``v`` / ``V`` is stripped before parsing.
+
+    :param version: The ``version`` value from :func:`parse_nvr`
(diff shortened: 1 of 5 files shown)


joepvd:
/hold
I think it is wrong to introduce patch versions in component names. What is the reason to want this?

artcommon/artcommonlib/rpm_utils.py
```diff
@@ -68,6 +68,34 @@ def parse_nvr(nvre: str):
...
+        v = v[1:]
+    parts = v.split(".")
+    if not parts:
+        raise ValueError(f"Invalid RPM version: {version!r}")
+    head = parts[:3]
+    if any(not seg for seg in head):
+        raise ValueError(f"Invalid RPM version (empty segment): {version!r}")
+    while len(head) < 3:
+        head.append("0")
+    return f"v{head[0]}.{head[1]}.{head[2]}"
```
thegreyd, line 96: Can we explore using the semver library here to parse - there's usage in our repo

PR #2851, opened 2026-04-29
ART-18048: feat(images): implement synchronous base image registry coordination

lgarciaaco:
## Summary
Fixes coordination gap where 235+ dependent images continued using stale quay.io references after base images were released to registry.redhat.io. Implements synchronous base image release handling with proper failure management and database updates using correct registry.redhat.io pullspecs.

## Problem
**Before:** Base images release to registry.redhat.io but 235+ dependent images continue using stale quay.io references in their FROM statements because Konflux database pullspecs are not updated after registry releases.

**After:** Dependent images automatically resolve to use the released registry.redhat.io pullspec in their FROM statements after base image releases complete successfully.

## Implementation Details
Adds synchronous base image release coordination through:

### Key Changes
- Add `released` parameter to `update_konflux_db` for registry pullspec coordination
- Implement synchronous base image release with `block_until_complete=True`  
- Add comprehensive test coverage for success/failure scenarios (7 new test cases)
- Ensure proper failure handling with FAILURE outcome updates when registry release fails

The solution eliminates the database coordination gap by making base image release synchronous and updating database records with correct registry.redhat.io pullspecs upon successful release, while marking builds as FAILURE when release fails.

PR #2876, opened 2026-05-07
fix(doozer): seed lockfile RPMs from latest success or unreleased build

lgarciaaco:
## Summary

Regression follow-up to https://github.com/openshift-eng/art-tools/pull/2851. That change introduced `KonfluxBuildOutcome.UNRELEASED` and persists many successful Konflux builds as `unreleased` (with SBOM-backed `installed_rpms`) before or instead of a `success` row. Lockfile generation still called `get_latest_build(outcome=success)` only, so the **newest** install set could live on `unreleased` while we seeded from an **older** `success` row—stale NVR pins and failing hermetic `dnf` (notably golang-builder).

## Problem

**Before:** `fetch_rpms_from_build` ignored newer `unreleased` DB rows when a `success` row existed with an earlier completion time.

**After:** We load the latest `success` and latest `unreleased`, take the newer by `end_time` (else `start_time`), dedupe identical objects from the two lookups, and if the time-winner has no `installed_rpms`, prefer the other candidate when it does. Unit tests cover ordering and fallback.

## Implementation

- `doozer/doozerlib/image.py`: `_lockfile_seed_build_completion_ts`, `_pick_lockfile_seed_build`, dual `get_latest_build` + info logging in `fetch_rpms_from_build`.
- `doozer/tests/test_image.py`: expect two DB calls where applicable; add tests for newer unreleased vs success and empty-rpms fallback.

No change to `pull_url()` / base-image coordination—this is scoped to lockfile RPM seeding only.

## Validation / test runs

Test Jenkins build (base-image-release): https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fbase-image-release/5140/

Test Konflux PipelineRun (rhel-9-golang-1-22 / openshift-golang-builder): https://konflux-ui.apps.kflux-ocp-p01.7ayg.p1.openshiftapps.com/ns/ocp-art-tenant/applications/rhel-9-golang-1-22/pipelineruns/rhel-9-golang-1-22-openshift-golang-builder-hprk5

The PLR **builds with hermetic** prefetch as expected; the overall run still **fails at release policy (Conforma / EC)**—that failure is separate from the image build+hermetic path exercised here.

### DIFF
--- a/doozer/doozerlib/image.py
+++ b/doozer/doozerlib/image.py
@@ -47,27 +47,17 @@ def _pick_lockfile_seed_build(
     Choose which Konflux DB row should seed lockfile installed_rpms: the latest of SUCCESS and UNRELEASED
     by completion time. If the time-newest row has no installed_rpms, prefer the other when it does.
     """
-    raw = [b for b in (success_build, unreleased_build) if b is not None]
-    candidates = []
-    seen_ids: set[int] = set()
-    for b in raw:
-        bid = id(b)
-        if bid not in seen_ids:
-            seen_ids.add(bid)
-            candidates.append(b)
+    candidates = [b for b in (success_build, unreleased_build) if b is not None]
     if not candidates:
         return None
-    if len(candidates) == 1:
-        chosen = candidates[0]
-    else:
-        a, b = candidates[0], candidates[1]
-        chosen = a if _lockfile_seed_build_completion_ts(a) > _lockfile_seed_build_completion_ts(b) else b
-
-    if not (chosen.installed_rpms or []):
-        for other in candidates:
-            if other is not chosen and (other.installed_rpms or []):
-                return other
-    return chosen
+    # Newer completion time wins; on equal timestamps prefer the latter (UNRELEASED) like strict `>` ordering.
+    chosen = max(
+        enumerate(candidates),
+        key=lambda ib: (_lockfile_seed_build_completion_ts(ib[1]), ib[0]),
+    )[1]
+    if chosen.installed_rpms:
+        return chosen
+    return next((c for c in candidates if c is not chosen and c.installed_rpms), chosen)
 
 
 @lru_cache(maxsize=256)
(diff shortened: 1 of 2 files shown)


doozer/doozerlib/image.py
```diff
@@ -28,6 +29,37 @@
...
+    if not candidates:
+        return None
+    # Newer completion time wins; on equal timestamps prefer the latter (UNRELEASED) like strict `>` ordering.
+    chosen = max(
+        enumerate(candidates),
+        key=lambda ib: (_lockfile_seed_build_completion_ts(ib[1]), ib[0]),
+    )[1]
+    if chosen.installed_rpms:
+        return chosen
+    return next((c for c in candidates if c is not chosen and c.installed_rpms), chosen)
```
locriandev:
```suggestion
  def _pick_lockfile_seed_build(
      success_build: Optional[KonfluxBuildRecord],
      unreleased_build: Optional[KonfluxBuildRecord],
  ) -> Optional[KonfluxBuildRecord]:
      """
      Choose which Konflux DB row should seed lockfile installed_rpms: the latest of SUCCESS and UNRELEASED
      by completion time. If the time-newest row has no installed_rpms, prefer the other when it does.
      """
      candidates = [b for b in (success_build, unreleased_build) if b is not None]
      if not candidates:
          return None

      # Find the build with the latest completion time
      if len(candidates) == 1:
          chosen = candidates[0]
      else:
          # Both success and unreleased exist - compare timestamps
          success_ts = _lockfile_seed_build_completion_ts(candidates[0])
          unreleased_ts = _lockfile_seed_build_completion_ts(candidates[1])

          # Prefer unreleased on tie (>=) to favor newer outcome status
          if unreleased_ts >= success_ts:
              chosen = candidates[1]  # unreleased
          else:
              chosen = candidates[0]  # success

      # If chosen has RPMs, use it; otherwise try the other candidate
      if chosen.installed_rpms:
          return chosen
      return next((c for c in candidates if c is not chosen and c.installed_rpms), chosen)

```

locriandev: Doesn't change the outcome, just trying to make the function more readable. Having that `max()` comparison with a lambda key operating on tuples is not... ideal for understanding.

lgarciaaco: refactored the function for better readability [b3c71a7](https://github.com/openshift-eng/art-tools/pull/2876/commits/b3c71a746ea241127c96880060353acdf8ed9a24)

joepvd (approved):
/lgtm

PR #2880, opened 2026-05-08
Revert "ART-18048: feat(images): implement synchronous base image registry coordination"

locriandev:
Reverts openshift-eng/art-tools#2851


locriandev:
/hold

as we're trying to better assess the impact of this change

PR #2882, opened 2026-05-08
Add --skip-rpms flag to scan-sources-konflux for local testing

locriandev:
## Summary
- Adds `--skip-rpms` flag to `scan-sources-konflux` command
- When set, initializes runtime with `mode='images'` only, skipping all RPM metadata loading
- Prevents RPM distgit cloning, which requires SSH access to internal Red Hat systems

## Use Case
This flag is useful for local development and testing when developers:
- Don't have SSH access to pkgs.devel.redhat.com (internal RPM distgit)
- Only need to test image-related scan functionality
- Want faster scan runs without RPM checks

## Implementation
Follows the same pattern already used for OKD variant, which also skips RPMs by using `mode='images'`.

## Test Plan
- [x] Tested locally with `--skip-rpms` flag to verify no RPM distgit cloning occurs
- [x] Pre-commit hooks pass

### DIFF
--- a/doozer/doozerlib/cli/scan_sources_konflux.py
+++ b/doozer/doozerlib/cli/scan_sources_konflux.py
@@ -59,6 +59,7 @@ class ConfigScanSources:
         rebase_priv: bool = False,
         dry_run: bool = False,
         variant: str = 'ocp',
+        skip_rpms: bool = False,
     ):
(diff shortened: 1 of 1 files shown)


joepvd (approved):
/lgtm

PR #2883, opened 2026-05-08
fix: Record FAILURE build records when base-image release workflow fails

locriandev:
Previously, when the base-image release workflow failed (e.g., during conformance checks, snapshot creation, or release creation), the build records remained in PENDING status instead of being updated to FAILURE. This caused child images to not be attempted (which is correct), but left no failure record in the database for the parent build.

Changes:
- Added `_record_build_failures()` helper method to create FAILURE build records in the database
- Call `_record_build_failures()` when:
  1. Snapshot creation fails
  2. Release creation fails
  3. Release completion fails (e.g., conformance check failure)
- Each failure record includes the reason for better diagnostics

This ensures that failed base-image builds are properly tracked in the database with FAILURE status instead of remaining in PENDING.

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>

### DIFF
--- a/doozer/doozerlib/backend/base_image_handler.py
+++ b/doozer/doozerlib/backend/base_image_handler.py
@@ -7,6 +7,7 @@ with URL extraction and streams.yml updates to be implemented later.
 """
 
 import asyncio
+from datetime import datetime, timezone
 from typing import Dict, List, Optional, Tuple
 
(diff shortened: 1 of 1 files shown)


locriandev:
Superseded by https://github.com/openshift-eng/art-tools/pull/2886

PR #2885, opened 2026-05-11
fix(doozer): Konflux digest in DB; art-images-base FROM handling in rebaser

lgarciaaco:
## Summary

Base and golang builder images get a Konflux build on quay with SBOMs attached there. The earlier base-image release work started storing **registry.redhat.io** on the Konflux DB row after release so dependents could point their Dockerfiles at Red Hat. That helped **FROM** lines, but anything that still uses **image_pullspec** to pull metadata or SBOMs was suddenly asking **cosign** (and the registry) for the wrong place—the SBOM never moved to that RH tag the same way. So we were trading one fix for a quiet break.

This change keeps the **digest-pinned Konflux pullspec** on the build record (what the pipeline actually produced), and moves the “use Red Hat for **FROM**” decision into the **rebaser** where it belongs.

## Problem

**Before:** After a successful base-image release, `update_konflux_db` could replace `image_pullspec` with `registry.redhat.io/.../art-images-base:<nvr>`. Downstream tooling that still reads `image_pullspec` for SBOM or registry access was no longer hitting the image the pipeline published.

**After:** Success records keep the Konflux digest pullspec. Coordinated rebases (parent in the same run) still set **FROM** to the **prospective** art-images-base tag from `_rebased_member_image_nvr`—we assume the base will build and publish; no registry probe there. If the parent is only **late-resolved from the DB**, we try the RH tag for that build’s NVR and **only if** `oc image info` says it exists; otherwise we fall back to the stored Konflux digest so the rebase doesn’t assume a tag that was never published.

## Implementation

- `konflux_image_builder.update_konflux_db`: always persist pipeline `repo@digest` for SUCCESS/UNRELEASED; remove `released` parameter and second-call kwarg.
- `KonfluxRebaser._resolve_member_parent`: restore unconditional RH for in-group base/golang; add `_registry_pullspec_exists` + RH-then-fallback only on the **late-resolve** + `get_latest_build` path.
- Tests updated for builder and rebaser.

_Add an **ART-** ticket to the title when you have one._

### DIFF
--- a/doozer/doozerlib/backend/konflux_image_builder.py
+++ b/doozer/doozerlib/backend/konflux_image_builder.py
@@ -381,7 +381,7 @@ class KonfluxImageBuilder:
                     if outcome is KonfluxBuildOutcome.UNRELEASED and metadata.should_trigger_base_image_release():
                         base_image_release_success = await self._trigger_base_image_release(metadata, nvr)
                         if base_image_release_success:
-                            logger.info(f"Base image release succeeded for {nvr}, updating with registry pullspec")
+                            logger.info(f"Base image release succeeded for {nvr}, updating build record")
                         else:
                             logger.error(f"Base image release failed for {nvr}, updating build record to FAILURE")
                         outcome = (
@@ -396,7 +396,6 @@ class KonfluxImageBuilder:
                             build_priority,
                             ec_status=ec_status,
                             ec_pipeline_url=ec_pipeline_url,
-                            released=base_image_release_success,
                         )
 
                 if outcome is not KonfluxBuildOutcome.SUCCESS:
@@ -908,7 +907,6 @@ class KonfluxImageBuilder:
         build_priority,
         ec_status=KonfluxECStatus.NOT_APPLICABLE,
         ec_pipeline_url='',
-        released=False,
     ) -> Optional[KonfluxBuildRecord]:
         logger = self._logger.getChild(f"[{metadata.distgit_key}]")
         if not metadata.runtime.konflux_db:
@@ -996,10 +994,7 @@ class KonfluxImageBuilder:
                     f"pipelinerun {pipelinerun_name}"
                 )
 
-            if released:
-                definitive_image_pullspec = util.rh_art_images_base_pullspec(nvr)
-            else:
-                definitive_image_pullspec = f"{image_pullspec.split(':')[0]}@{image_digest}"
+            definitive_image_pullspec = f"{image_pullspec.split(':')[0]}@{image_digest}"
 
             # use image_digest here to be precise, image_pullspec can collide in case of golang-builder images
             package_nvrs, source_rpms = await self.get_installed_packages(
--- a/doozer/doozerlib/backend/rebaser.py
+++ b/doozer/doozerlib/backend/rebaser.py
@@ -22,7 +22,7 @@ from artcommonlib.build_visibility import BuildVisibility, get_visibility_suffix
 from artcommonlib.konflux.konflux_build_record import Engine, KonfluxBuildRecord
 from artcommonlib.model import ListModel, Missing, Model
 from artcommonlib.telemetry import start_as_current_span_async
-from artcommonlib.util import deep_merge, detect_package_managers, is_cachito_enabled
+from artcommonlib.util import deep_merge, detect_package_managers, is_cachito_enabled, oc_image_info_for_arch_async
 from artcommonlib.variants import BuildVariant
 from dockerfile_parse import DockerfileParser
 from doozerlib import constants, util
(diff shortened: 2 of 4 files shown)


doozer/doozerlib/backend/konflux_image_builder.py
```diff
@@ -996,10 +994,7 @@ async def update_konflux_db(
                     f"pipelinerun {pipelinerun_name}"
                 )
 
-            if released:
-                definitive_image_pullspec = util.rh_art_images_base_pullspec(nvr)
-            else:
-                definitive_image_pullspec = f"{image_pullspec.split(':')[0]}@{image_digest}"
+            definitive_image_pullspec = f"{image_pullspec.split(':')[0]}@{image_digest}"
```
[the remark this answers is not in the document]
lgarciaaco: True in the general case (`registry:port/`). This pattern is not introduced only here—we use the same approach elsewhere—and our Konflux `IMAGE_URL` values do not use registry host ports, so we are not changing this in this PR. A proper fix would be a shared helper and updating every call site, not this file alone.

doozer/doozerlib/backend/rebaser.py
```diff
@@ -539,6 +549,16 @@ async def _resolve_member_parent(self, member: str, original_parent: str):
...
+                if parent_metadata.should_trigger_base_image_release():
+                    rh_pullspec = util.rh_art_images_base_pullspec(build.nvr)
+                    if await self._registry_pullspec_exists(rh_pullspec):
+                        return rh_pullspec, build.embargoed
+                    self._logger.info(
+                        "art-images-base %s not reachable; using Konflux image_pullspec for late-resolved parent %s",
+                        rh_pullspec,
+                        member,
+                    )
+                    return build.image_pullspec, build.embargoed
```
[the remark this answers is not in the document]
lgarciaaco: When the value we use is a `registry.redhat.io/...` NVR-style pullspec—whether from `build.image_pullspec` after a failed RH probe or otherwise—we treat that as the canonical released reference and assume it is valid to use for late-resolve. Recovering a historical Konflux digest row or doing a broad backfill is out of scope for this change.

lgarciaaco:
[Jenkins build](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-konflux/34098/console)

doozer/doozerlib/backend/konflux_image_builder.py
locriandev, line 997: No longer need to call this `definitive_image_pullspec`?

doozer/doozerlib/backend/rebaser.py
```diff
@@ -503,6 +503,16 @@ async def _resolve_parents(self, metadata: ImageMetadata, dfp: DockerfileParser)
         private_fix = any(private_fix for _, private_fix in mapped_images)
         return downstream_parents, private_fix
 
+    async def _registry_pullspec_exists(self, pullspec: str) -> bool:
```
locriandev, line 506: makes sense to `@alru_cache` this?

doozer/doozerlib/backend/konflux_image_builder.py
lgarciaaco: it was always called that: https://github.com/openshift-eng/art-tools/pull/2851/changes#diff-e45d70130344e3d45bd59cb091bd0a794121c675ce90f17c5b4e4702ded620ffL978

doozer/doozerlib/backend/rebaser.py
lgarciaaco: Good question — we’re not using `@lru_cache` here on purpose.

1. **Stdlib `lru_cache` and `async def` don’t mix** the way people expect: it ends up caching the coroutine object, not the awaited result. Anything we’d want would need an async-aware helper (`async_lru`-style), `cachetools` + wrapper, or a small hand-rolled memo.

2. **Semantics of this probe:** we’re checking whether an RH NVR-style pullspec is actually readable. Caching a **positive** result for a pullspec is usually fine within a process; caching **negative** results is risky — the tag can show up immediately after a failed probe (release racing with rebase), which is exactly the scenario this path is meant to tolerate. So we’d need *no* negative caching, a **short TTL**, or **dedupe only within the same call graph**, not a blind LRU of “not found.”

3. **If we optimize later:** if profiling shows the same `rh_pullspec` probed many times in one job, a **narrow** dedupe (e.g. an instance dict for the current rebase run, or only memoizing successes) could save redundant `oc image info` without pinning “not found” forever.

So: not a drop-in `@lru_cache`; if we add caching, it should be deliberate and asymmetric (safe for hits, careful on misses).

joepvd (approved):
/lgtm

PR #2888, opened 2026-05-12
ART-18733: fix(doozer): gate base image release on stream assembly

lgarciaaco:
## Summary

`ImageMetadata.should_trigger_base_image_release()` now returns false unless `runtime.assembly == "stream"` (after the existing OKD guard). This aligns with how rebaser and Konflux code treat stream-only behavior.

## Problem

**Before:** Base and golang-builder images used the base-image release path (UNRELEASED outcome, Jenkins snapshot-to-release, registry.redhat.io parent resolution, lockfile seeding with unreleased builds) for every assembly.

**After:** That path runs only for the **stream** assembly. Non-stream assemblies behave like ordinary images for those mechanisms.

## Implementation

- Added `self.runtime.assembly != "stream"` early return in `should_trigger_base_image_release()` (`doozer/doozerlib/image.py`).
- Tests: OCP cases set `assembly='stream'`; added coverage for `assembly='test'`; `_create_image_metadata` sets `assembly='stream'` so existing golang lockfile tests still exercise the UNRELEASED merge.

https://redhat.atlassian.net/browse/ART-18048

### DIFF
--- a/doozer/doozerlib/image.py
+++ b/doozer/doozerlib/image.py
@@ -1395,11 +1395,15 @@ class ImageMetadata(Metadata):
         Determines whether this image should trigger the base image release workflow.
 
         OKD builds never trigger the OCP registry.redhat.io base-image release path.
+        Only the stream assembly uses this path (Konflux UNRELEASED, Jenkins snapshot-to-release,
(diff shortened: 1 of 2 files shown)


locriandev (approved):
/lgtm

PR #2889, opened 2026-05-12
ART-18048: fix(doozer): skip base image release path for test assembly

lgarciaaco:
## Summary

`should_trigger_base_image_release()` now returns false only when `runtime.assembly == "test"` (after the OKD guard). Stream and named assemblies keep the Konflux UNRELEASED / Jenkins snapshot-to-release / RH parent pullspec / lockfile UNRELEASED behavior for base and golang-builder images.

## Problem

**Before:** Only `assembly == stream` used that path (PR #2888).

**After:** Every assembly **except** `test` uses the path, so product assemblies (e.g. `4.17.1`) behave like stream for this mechanism while `test` stays excluded.

## Implementation

- `doozer/doozerlib/image.py`: `assembly == "test"` early return; docstring updated.
- Tests: `test` still negative; added `4.17.1` positive cases; `_create_image_metadata` still sets a non-test assembly for lockfile tests.

https://redhat.atlassian.net/browse/ART-18048

### DIFF
--- a/doozer/doozerlib/image.py
+++ b/doozer/doozerlib/image.py
@@ -1395,7 +1395,7 @@ class ImageMetadata(Metadata):
         Determines whether this image should trigger the base image release workflow.
 
         OKD builds never trigger the OCP registry.redhat.io base-image release path.
-        Only the stream assembly uses this path (Konflux UNRELEASED, Jenkins snapshot-to-release,
+        The test assembly does not use this path (Konflux UNRELEASED, Jenkins snapshot-to-release,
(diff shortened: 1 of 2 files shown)


lgarciaaco:
Test builds:
 - [build 1](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fbase-image-release/6405/console)
 - [build 2](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fbase-image-release/6406/console)

PR #2893, opened 2026-05-12
fix(doozer): skip conflicted inherited dependency check for external RPMs

thegreyd:
slack ref: https://redhat-internal.slack.com/archives/C07RAB68T5G/p1778603952146369?thread_ts=1778602266.618429&cid=C07RAB68T5G

## Summary

- Skip the `conflicted_inherited_dependency` check when the installed RPM is not tagged in given tag in group.yml `check_external_packages` config (ART's candidate Brew tag)
- Add `_is_installed_rpm_in_tag` helper that queries Brew to determine whether the installed RPM actually comes from defined repo
- Applies to both RHCOS and image inspection paths

## Rationale

The `conflicted_inherited_dependency` check was put in place so that if we have `rpm_old` and `rpm_new` tagged in our repo, RHCOS or an image should have `rpm_new` (https://redhat.atlassian.net/browse/ART-8416). This assumes the package is coming from the repo that is defined in the group config (our candidate tag). 

https://github.com/openshift-eng/ocp-build-data/blob/d7c0293dfde19e573e11bc876e590e713f1c03a9/group.yml#L128

But if it comes from another repo (like RHEL), this check becomes invalid — the installed RPM was never sourced from our candidate tag, so flagging a version mismatch against our tagged build is a false positive.

## Test plan

- [x] Unit test for `_is_installed_rpm_in_tag` (RPM tagged, RPM not tagged, None input)
- [ ] Verify with a real assembly inspection that previously noisy false positives are suppressed

### DIFF
--- a/doozer/doozerlib/assembly_inspector.py
+++ b/doozer/doozerlib/assembly_inspector.py
@@ -202,12 +202,20 @@ class AssemblyInspector:
                 build_dict = external_rpms.get(entry["tag"], {}).get(package_name)
                 if not build_dict:
                     continue
+                installed_rpm = installed_packages.get(package_name)
+                if installed_rpm and installed_rpm["nvr"] != build_dict["nvr"]:
+                    if not self._is_installed_rpm_in_tag(installed_rpm, entry["tag"]):
+                        self.runtime.logger.info(
+                            "Installed rpm %s in RHCOS %s (%s) is not tagged in %s; skipping check_external_packages complaint",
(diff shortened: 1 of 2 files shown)


joepvd:
/hold
while this may be an improvement, I would like to find out if this is not a regression in some cases, and if the basic logic makes sense.

joepvd:
/unhold
Sorry, lost track of this. I think this is the correct approach:
/lgtm
/approve

PR #2895, opened 2026-05-13
refactor(pyartcd): make EC verification opt-out via parameter

locriandev:
## Summary
- Replace automatic EC skip for test assemblies with explicit `--skip-ec-verify` parameter
- Changes behavior: EC verification now runs by default for ALL assemblies unless explicitly disabled

## Changes
- Add `skip_ec_verify` parameter to `KonfluxOcpPipeline` (default: `False`)
- Add `--skip-ec-verify` CLI option to `beta:ocp4-konflux` command
- Replace `assembly=='test'` check with `skip_ec_verify` parameter check
- Update log message from "for test assembly" to "as requested"

## Context
This is a follow-up to PR #2892 which added automatic EC skip for test assemblies. The new approach gives operators explicit control over EC validation rather than making assumptions based on assembly name.

## Related
- Companion PR in aos-cd-jobs: https://github.com/openshift-eng/aos-cd-jobs/pull/4688

### DIFF
--- a/pyartcd/pyartcd/pipelines/ocp4_konflux.py
+++ b/pyartcd/pyartcd/pipelines/ocp4_konflux.py
@@ -96,6 +96,7 @@ class KonfluxOcpPipeline:
         skip_rebase: bool = None,
         skip_bundle_build: bool = None,
         skip_build_sync_konflux: bool = False,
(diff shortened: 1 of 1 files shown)


joepvd (approved):
/lgtm

PR #2902, opened 2026-05-13
ART-18751: OCP-only base image registry release trigger

lgarciaaco:
## Summary

Restrict `ImageMetadata.should_trigger_base_image_release()` so the ocp-art-images registry base-image workflow applies only when `runtime.product` is the string `ocp`. Layered products (e.g. `rhmtc`) return false early.

Uses `isinstance(product, str)` so unit tests that leave `runtime.product` as a `MagicMock` keep the previous behavior.

## Testing

`pytest doozer/tests/test_image.py`

### DIFF
--- a/doozer/doozerlib/image.py
+++ b/doozer/doozerlib/image.py
@@ -1419,17 +1419,23 @@ class ImageMetadata(Metadata):
         Determines whether this image should trigger the base image release workflow.
 
         The method checks preconditions in the following order:
-        1. Variant must be OCP (not OKD)
-        2. Assembly must not be ``test``
-        3. Image must be ``base_only`` or a golang builder
-        4. Base image release enabled override:
+        1. Runtime ``product`` must be OCP (when set to a string). Layered products skip the
+           ocp-art-images registry base-image release workflow.
+        2. Variant must be OCP (not OKD)
+        3. Assembly must not be ``test``
+        4. Image must be ``base_only`` or a golang builder
+        5. Base image release enabled override:
            - Image metadata configuration (``self.config.base_image_release.enabled``)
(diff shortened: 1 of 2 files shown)


doozer/doozerlib/image.py
```diff
@@ -1419,17 +1419,23 @@ def should_trigger_base_image_release(self) -> bool:
...
+        5. Base image release enabled override:
            - Image metadata configuration (``self.config.base_image_release.enabled``)
            - Group configuration (``self.runtime.group_config.base_image_release.enabled``)
-        If no override is set for step 4, the enabled flag defaults to True.
+           If no override is set for step 5, the enabled flag defaults to True.
 
         Returns:
             bool: True if base image release workflow should be triggered, False otherwise.
         """
+        rp = getattr(self.runtime, "product", None)
```
locriandev: does Runtime have an explicit `product` field? If so, let's stick to the pattern of referencing `runtime.product` instead.

doozer/doozerlib/image.py
```diff
@@ -1419,17 +1419,23 @@ def should_trigger_base_image_release(self) -> bool:
...
+        if isinstance(rp, str) and rp != "ocp":
```
locriandev: What else could `rp` be? Maybe just None, in this case we could have `if rp and rp != "ocp":`

locriandev (commented):
/lgtm

lgarciaaco:
[test 2](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Flayered-products/3853/)
/approve
/unhold

PR #2905, opened 2026-05-14
use new conforma policies for build time

fgallott:


ashwindasr (approved):
/hold

Until https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/17971 merges

PR #2908, opened 2026-05-15
ART-18904: Remove ineffective task bundle age checking logic

locriandev:
## Problem

The task bundle age validation logic was ineffective because Konflux task bundle images have their `created` timestamp set to the Unix epoch (1970-01-01T00:00:00Z) instead of their actual creation time.

### Evidence

When checking task bundle metadata using `oc image info`:
```bash
$ oc image info quay.io/konflux-ci/tekton-catalog/task-clair-scan@sha256:8fad4c2e2f470f82ee43d6b2ac72327b4d9c6e9cb514a678911c1c9359c29894 -o json | jq .config.created
"1970-01-01T00:00:00Z"
```

### Impact

- The age check always evaluated as "older than 10 days"
- This triggered unnecessary rebuilds when they should be skipped based on age
- The staggered rebuild logic added complexity without providing the intended benefit
- Mass rebuilds occurred unnecessarily

## Solution

Removed the task age checking logic entirely and simplified to: **if task bundle SHAs don't match → rebuild**.

### DIFF
--- a/doozer/doozerlib/cli/scan_sources_konflux.py
+++ b/doozer/doozerlib/cli/scan_sources_konflux.py
@@ -2,7 +2,6 @@ import asyncio
 import json
 import logging
 import os
-import random
 import re
 import tempfile
 from datetime import datetime, timedelta, timezone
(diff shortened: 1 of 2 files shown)


ashwindasr (approved):
/hold

Unhold when ready

PR #2912, opened 2026-05-18
ART-18909: Disable JIRA link unfurling in images-health Slack reports

locriandev:
## Summary
The images-health reports in Slack were becoming difficult to read because JIRA ticket links were being expanded with full preview cards. This change disables unfurling for those links while keeping them clickable.

## Changes
- Added `unfurl_links` parameter to `SlackClient.say()` method (defaults to `True` for backward compatibility)
- Set `unfurl_links=False` for threaded messages in `notify_release_channel()`
- Set `unfurl_links=False` for build concern and rebase failure messages in `notify_forum_ocp_art()`

## Test Plan
- [x] All existing tests pass
- [x] Linting passes
- [ ] Deploy to test environment and verify Slack messages are more readable

## Example
Before: JIRA links expanded with full ticket preview cards, making reports very long
After: JIRA links remain clickable but no longer expand, keeping reports compact

### DIFF
--- a/pyartcd/pyartcd/pipelines/images_health.py
+++ b/pyartcd/pyartcd/pipelines/images_health.py
@@ -308,7 +308,7 @@ class ImagesHealthPipeline:
                     report += f' ({self.url_text(job_url, "Last failure job")})'
                 report += '\n'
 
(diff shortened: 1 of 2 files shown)


joepvd (approved):
/lgtm

PR #2917, opened 2026-05-18
ART-12597: Improve _fix_failed_imagestream_imports

adobes1:
Follow-up to #2913.

## Summary
- Add inner async `get_failed_tags()` helper to encapsulate imagestream fetch and tag filtering
- Add a final recheck after all retry attempts, raising with the count of still-failing tags
- Run `oc import-image` sequentially instead of concurrently to avoid overwhelming the registry
## Test plan
- [ ] Unit tests passing (`uv run pytest pyartcd/tests/pipelines/test_promote.py`)


joepvd (approved):
/lgtm

PR #2919, opened 2026-05-18
ART-18909: Fix JIRA link unfurling by adding unfurl_media parameter

locriandev:
PR #2912 only set unfurl_links=False, but the Slack API requires both unfurl_links and unfurl_media to be set to False to prevent all types of link unfurling.

According to Slack API documentation:
"To prevent links from unfurling, set both unfurl_links and unfurl_media to false when posting messages to stop Slack from crawling a link and attempting to display an unfurl."

Changes:
- Added unfurl_media parameter to SlackClient.say() method
- Set unfurl_media=False for all images-health Slack messages that already had unfurl_links=False
- This should prevent JIRA ticket previews from appearing in health reports

## Test
https://redhat-internal.slack.com/archives/C04L1FPFHNH/p1779109969257409

### DIFF
--- a/pyartcd/pyartcd/pipelines/images_health.py
+++ b/pyartcd/pyartcd/pipelines/images_health.py
@@ -308,7 +308,7 @@ class ImagesHealthPipeline:
                     report += f' ({self.url_text(job_url, "Last failure job")})'
                 report += '\n'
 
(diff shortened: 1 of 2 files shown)


joepvd (commented):
/lgtm

PR #2920, opened 2026-05-18
ART-18938: Add Redis fail counters for EC and base image release errors

locriandev:
## Summary
- Add Redis failure counters for EC (Enterprise Contract / ITS) verification failures and base image release pipeline failures
- Failures are now categorized into three types with separate Redis key patterns:
  - `count:build-failure:konflux:{group}:{image}` — regular build failures (existing)
  - `count:ec-failure:konflux:{group}:{image}` — EC/ITS verification failures (new)
  - `count:release-failure:konflux:{group}:{image}` — base image release failures (new)
- Successfully built images reset all three counter types

### doozer/doozerlib/backend/konflux_image_builder.py
- Add `ec_failed`, `ec_pipeline_url`, and `base_image_release_failed` fields to build record so they appear in `record_log`
- Set `ec_failed=true` and populate `ec_pipeline_url` when EC verification fails
- Set `base_image_release_failed=true` when base image snapshot-release fails

### pyartcd/pyartcd/pipelines/ocp4_konflux.py
- Update `update_build_fail_counters()` to categorize failures by type (build vs EC vs release)
- Each type uses its own Redis key pattern and stores relevant metadata
- Successful builds reset all three counter types

### pyartcd/pyartcd/util.py
- Add `get_ec_failures()` to query EC failure counters from Redis
- Add `get_release_failures()` to query release failure counters from Redis

## Testing
https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-konflux/35686/

image

### DIFF
--- a/doozer/doozerlib/backend/konflux_image_builder.py
+++ b/doozer/doozerlib/backend/konflux_image_builder.py
@@ -131,6 +131,9 @@ class KonfluxImageBuilder:
             "task_url": "n/a",
             "status": -1,  # Status defaults to failure until explicitly set by success. This handles raised exceptions.
             "has_olm_bundle": 1 if metadata.is_olm_operator else 0,
+            "ec_failed": "false",
+            "ec_pipeline_url": "",
+            "base_image_release_failed": "false",
         }
         try:
             if dest_dir.exists():
@@ -336,6 +339,8 @@ class KonfluxImageBuilder:
                     ec_failed = ec_result.ec_failed
                     if ec_failed:
                         outcome = KonfluxBuildOutcome.FAILURE
+                        record["ec_failed"] = "true"
+                        record["ec_pipeline_url"] = ec_pipeline_url
 
                 elif outcome is KonfluxBuildOutcome.SUCCESS:
                     if self._config.skip_ec_verify:
@@ -368,6 +373,7 @@ class KonfluxImageBuilder:
                             logger.info("Base image release succeeded for %s, persisting build record", nvr)
                         else:
                             logger.error("Base image release failed for %s, persisting build record as FAILURE", nvr)
+                            record["base_image_release_failed"] = "true"
                         released_outcome = (
                             KonfluxBuildOutcome.SUCCESS if base_image_release_success else KonfluxBuildOutcome.FAILURE
                         )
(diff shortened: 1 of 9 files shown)


pyartcd/pyartcd/util.py
```diff
@@ -1171,16 +1171,29 @@ async def get_rebase_failures(version: str, branches: list[str], build_systems:
...
         logger (Logger): Optional logger for debugging
     Return Value(s):
         dict: {image_name: {failure_count, <all_metadata>}}
     """
-    pattern = f'count:build-failure:{build_system}:{group}:*:failure'
+    pattern = f'count:{counter_type}:{build_system}:{group}:*:failure'
     return await get_failures(pattern, entity_index=-2, logger=logger)
+
+
+async def get_build_failures(group: str, build_system: str = 'konflux', logger=None):
```
lgarciaaco: Does it make sense a method wrapping a one liner?

doozer/doozerlib/backend/konflux_image_builder.py
```diff
@@ -131,6 +131,9 @@ async def build(self, metadata: ImageMetadata, git_auth_secret: Optional[str] =
             "task_url": "n/a",
             "status": -1,  # Status defaults to failure until explicitly set by success. This handles raised exceptions.
             "has_olm_bundle": 1 if metadata.is_olm_operator else 0,
+            "ec_failed": "false",
+            "ec_pipeline_url": "",
+            "base_image_release_failed": "false",
```
lgarciaaco, line 136: I wonder if we also need released_pipeline (which we will have with #2944).`base_image_release_failed` could be shorter `released_failed` to keep aligned with #2944

pyartcd/pyartcd/util.py
locriandev: I think so, you never know when things will change that will make us regret not having one.

doozer/doozerlib/backend/konflux_image_builder.py
locriandev: Sure thing, I will follow up after #2944 merges. let's not block on this, I can already start consuming some data

pyartcd/pyartcd/util.py
lgarciaaco: We can always build one when there is enough logic / content to create one. Not a fan of having unnecessary hops

PR #2935, opened 2026-05-20
ART-18958: fix get_inflight() selecting future Y-1

rayfordj:
fix get_inflight() selecting future-dated Y-1 releases

get_inflight() used is_future_release_date() combined with abs(days_diff) <= 5, which could select a Y-1 release scheduled *after* the assembly date — injecting a non-existent release into upgrade edges.

Replace the algorithm to only consider Y-1 releases with date_start <= assembly_release_date, picking the latest one within a 7-day window. This ensures we never reference a release that doesn't yet exist while still capturing "in-flight" releases shipping in the same week.

Signed-off-by: Rayford Johnson <rayfordj@users.noreply.github.com>

rh-pre-commit.version: 2.4.0

### DIFF
--- a/artcommon/artcommonlib/util.py
+++ b/artcommon/artcommonlib/util.py
@@ -334,11 +334,18 @@ def is_release_next_week(group):
 
 def get_inflight(assembly, group, date=None):
     """
-    Get inflight release name from current assembly release
+    Get inflight release name from current assembly release.
+
+    Searches the previous minor's release schedule for a Y-1 z-stream release
+    whose date_start falls within 7 days before (or on the same day as) the
+    assembly release date. When multiple releases qualify, the latest is chosen.
+    A same-day release is included because it is expected to ship by the time
+    the assembly reaches customers.
 
     :param assembly: Assembly name
     :param group: Group name
     :param date: Optional explicitly provided date in format YYYY-MMM-DD (e.g., 2026-Mar-31)
+    :return: Version string (e.g., '4.21.16') of the in-flight release, or None
     """
     inflight_release = None
     assembly_release_date = get_assembly_release_date(assembly, group, date)
@@ -357,22 +364,26 @@ def get_inflight(assembly, group, date=None):
         response.raise_for_status()
         try:
             data = response.json()
+            assembly_date = datetime.strptime(assembly_release_date, "%Y-%b-%d")
+            best_candidate = None
+            best_date = None
+
             for release in data['all_ga_tasks']:
-                is_future = is_future_release_date(release['date_start'])
-                if is_future:
-                    days_diff = abs(
-                        (
-                            datetime.strptime(assembly_release_date, "%Y-%b-%d")
-                            - datetime.strptime(release['date_start'], "%Y-%m-%d")
-                        ).days
-                    )
-                    if days_diff <= 5:  # if next Y-1 release and assembly release in the same week
-                        match = re.search(r'\d+\.\d+\.\d+', release['name'])
-                        if match:
-                            inflight_release = match.group()
-                            break
-                        else:
-                            raise ValueError(f"Didn't find in_inflight release in {release['name']}")
+                release_date = datetime.strptime(release['date_start'], "%Y-%m-%d")
+                if release_date > assembly_date:
+                    continue
+                days_diff = (assembly_date - release_date).days
+                if days_diff <= 7:  # if next Y-1 release and assembly release in the same week
+                    match = re.search(r'\d+\.\d+\.\d+', release['name'])
+                    if match:
+                        if best_date is None or release_date > best_date:
+                            best_candidate = match.group()
+                            best_date = release_date
+                    else:
+                        LOGGER.warning(f"Release '{release['name']}' qualified but has no X.Y.Z version in its name")
+
+            if best_candidate:
+                inflight_release = best_candidate
         except json.JSONDecodeError as e:
             raise ValueError(f'Failed to parse JSON for {prev_group}: {e}')
         except ValueError as e:
--- a/artcommon/artcommonlib/util.py
+++ b/artcommon/artcommonlib/util.py
@@ -337,14 +337,16 @@ def get_inflight(assembly, group, date=None):
     Get inflight release name from current assembly release.
 
     Searches the previous minor's release schedule for a Y-1 z-stream release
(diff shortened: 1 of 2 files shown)


rayfordj:
@coderabbitai full review

artcommon/artcommonlib/util.py
```diff
@@ -332,11 +332,18 @@ def is_release_next_week(group):
 
 def get_inflight(assembly, group, date=None):
     """
-    Get inflight release name from current assembly release
+    Get inflight release name from current assembly release.
+
+    Searches the previous minor's release schedule for a Y-1 z-stream release
+    whose date_start falls within 7 days before (or on the same day as) the
+    assembly release date. When multiple releases qualify, the latest is chosen.
```
wking: No need to limit to the last one, is there?  If we move to a schedule where we build a new release for both x.y and x.(y-1) each day, but it still takes a few days for them to work through to their GA/Errata release, we could have multiple in-flight x.y and x.(y-1) patch releases that hadn't arrived in `candidate-x.y` or `candidate-x.(y-1)` yet, but which we wanted to include in the new x.y.z's update sources.

artcommon/artcommonlib/util.py
```diff
@@ -355,22 +362,26 @@ def get_inflight(assembly, group, date=None):
...
-                    if days_diff <= 5:  # if next Y-1 release and assembly release in the same week
-                        match = re.search(r'\d+\.\d+\.\d+', release['name'])
-                        if match:
-                            inflight_release = match.group()
-                            break
-                        else:
-                            raise ValueError(f"Didn't find in_inflight release in {release['name']}")
+                release_date = datetime.strptime(release['date_start'], "%Y-%m-%d")
+                if release_date > assembly_date:
+                    continue
```
wking, line 376: There are two competing risks we're trying to manage:

* If we leave off an older release from the update-source list, folks on that release will have to wait longer before they can update to the next major/minor release and get those new features.  Slightly annoying for them if they're trying to ride recent patches in their current x.y stream.  But unlikely to be a huge deal.
* If we add a newer release to the update-source list, we risk undocumented regressions as the customer updates from the newer x.y.z release to the older x.(y-1).z' release.

Previously, `days_diff = abs(...)` was reaching further into the future, and pulling 4.21.17 (which doesn't exist yet) in as an update source for 4.22.0-rc.4:

```console
$ oc adm release info quay.io/openshift-release-dev/ocp-release:4.22.0-rc.4-x86_64 | grep Upgrades
  Upgrades: 4.21.6, ..., 4.21.15, 4.21.17, 4.22.0-ec.0, ..., 4.22.0-rc.3
```

So +1 from me on removing that ability to reach 5d into the future.  However, trying to work this through some more examples, [4.21.16 shipped its errata on the 19th][1], and [4.20.23 shipped on the 20th][2].  And 4.21.16 includes 4.20.23 in its update sources:

```console
$ oc adm release info quay.io/openshift-release-dev/ocp-release:4.21.16-x86_64 | grep Upgrades
  Upgrades: 4.20.15, ..., 4.20.22, 4.20.23, 4.21.0, ..., 4.21.15
```

Despite that release order, 4.20.23 was built an hour earlier (at least for x86_64):

```console
$ oc image info -o json quay.io/openshift-release-dev/ocp-release:4.21.16-x86_64 | jq -r '.config.created'
2026-05-14T10:50:57Z
$ oc image info -o json quay.io/openshift-release-dev/ocp-release:4.20.23-x86_64 | jq -r '.config.created'
2026-05-14T09:19:37Z
```

I haven't checked the release contents to have an opinion on whether 4.20.23 has fixes that 4.21.16 lacks (if it does, it would ideally not include 4.20.23 in its update sources).  But that kind of wiggle in build time and release date makes me a bit nervous about excluding update-source-releases with planned release dates that come after the expected release data of the image we're building.  Maybe we need a way to reach one day into the future, or some such?

And ideally we'd have a way to reliably track "that x.(y-1).z update-source might have some fixes I don't have", without having to put so much weight in the date we expected to push an Errata on.  But I realize that would be a larger pivot, and have no problem tightening things up here as best we can to try and get some cheaper wins, even if the more robust fixes are still out in the future.

[1]: https://access.redhat.com/errata/RHSA-2026:17474
[2]: https://access.redhat.com/errata/RHSA-2026:17468

artcommon/artcommonlib/util.py
```diff
@@ -332,11 +332,18 @@ def is_release_next_week(group):
 
 def get_inflight(assembly, group, date=None):
     """
-    Get inflight release name from current assembly release
+    Get inflight release name from current assembly release.
+
+    Searches the previous minor's release schedule for a Y-1 z-stream release
+    whose date_start falls within 7 days before (or on the same day as) the
```
joepvd: `date_start` is an unknown concept to me. Is that "dev cutoff"?

artcommon/artcommonlib/util.py
```diff
@@ -332,11 +332,18 @@ def is_release_next_week(group):
...
+    assembly release date. When multiple releases qualify, the latest is chosen.
+    A same-day release is included because it is expected to ship by the time
+    the assembly reaches customers.
 
     :param assembly: Assembly name
     :param group: Group name
     :param date: Optional explicitly provided date in format YYYY-MMM-DD (e.g., 2026-Mar-31)
+    :return: Version string (e.g., '4.21.16') of the in-flight release, or None
```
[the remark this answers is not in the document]
rayfordj: sufficiently addressed in latest PR changes?

artcommon/artcommonlib/util.py
```diff
@@ -355,22 +362,26 @@ def get_inflight(assembly, group, date=None):
...
+                    match = re.search(r'\d+\.\d+\.\d+', release['name'])
+                    if match:
+                        if best_date is None or release_date > best_date:
+                            best_candidate = match.group()
+                            best_date = release_date
+                    else:
+                        LOGGER.warning(f"Release '{release['name']}' qualified but has no X.Y.Z version in its name")
+
+            if best_candidate:
+                inflight_release = best_candidate
```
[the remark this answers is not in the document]
rayfordj: While theoretically correct, practically unlikely; existing code should be correct for currently expected realistic inputs.

ashwindasr (approved):

PR #2939, opened 2026-05-21
ART-14750: Add golang plashet + signing support to update-golang pipeline

adobes1:
## Summary

Golang-builder images currently install golang RPMs from brewroot repos, which only serve unsigned copies. This PR ensures signed RPMs are used by switching the pipeline to build plashet repos (which serve signed RPMs) and enforcing signing before proceeding.

The `update-golang` pipeline is extended with two new steps that run after the golang RPM builds are tagged and available:

1. **Signing enforcement** — checks whether the golang RPM is already signed. For sustaining golang versions (1.20, 1.22, 1.23, 1.24), which are signed by the ART team rather than through the standard RHEL process, the pipeline tags the build into a `rhaos-{version}-rhel-{el}-golang` brew tag to trigger auto-signing. For RHEL-shipped versions (1.25+), the RPM must already be signed or the pipeline fails.

2. **Plashet build** — after signing, triggers the `build-plashets` Jenkins job to build a plashet repo containing the signed golang RPMs. Golang-builder images will then install from this plashet instead of brewroot.

Whether to sign a given golang version is controlled by a new `sign_golang_rpm` field in the ocp-build-data golang branch `group.yml` (defaults to `false`). The companion ocp-build-data PRs (which also switch the repo URLs from brewroot to plashet) are tracked [here](https://github.com/openshift-eng/ocp-build-data/pulls?q=is%3Aopen+is%3Apr+author%3Aadobes1+[ART-14750](https://redhat.atlassian.net/browse/ART-14750)).

## TODO before merge
- [ ] Merge ocp-build-data PRs that add `sign_golang_rpm` and plashet URLs to golang branches

## Test plan
- [ ] Run `uv run pytest pyartcd/tests/pipelines/test_update_golang.py` — all tests pass
- [ ] Verify `ensure_signed` correctly identifies unsigned sustaining golang RPMs and tags them for signing
- [ ] Verify `should_sign_golang_rpm` reads `sign_golang_rpm` from the correct ocp-build-data branch (including fork overrides via `--data-path`/`--data-gitref`)
- [ ] Verify `_build_golang_plashets` triggers the `build-plashets` job and waits for completion

## Summary by CodeRabbit

* **New Features**
  * Added the ability to configure plashet-based repos to include specific packages.
  * Extended golang release processing to trigger additional plashet builds per RHEL version, with automated signing when enabled.
  * Enhanced build configuration resolution to support alternate content repository and ref selection.

* **Bug Fixes**
  * Updated signing-key handling to accept an additional valid key.
  * Improved fallback behavior for missing group version details and tightened golang content URL validation.
  * Adjusted plashet command generation to properly include packages and signing advisory flags.

* **Tests**
  * Added coverage for golang RPM signing, dry-run behavior, and plashet build success/failure handling.

### DIFF
--- a/pyartcd/pyartcd/pipelines/update_golang.py
+++ b/pyartcd/pyartcd/pipelines/update_golang.py
@@ -1064,14 +1064,14 @@ class UpdateGolangPipeline:
             )
 
         major, minor = group_config['vars']['MAJOR'], group_config['vars']['MINOR']
-        content_repo_url_suffix = self.get_content_repo_url_suffix(el_v, major, minor)
+        expected_suffixes = self.get_expected_golang_url_suffixes(el_v, major, minor)
         err = False
         for arch, template_url in group_config['repos'][golang_repo]['conf']['baseurl'].items():
-            expected_suffix = f'{content_repo_url_suffix}/{arch}/os/'
             actual_url = template_url.format(MAJOR=major, MINOR=minor)
-            if not actual_url.endswith(expected_suffix):
+            if not any(actual_url.endswith(f'{suffix}/{arch}/') or actual_url.endswith(f'{suffix}/{arch}/os/')
+                       for suffix in expected_suffixes):
                 err = True
-                _LOGGER.error(f"{expected_suffix} not found in URL {actual_url}")
+                _LOGGER.error(f"URL {actual_url} does not match any expected pattern for arch {arch}")
 
         if err:
             raise ValueError(
@@ -1080,8 +1080,10 @@ class UpdateGolangPipeline:
 
         _LOGGER.info(f"Builder branch {branch} has the expected content set urls")
 
-    def get_content_repo_url_suffix(self, el_v, major, minor):
-        return f'/pub/RHOCP/plashets/{major}.{minor}/stream/golang-el{el_v}/latest'
+    def get_expected_golang_url_suffixes(self, el_v, major, minor):
+        brewroot_suffix = f'/brewroot/repos/rhaos-{self.ocp_version}-rhel-{el_v}-build/latest'
+        plashet_suffix = f'/pub/RHOCP/plashets/{major}.{minor}/stream/golang-el{el_v}/latest'
+        return [brewroot_suffix, plashet_suffix]
 
     def get_module_tag(self, nvr, el_v) -> str:
         tags = [t['name'] for t in self.koji_session.listTags(build=nvr)]
--- a/pyartcd/pyartcd/pipelines/update_golang.py
+++ b/pyartcd/pyartcd/pipelines/update_golang.py
@@ -23,7 +23,7 @@ from artcommonlib.konflux.konflux_db import KonfluxDb
 from artcommonlib.release_util import isolate_el_version_in_release
 from artcommonlib.rpm_utils import parse_nvr
 from artcommonlib.util import new_roundtrip_yaml_handler
-from doozerlib.backend.base_image_handler import ART_IMAGES_BASE_APPLICATION
+from doozerlib.constants import ART_IMAGES_BASE_APPLICATION
 from doozerlib.cli.config_plashet import KNOWN_SIGNING_KEYS
 from elliottlib import util as elliottutil
 from elliottlib.constants import GOLANG_BUILDER_CVE_COMPONENT
@@ -495,11 +495,10 @@ class UpdateGolangPipeline:
 
     async def process_build(self, el_v, nvr):
         await self.ensure_signed(el_v, nvr)
-        if await is_latest_and_available(self.ocp_version, el_v, nvr, self.koji_session):
-            return True
-        if not self.tag_builds:
-            return False
-        await self.tag_build(el_v, nvr)
+        if not is_latest(self.ocp_version, el_v, nvr, self.koji_session):
+            if not self.tag_builds:
+                return False
+            await self.tag_build(el_v, nvr)
 
         # retry 5 times to request repo regen and check availability,
         # which can take around 5 hours in worst case, before giving up and raise error
--- a/pyartcd/pyartcd/pipelines/update_golang.py
+++ b/pyartcd/pyartcd/pipelines/update_golang.py
@@ -233,6 +233,18 @@ class UpdateGolangPipeline:
     def _get_upstream_ocp_build_data_repo(self):
         return get_github_client_for_org("openshift-eng").get_repo("openshift-eng/ocp-build-data")
 
+    def _get_ocp_build_data_repo_and_branch(self, default_branch):
+        """Get the ocp-build-data repo and branch, respecting data_path/data_gitref overrides."""
+        if self.data_path:
+            match = re.search(r'github\.com[:/](.+?)(?:\.git)?$', self.data_path)
+            if match:
+                repo_name = match.group(1)
+                org = repo_name.split('/')[0]
+                repo = get_github_client_for_org(org).get_repo(repo_name)
+                branch = self.data_gitref or default_branch
+                return repo, branch
+        return self._get_upstream_ocp_build_data_repo(), default_branch
+
     def _get_branch_content(self):
         if self._branch_content is None:
             branch = f"openshift-{self.ocp_version}"
@@ -574,8 +586,8 @@ class UpdateGolangPipeline:
         """Check sign_golang_rpm in the golang branch group.yml.
         Defaults to False so that RHEL-shipped golang is never accidentally signed by us.
         """
-        branch = self.get_golang_branch(el_v, go_version)
-        repo = self._get_upstream_ocp_build_data_repo()
+        default_branch = self.get_golang_branch(el_v, go_version)
+        repo, branch = self._get_ocp_build_data_repo_and_branch(default_branch)
         content = repo.get_contents("group.yml", ref=branch)
         group_config = yaml.load(content.decoded_content)
         return group_config.get('sign_golang_rpm', False)
@@ -1046,12 +1058,12 @@ class UpdateGolangPipeline:
 
     def verify_golang_builder_repo(self, el_v, go_version):
         if self.use_new_golang_branch:
-            branch = self.GOLANG_DATA_BRANCH
+            default_branch = self.GOLANG_DATA_BRANCH
         else:
-            branch = self.get_golang_branch(el_v, go_version)
+            default_branch = self.get_golang_branch(el_v, go_version)
         filename = 'group.yml'
 
-        repo = get_github_client_for_org("openshift-eng").get_repo("openshift-eng/ocp-build-data")
+        repo, branch = self._get_ocp_build_data_repo_and_branch(default_branch)
         content = repo.get_contents(filename, ref=branch)
         group_config = yaml.load(content.decoded_content)
 
@@ -1064,14 +1076,14 @@ class UpdateGolangPipeline:
             )
 
         major, minor = group_config['vars']['MAJOR'], group_config['vars']['MINOR']
-        expected_suffixes = self.get_expected_golang_url_suffixes(el_v, major, minor)
+        content_repo_url_suffix = self.get_content_repo_url_suffix(el_v, major, minor)
         err = False
         for arch, template_url in group_config['repos'][golang_repo]['conf']['baseurl'].items():
+            expected_suffix = f'{content_repo_url_suffix}/{arch}/os/'
             actual_url = template_url.format(MAJOR=major, MINOR=minor)
-            if not any(actual_url.endswith(f'{suffix}/{arch}/') or actual_url.endswith(f'{suffix}/{arch}/os/')
-                       for suffix in expected_suffixes):
+            if not actual_url.endswith(expected_suffix):
                 err = True
-                _LOGGER.error(f"URL {actual_url} does not match any expected pattern for arch {arch}")
+                _LOGGER.error(f"{expected_suffix} not found in URL {actual_url}")
 
         if err:
             raise ValueError(
@@ -1080,10 +1092,8 @@ class UpdateGolangPipeline:
 
         _LOGGER.info(f"Builder branch {branch} has the expected content set urls")
 
-    def get_expected_golang_url_suffixes(self, el_v, major, minor):
-        brewroot_suffix = f'/brewroot/repos/rhaos-{self.ocp_version}-rhel-{el_v}-build/latest'
-        plashet_suffix = f'/pub/RHOCP/plashets/{major}.{minor}/stream/golang-el{el_v}/latest'
-        return [brewroot_suffix, plashet_suffix]
+    def get_content_repo_url_suffix(self, el_v, major, minor):
+        return f'/pub/RHOCP/plashets/{major}.{minor}/stream/golang-el{el_v}/latest'
 
     def get_module_tag(self, nvr, el_v) -> str:
         tags = [t['name'] for t in self.koji_session.listTags(build=nvr)]
--- a/pyartcd/pyartcd/pipelines/update_golang.py
+++ b/pyartcd/pyartcd/pipelines/update_golang.py
@@ -334,7 +334,7 @@ class UpdateGolangPipeline:
         el_nvr_map_for_images = {el_v: nvr for el_v, nvr in el_nvr_map.items() if el_v != 10}
         if el_nvr_map.keys() - el_nvr_map_for_images.keys():
             _LOGGER.info("RHEL 10 NVRs will only be used for build root tagging, not for golang-builder images")
-        self._slack_client.dry_run = True
+        self._slack_client.bind_channel(self.ocp_version)
         running_in_jenkins = os.environ.get('BUILD_ID', False)
         if running_in_jenkins:
             title_update = f" {self.ocp_version} - {go_version} - el{list(el_nvr_map.keys())} - {self.build_system}"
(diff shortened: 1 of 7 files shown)


pyartcd/pyartcd/pipelines/update_golang.py
```diff
@@ -243,6 +245,18 @@ def _load_yaml_from_repo(repo, path: str, ref: str):
...
+        """Get the ocp-build-data repo and branch, respecting data_path/data_gitref overrides."""
+        if self.data_path:
+            match = re.search(r'github\.com[:/](.+?)(?:\.git)?$', self.data_path)
+            if match:
+                repo_name = match.group(1)
+                org = repo_name.split('/')[0]
+                repo = get_github_client_for_org(org).get_repo(repo_name)
+                branch = self.data_gitref or default_branch
+                return repo, branch
+        return self._get_upstream_ocp_build_data_repo(), default_branch
```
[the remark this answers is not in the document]
adobes1: I think this validation should continue to be done against the upstream repo+branch to avoid accidentally tagging wrong rpms into the brew tags (such as when the fork's contents are outdated).

pyartcd/pyartcd/pipelines/update_golang.py
```diff
@@ -530,6 +533,71 @@ async def tag_build(self, el_v, nvr):
...
+        brewroot = Path('/mnt/redhat/brewroot')
+        build_path = brewroot / 'packages' / parsed_nvr['name'] / parsed_nvr['version'] / parsed_nvr['release']
+        for key in KNOWN_SIGNING_KEYS:
+            if (build_path / 'data' / 'signed' / key).exists():
+                return True
+        return False
+
+    def should_sign_golang_rpm(self, el_v, go_version) -> bool:
+        """Check sign_golang_rpm in the golang branch group.yml.
+        Defaults to False so that RHEL-shipped golang is never accidentally signed by us.
```
thegreyd, line 586: i think there can be a check for this - check tags of rpm - if it doesn't have a tag like rhba/rhsa... -pending / -released ?

thegreyd:
Nice work, let's plan to test this flow via builder job and plashet job

pyartcd/pyartcd/pipelines/update_golang.py
adobes1: I don't think we can use the `RHBA-...`/`RHSA-...` tags as the RPM is only tagged there once attached to the advisory iiuc. Some RPMs might not have those tags but are still supposed to go out through RHEL (so we shouldn't sign them).
For example, [golang-1.26.3-1.el8_10](https://brewweb.engineering.redhat.com/brew/buildinfo?buildID=4021502) will be released in RHEL but it hasn't been attached to an advisory yet so it doesn't have those tags. 
It is tagged into `rhel-8.10.0-z-gate` but I don't think builds are tagged into these `-gate` tags right after finishing so there would still be a window where we could accidentally sign a wrong build.

But I might be completely wrong, RHEL tagging process is not very clear to me

thegreyd: Hmm so how would we know which rpm to sign and which to not..

adobes1: Unfortunately, I did not find a way to do this automatically or per RPM build. The way I set `sign_golang_rpm` fields in the ocp-build-data PRs (such as [here](https://github.com/openshift-eng/ocp-build-data/pull/10763/changes#diff-005fe59037264b0a0c3005a1c55f7043bdfcfab6c740a26bf8d9cb9ea3750f88R13)) was according to [this sustaining doc](https://redhat.atlassian.net/wiki/spaces/SustainingEngineering/pages/343770677/Supportability+of+Go+versions). I think that once RHEL stops supporting given golang version, we can assume that all future builds are from sustaining and can be signed

pyartcd/pyartcd/pipelines/update_golang.py
```diff
@@ -531,6 +549,71 @@ async def tag_build(self, el_v, nvr):
...
+            )
+
+    @staticmethod
+    def is_rpm_signed(parsed_nvr):
+        brewroot = Path('/mnt/redhat/brewroot')
+        build_path = brewroot / 'packages' / parsed_nvr['name'] / parsed_nvr['version'] / parsed_nvr['release']
+        for key in KNOWN_SIGNING_KEYS:
+            if (build_path / 'data' / 'signed' / key).exists():
+                return True
+        return False
```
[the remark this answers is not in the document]
adobes1: build-plashet job checks again (and more comprehensively) if the rpms are signed and proceeds only if they are

pyartcd/pyartcd/pipelines/update_golang.py
```diff
@@ -332,7 +346,7 @@ async def run(self):
         el_nvr_map_for_images = {el_v: nvr for el_v, nvr in el_nvr_map.items() if el_v != 10}
         if el_nvr_map.keys() - el_nvr_map_for_images.keys():
             _LOGGER.info("RHEL 10 NVRs will only be used for build root tagging, not for golang-builder images")
-        self._slack_client.bind_channel(self.ocp_version)
+        self._slack_client.dry_run = True
```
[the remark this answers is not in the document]
thegreyd: @adobes1 is this a leftover?

pyartcd/pyartcd/pipelines/update_golang.py
```diff
@@ -531,6 +549,71 @@ async def tag_build(self, el_v, nvr):
...
+
+    def should_sign_golang_rpm(self, el_v, go_version) -> bool:
+        """Check sign_golang_rpm in the golang branch group.yml.
+        Defaults to False so that RHEL-shipped golang is never accidentally signed by us.
+        """
+        default_branch = self.get_golang_branch(el_v, go_version)
+        repo, branch = self._get_ocp_build_data_repo_and_branch(default_branch)
+        content = repo.get_contents("group.yml", ref=branch)
+        group_config = yaml.load(content.decoded_content)
+        return group_config.get('sign_golang_rpm', False)
```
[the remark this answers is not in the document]
thegreyd: @adobes1 this looks like a valid suggestion

pyartcd/pyartcd/pipelines/update_golang.py
```diff
@@ -516,6 +533,71 @@ async def tag_build(self, el_v, nvr):
...
+            if self.dry_run:
+                _LOGGER.info("[DRY RUN] Would have triggered build-plashets for %s", group)
+                continue
+            result = jenkins.start_build_plashets(
+                group=group,
+                release=default_release_suffix(),
+                assembly="stream",
+                repos=[repo_name],
+                block_until_complete=True,
+            )
```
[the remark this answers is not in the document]
thegreyd: it is fine, build-data override is for testing or hotfix purposes and this might be overkill - if it is needed then in future we can put it in

pyartcd/pyartcd/pipelines/update_golang.py
```diff
@@ -1008,9 +1091,8 @@ def verify_golang_builder_repo(self, el_v, go_version):
 
         _LOGGER.info(f"Builder branch {branch} has the expected content set urls")
 
-    def get_content_repo_url_suffix(self, el_v):
-        url = '/brewroot/repos/{repo}/latest'
-        return url.format(repo=f'rhaos-{self.ocp_version}-rhel-{el_v}-build')
+    def get_content_repo_url_suffix(self, el_v, major, minor):
+        return f'/pub/RHOCP/plashets/{major}.{minor}/stream/golang-el{el_v}/latest'
```
thegreyd: nit: self.ocp_version is already major.minor and can be used?

pyartcd/pyartcd/plashets.py
```diff
@@ -143,6 +143,24 @@ def plashet_config_for_major_minor(major, minor):
             "embargoed_tags": [],
             "include_previous_packages": [],
         },
+        "rhel-9-golang-rpms": {
```
thegreyd, line 155: let's put in a rhel-10 as well, since soon we'll have that requirement

pyartcd/pyartcd/pipelines/update_golang.py
```diff
@@ -531,6 +549,71 @@ async def tag_build(self, el_v, nvr):
...
+        content = repo.get_contents("group.yml", ref=branch)
+        group_config = yaml.load(content.decoded_content)
+        return group_config.get('sign_golang_rpm', False)
+
+    async def _build_golang_plashets(self, go_version, el_versions):
+        go_v = ".".join(go_version.split(".")[:2])
+        _LOGGER.info("Building golang plashets for go %s, RHEL versions: %s", go_v, list(el_versions))
+
+        for el_v in el_versions:
+            group = f"rhel-{el_v}-golang-{go_v}"
```
thegreyd, line 599: hm I don't understand why we're building plashet for this group instead of self.ocp_version ?

pyartcd/pyartcd/plashets.py
```diff
@@ -253,10 +272,15 @@ async def build_plashets(
         plashet_config = PlashetConfig.model_validate(group_config.get('plashet', {}))
 
     else:  # Fall back to old-style config
-        # old-style config only supports openshift-x.y groups
         major, minor = isolate_major_minor_in_group(group)
         if major is None:
-            raise ValueError("Old-style plashet config only supports openshift-x.y groups")
```
thegreyd, line 259: i don't think we need to support non openshift-x.y groups here, unless i'm missing something .

pyartcd/pyartcd/plashets.py
```diff
@@ -281,7 +305,7 @@ async def build_plashets(
     logger.info("Building plashet repos: %s", ", ".join(plashet_repo_names))
 
     # Check release state
-    signing_mode = await util.get_signing_mode(group_config=group_config)
+    signing_mode = await util.get_signing_mode(group=group, group_config=group_config)
```
thegreyd, line 317: what's the impact of this change?

pyartcd/pyartcd/pipelines/update_golang.py
thegreyd, line 592: we would need to think about how "sign_golang_rpm" would live in mono golang branch. 
https://github.com/openshift-eng/ocp-build-data/tree/golang

group.yml file will be shared by all builder configs, and so we would need to have an override in the image config.

also a suggestion on the wording to make it more clear - `use_signed_golang_rpm`

thegreyd: also if you could do a dry-run with mono golang branch to make sure nothing else breaks - that'll be great

pyartcd/pyartcd/pipelines/update_golang.py
thegreyd: okay we can figure this out as we move along in the process

pyartcd/pyartcd/pipelines/update_golang.py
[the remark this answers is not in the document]
adobes1: @coderabbitai The `yaml.load()` does not come for PyYAML but from ruamel with `YAML(typ='rt')`. This pattern is used in other places in our codebase when reading from forks outside of the `openshift-eng` or `openshift` orgs. Would you still recommend switching to a safe loader or is `rt` enough?

pyartcd/pyartcd/pipelines/update_golang.py
adobes1: Thanks, fixed in https://github.com/openshift-eng/art-tools/pull/2939/changes/5f62237cffe7c22c2ee9c4ace1a61960ded73d51#diff-3d94c58ebda6cca16398fc503f6313a39a8636b63a3cbb07567e550c261d2518R1094-R1095

pyartcd/pyartcd/pipelines/update_golang.py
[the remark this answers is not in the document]
adobes1: Yes, I left it there in case we wanted to do more live runs. Removed in https://github.com/openshift-eng/art-tools/pull/2939/changes/5f62237cffe7c22c2ee9c4ace1a61960ded73d51

pyartcd/pyartcd/plashets.py
adobes1: Good idea! Added in https://github.com/openshift-eng/art-tools/pull/2939/changes/03b92ec877731c59acd8e6a3a38996c3f1b87753

pyartcd/pyartcd/pipelines/update_golang.py
adobes1: The `build-plashets` job selects repos to build based on (among other factors) which repos are defined in the `repos` section in `group.yml` (in the "old plashet config style") of the group it receives as a parameter. Since the `openshift-golang-builder` metadata files are in the `rhel-{el_v}-golang-{go_v}` branches, we need to have `rhel-{el_v}-golang-rpms` defined in those branches' group.yml so that we can pass this repo to `openshift-golang-builder` in `enabled_repos`. When we pass `rhel-{el_v}-golang-{go_v}` group to the `build-plashets` job, it sees the golang repo in group.yml and decides to build it.

If we wanted to build the plashet for `openshift-x.y`, we'd need to have duplicate definitions of the same repo:
- `rhel-{el_v}-golang-rpms` in `openshift-x.y` group.yml so that the plashet job builds it
- `rhel-{el_v}-golang-rpms` in `rhel-{el_v}-golang-{go_v}` branches to be able to pass it in `enabled_repos` and inject it into the build.

Hoping I understood your question correctly

EDIT: Or for the `openshift-x.y` case we could modify the build-plashets code to skip the lookup of repos in group.yml when a list of plashets to build is explicitly passed (and match directly against the hardcoded plashet config in `plashet_config_for_major_minor()` instead) 🤔 Would need to check what impact this would have on the rest of our codebase

pyartcd/pyartcd/plashets.py
adobes1: I think this is necessary as long as we build the plashet against the `rhel-{el_v}-golang-{go_v}` branches (further discussed [here](https://github.com/openshift-eng/art-tools/pull/2939#discussion_r3467334474)). If we decide to go forward with a different approach, this can be removed

pyartcd/pyartcd/plashets.py
adobes1: The `group.yml` in golang branches doesn't have a `software_lifecycle` section. Without this change `get_signing_mode` would throw a `KeyError` when trying to read `group_config['software_lifecycle']['phase']`. Passing `group` allows `get_signing_mode` to short-circuit for non-`openshift-` groups and return `signed` directly.

It has no impact on existing behavior afaik. The `group` param is used only for non-openshift groups and afaict we haven't been building plashets for non-openshift groups until now.

But maybe there's a better approach here - we could simply define `software_lifecycle.phase` for the golang branches and set it to `signing` (though imo this would be abusing the meaning of software lifecycle since there's no lifecycle in the typical sense for golang branches)

pyartcd/pyartcd/pipelines/update_golang.py
adobes1: Great points, thanks! I was originally thinking of raising a follow-up PR to this one to handle mono-branch cases but happy to include it here if you prefer.

> group.yml file will be shared by all builder configs, and so we would need to have an override in the image config.

Makes sense, thank you

The actual mono-branch implementation would depend on which group we decide to build the plashet against (i.e. `openshift-x.y` or `golang`). The build-plashets job needs to have `MAJOR`, `MINOR` set for several steps. 
- With the `openshift-x.y` group, these would be derived directly from the group name. 
- For the `golang` group, we'd need a way to pass the version from the `golang-builder` pipeline to the `build-plashets` pipeline and derive `MAJOR` and `MINOR` from that.

(So in this case, building against the `openshift-x.y` group seems better to me but I'll defer to your opinion :) )

pyartcd/pyartcd/pipelines/update_golang.py
thegreyd: Right, thanks for the analysis. I read the flow and details again and since it works with this now and what you say, it is needed, so i am good with keeping the way it is 👍

pyartcd/pyartcd/pipelines/update_golang.py
thegreyd: We can keep the plashet run group pointing to the old and new golang branches. See _get_doozer_var_args method, if you can use that in the plashet run for the mono branch. i'm also fine if we want to do it in a followup pr

pyartcd/pyartcd/plashets.py
thegreyd: I'm confused by the method interface here.. it is a bit messy, but we don't have to tackle it in this PR.

pyartcd/tests/pipelines/test_update_golang.py
```diff
@@ -1500,5 +1508,259 @@ async def test_rebase_and_build_konflux(self, mock_cmd_assert, mock_konflux_db):
...
+
+        with self.assertRaisesRegex(RuntimeError, "failed with result: FAILURE"):
+            await pipeline._build_golang_plashets("1.22.9", [9])
+
+    @patch("pyartcd.pipelines.update_golang.jenkins")
+    @patch("pyartcd.pipelines.update_golang.KonfluxDb")
+    async def test_dry_run_skips_jenkins(self, mock_konflux_db, mock_jenkins):
+        mock_slack = Mock()
+        mock_slack.say_in_thread = AsyncMock()
+        mock_runtime = Mock(dry_run=True, working_dir=Path("/tmp/working"))
```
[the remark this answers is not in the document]
adobes1: S108 isn't enabled in our ruff config, so this be caught in CI. Skipping for now.

pyartcd/pyartcd/pipelines/update_golang.py
adobes1: Thanks. If you're okay with it then I'd prefer to take a stab at this in a follow-up PR. I've explored the approach you suggested and I think I have an idea of how to make it work, but it would require more changes to the plashet code that I'd rather not rush (and I'd like to get the current change in before the "unsigned rpms" exception expires). I would take a look at the support for monobranch as the next thing

PR #2946, opened 2026-05-26
add doozer_data_gitref to build-microshift

fgallott:
rh-pre-commit.version: 2.3.2

### DIFF
--- a/pyartcd/pyartcd/pipelines/build_microshift.py
+++ b/pyartcd/pyartcd/pipelines/build_microshift.py
@@ -69,10 +69,13 @@ class BuildMicroShiftPipeline:
         skip_prepare_advisory: bool,
         data_path: str,
         slack_client,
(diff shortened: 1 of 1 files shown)


pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -138,11 +141,14 @@ async def run(self):
             await self._run_pipeline()
 
     async def _run_pipeline(self):
-        group_config = await load_group_config(self.group, self.assembly, env=self._doozer_env_vars)
+        data_path = self._doozer_env_vars.get("DOOZER_DATA_PATH", None) or constants.OCP_BUILD_DATA_URL
```
locriandev: Maybe worth setting self.data path as a class attribute and remove the extra initialization at lines 99-102

PR #2954, opened 2026-05-28
ART-19302: feat(images) add base_image_release.force override

lgarciaaco:
## Summary
Add image-level `base_image_release.force` so operators can always run the Konflux base-image snapshot-to-release workflow for a specific image, bypassing normal qualification guards.

## Problem
**Before:** Base-image release only triggered when an image passed all guards: OCP variant, non-test assembly, base_only or golang-builder, and enabled flag resolution.

**After:** Setting `base_image_release.force: true` in image metadata unconditionally triggers the release workflow, regardless of variant, assembly, image type, or enabled settings.

## Implementation Details
- Short-circuit `should_trigger_base_image_release()` when image metadata sets `base_image_release.force: true`
- Extend image config schema with `base_image_release.force` (image-level only, no group override)
- Add unit tests covering force precedence and unchanged behavior when force is unset or false

### Key Changes
- `doozer/doozerlib/image.py`: force check before existing guards
- `ocp-build-data-validator/validator/json_schemas/image_config.base.schema.json`: schema support for `force`
- `doozer/tests/test_image.py`: coverage for bypass and control cases

### DIFF
--- a/doozer/doozerlib/image.py
+++ b/doozer/doozerlib/image.py
@@ -1393,6 +1393,7 @@ class ImageMetadata(Metadata):
         Determines whether this image should trigger the base image release workflow.
 
         The method checks preconditions in the following order:
+        0. Image metadata `base_image_release.force` set to true always enables workflow
         1. Variant must be OCP (not OKD)
         2. Assembly must not be ``test``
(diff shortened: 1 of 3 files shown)


locriandev (commented):
/lgtm

PR #2967, opened 2026-05-29
ART-18202: fix cross-arch lockfile reconciliation ignoring version pins

fgallott:
When reconciling cross-arch RPM version mismatches, version pins were appended to the packages list without removing the unversioned package names. DNF saw both "libsemanage" (install latest) and "libsemanage-3.6-1.el9" (pin), picked the latest per arch, and the pin was effectively ignored. Same issue with update_targets where base.upgrade() would override pins.

Fix: exclude unversioned names of mismatched packages from both the packages list and update_targets before re-resolution. Also raise RuntimeError instead of silently falling back to the mismatched lockfile so builds fail early with a clear message instead of producing Conforma violations later.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cross-platform package reconciliation by excluding mismatched packages from follow-up resolution and update steps.
  * Version-pinned package requirements are now preserved when architectures disagree.
  * Resolution failures, empty results, and unresolved mismatches now report clear errors identifying affected packages instead of silently using incomplete results.
* **Tests**
  * Expanded coverage for cross-platform mismatch handling, update targets, and reconciliation failures.

### DIFF
--- a/doozer/doozerlib/lockfile_prototype/generator.py
+++ b/doozer/doozerlib/lockfile_prototype/generator.py
@@ -1366,6 +1366,17 @@ class RpmLockfilePrototypeGenerator:
             pins.append(format_version_pin(name, min_evr))
         return pins
 
+    @staticmethod
+    def _format_mismatches(mismatches: dict[str, dict[str, str]]) -> str:
+        """
+        Format cross-arch mismatches for error messages.
+        """
+        parts = []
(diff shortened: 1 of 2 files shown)


rayfordj (approved):
/unhold
/lgtm
/approve

I don't know an easy enough way to test this directly with mismatch arch package versions, but testing at least doesn't appear to break any existing behavior. 😅

PR #2969, opened 2026-05-29
ART-18202: support [ arch != X ] || dnf install pattern in lockfile parser

fgallott:
## Summary

  Support `[ arch != X ] || cmd` and `[ arch = X ] && cmd` conditional patterns in the lockfile shell parser, including `$(go env GOARCH)` as an architecture keyword.

  ### Problem

  Some Dockerfiles use list-style arch conditionals instead of `if/then/fi`:

  ```dockerfile
  RUN [ $(go env GOARCH) != "amd64" ] || (\
      yum install -y llvm-toolset-17* cmake3 gcc-c++ libxml2-devel \
      glibc mingw64-gcc && ...)
```
  The parser didn't recognize this pattern, causing arch-specific packages to be treated as unconditional — they'd appear in lockfiles for all architectures.

  ### Changes

  - `shell_parser.py`: Add `ARCH_NEQ_VALUE_RE` regex for `!=` operator. Add `_extract_list_arch_context()` to compute effective arch from `[ test ] || cmd / [ test ] && cmd` patterns. Add `_walk_list_node()` to detect these patterns in bashlex list nodes before falling through to regular walking.
  - `constants.py`: Add `$(go env GOARCH) `to `ARCH_SUBSHELL_KEYWORDS`.

  Supported patterns

  | Pattern | Meaning | Result |
  |---------|---------|--------|
  | `[ $(arch) != X ] \|\| cmd` | double negation → only on X | arch-specific to X |
  | `[ $(arch) = X ] && cmd` | direct match → only on X | arch-specific to X |
  | `[ $(arch) = X ] \|\| cmd` | all except X (needs full arch set) | treated as unconditional |
  | `[ $(arch) != X ] && cmd` | all except X (needs full arch set) | treated as unconditional |

  Works with all existing arch keywords` ($(arch), $(uname -m), $HOSTTYPE, etc.)` plus the new `$(go env GOARCH).`

### DIFF
--- a/doozer/doozerlib/lockfile_prototype/constants.py
+++ b/doozer/doozerlib/lockfile_prototype/constants.py
@@ -22,7 +22,7 @@ SYSTEM_PYTHON = "/usr/bin/python3"
 RPM_LOCKFILE_ENTRY_POINT = "from rpm_lockfile import main; main()"
 
 # Shell subshell expressions that evaluate to the current architecture
-ARCH_SUBSHELL_KEYWORDS = ("$(arch)", "$(uname -m)", "$(uname -p)")
+ARCH_SUBSHELL_KEYWORDS = ("$(arch)", "$(uname -m)", "$(uname -p)", "$(go env GOARCH)")
 
 # Shell variable names that hold the current architecture
 ARCH_VAR_NAMES = ("HOSTTYPE", "ARCH")
--- a/doozer/doozerlib/lockfile_prototype/constants.py
+++ b/doozer/doozerlib/lockfile_prototype/constants.py
@@ -25,7 +25,7 @@ RPM_LOCKFILE_ENTRY_POINT = "from rpm_lockfile import main; main()"
 ARCH_SUBSHELL_KEYWORDS = ("$(arch)", "$(uname -m)", "$(uname -p)", "$(go env GOARCH)")
 
 # Shell variable names that hold the current architecture
-ARCH_VAR_NAMES = ("HOSTTYPE", "ARCH")
+ARCH_VAR_NAMES = ("HOSTTYPE", "ARCH", "GOARCH")
 
 # All arch keywords: subshells + variables in both $VAR and ${VAR} forms
 ARCH_KEYWORDS = ARCH_SUBSHELL_KEYWORDS + tuple(form for name in ARCH_VAR_NAMES for form in (f"${name}", f"${{{name}}}"))
--- a/doozer/doozerlib/lockfile_prototype/shell_parser.py
+++ b/doozer/doozerlib/lockfile_prototype/shell_parser.py
@@ -39,6 +39,10 @@ ARCH_VALUE_RE = re.compile(
     rf"(?:{_ARCH_PATTERN})"
     r"[\"']?\s*==?\s*[\"']?(\w+)[\"']?"
 )
+ARCH_NEQ_VALUE_RE = re.compile(
+    rf"(?:{_ARCH_PATTERN})"
+    r"""[\"']?\s*!=\s*[\"']?(\w+)[\"']?"""
+)
 
 
 _MAX_EXPANSION_DEPTH = 10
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: doozer/doozerlib/lockfile_prototype/shell_parser.py [hunk 1]
```diff
@@ -148,6 +152,75 @@ def _extract_condition_arch(node) -> list[str]:
...
+        return None
+    text = " ".join(words)
+    if not _has_arch_test(text):
+        return None
+
+    neq_arches = ARCH_NEQ_VALUE_RE.findall(text)
+    eq_arches = ARCH_VALUE_RE.findall(text)
+
+    if neq_arches and operator == "||":
+        return neq_arches
```
lgarciaaco: This returns the literal string from the test (`amd64`, `arm64`, etc.). That works fine for `$(arch)` / `uname -m` since those already use RPM naming, but `$(go env GOARCH)` comparisons tend to use Go names.

Downstream we pass this straight into `ArchSpecificPackage(arches={"only": arch})` against assembly arches like `x86_64`. Wondering if a small normalization step here when the test text contains `go env GOARCH` would help — e.g. `amd64` → `x86_64`, `arm64` → `aarch64`. Otherwise the GOARCH tests pass but lockfile gen might not pick those packages up on the right arch.

doozer/tests/lockfile_prototype/test_shell_parser.py
```diff
@@ -402,6 +403,105 @@ def test_arch_value_regex_no_match():
...
+
+    def test_go_env_goarch_neq_or(self):
+        """
+        ``[ $(go env GOARCH) != "amd64" ] || yum install -y pkg``
+        """
+        pkgs, arch_pkgs = extract_packages_from_run_commands(
+            ['[ $(go env GOARCH) != "amd64" ] || yum install -y special-pkg']
+        )
+        self.assertEqual(pkgs, [])
+        self.assertEqual(arch_pkgs, {"amd64": ["special-pkg"]})
```
lgarciaaco: If you add the Go→RPM mapping, this might want to expect `x86_64` instead of `amd64` — would line up with the rest of the lockfile tests and what the generator resolves.

doozer/tests/lockfile_prototype/test_shell_parser.py
```diff
@@ -402,6 +403,105 @@ def test_arch_value_regex_no_match():
...
+    def test_go_env_goarch_neq_or_subshell(self):
+        """
+        ``[ $(go env GOARCH) != "amd64" ] || (yum install -y pkg1 pkg2 && other)``
+        — subshell grouping after ``||`` should still extract arch packages.
+        """
+        pkgs, arch_pkgs = extract_packages_from_run_commands(
+            ['[ $(go env GOARCH) != "amd64" ] || (yum install -y llvm-toolset cmake3 gcc-c++ && tar zfx cross.tar.gz)']
+        )
+        self.assertEqual(pkgs, [])
+        self.assertEqual(arch_pkgs, {"amd64": ["cmake3", "gcc-c++", "llvm-toolset"]})
```
lgarciaaco: Same thought as the test above — `x86_64` here if we normalize Go arch names.

doozer/tests/lockfile_prototype/test_shell_parser.py
```diff
@@ -402,6 +403,105 @@ def test_arch_value_regex_no_match():
...
+        ``[ $(arch) != x86_64 ] && dnf install -y pkg`` means "all except x86_64"
+        — treated as unconditional.
+        """
+        pkgs, arch_pkgs = extract_packages_from_run_commands(['[ $(arch) != x86_64 ] && dnf install -y pkg'])
+        self.assertIn("pkg", pkgs)
+        self.assertEqual(arch_pkgs, {})
+
+    def test_arch_context_only_applies_to_next_command(self):
+        """
+        In ``[ test ] || cmd1 && cmd2``, only cmd1 gets the arch context.
```
lgarciaaco, line 458: Might be worth a sentence in the docstring that this relies on bashlex flattening `[ test ] || cmd1 && cmd2` so only `cmd1` is `parts[i+2]` and `cmd2` stays in `remaining`. Matches bash left-associativity (`(test || cmd1) && cmd2`), but not obvious from the test name alone.

doozer/doozerlib/lockfile_prototype/constants.py
```diff
@@ -22,7 +22,7 @@
 RPM_LOCKFILE_ENTRY_POINT = "from rpm_lockfile import main; main()"
 
 # Shell subshell expressions that evaluate to the current architecture
-ARCH_SUBSHELL_KEYWORDS = ("$(arch)", "$(uname -m)", "$(uname -p)")
+ARCH_SUBSHELL_KEYWORDS = ("$(arch)", "$(uname -m)", "$(uname -p)", "$(go env GOARCH)")
```
lgarciaaco, line 25: Makes sense to add GOARCH here. Just noting this path probably needs the RPM arch normalization in `_extract_list_arch_context` since Go and RPM use different names for the same platform.

lgarciaaco (commented):
Looks good overall — the supported `||` / `&&` pairings match real Dockerfile patterns, and falling back to unconditional for the "all except X" combos seems like the right call. Tests are solid and CI is green.

One thing I noticed: when the comparison comes from `$(go env GOARCH)`, the parser keys packages under Go names like `amd64`, but lockfile resolution elsewhere uses RPM names (`x86_64`, `aarch64`, …). Might be worth a small Go→RPM mapping when GOARCH is the source, or a note in the code if that's out of scope for this PR.

Left a few inline thoughts — nothing blocking from my side once that's clarified.

PR #2980, opened 2026-06-01
ART-18937: feat(konflux): expand KonfluxBuildOutcome with stage-specific failures

lgarciaaco:
## Summary
Expand `KonfluxBuildOutcome` with granular failure types (`build_error`, `its_error`, `release_error`) so Konflux image build failures in BigQuery distinguish build PLR, EC/ITS, and base-image release stages.

## Problem
**Before:** Most Konflux image failures were stored as generic `failure`, and failure type was inferred from ephemeral Jenkins record flags (`ec_failed`, `base_image_release_failed`). New rows also wrote deprecated `ec_status`.

**After:** Failed build PLRs persist as `build_error`, EC failures as `its_error`, and base-image release failures as `release_error`. Pipeline URLs and release fields are written per stage; `ec_status` is no longer populated on new image build records. `ocp4-konflux` fail counters prefer the persisted `outcome` with fallback to legacy flags.

## Implementation Details
- Added `BUILD_ERROR`, `ITS_ERROR`, `RELEASE_ERROR` plus `is_success()` / `is_failure()` helpers on `KonfluxBuildOutcome`
- `extract_from_pipelinerun_succeeded_condition` maps failed build PLRs to `build_error`
- Updated `konflux_image_builder`, FBC/OLM bundler DB paths, and `ocp4_konflux` counter categorization
- Unit tests for enum serialization, image builder flows, EC failure outcome, and fail-counter routing

https://redhat.atlassian.net/browse/ART-18937

### DIFF
--- a/artcommon/artcommonlib/konflux/konflux_build_record.py
+++ b/artcommon/artcommonlib/konflux/konflux_build_record.py
@@ -20,6 +20,18 @@ class KonfluxEnum(Enum):
 
 
 class KonfluxBuildOutcome(KonfluxEnum):
+    """Konflux build lifecycle outcome stored in BigQuery.
+
+    Taxonomy:
+    - SUCCESS: releasable build completed all required stages.
+    - PENDING: in-progress or awaiting further processing (intermediate row).
+    - BUILD_ERROR: build PipelineRun failed (canonical PLR failure; prefer over FAILURE).
+    - ITS_ERROR: Enterprise Contract / IntegrationTestScenario verification failed after a successful build.
+    - RELEASE_ERROR: base-image snapshot/release failed after a successful build.
+    - FAILURE: legacy generic failure; retained for old rows and out-of-scope writers (FBC/OLM may still emit it).
+    - TIMEOUT / CANCELLED: pipelinerun terminal states from Tekton Succeeded condition.
+    """
+
     FAILURE = 'failure'
     SUCCESS = 'success'
     PENDING = 'pending'
@@ -28,6 +40,19 @@ class KonfluxBuildOutcome(KonfluxEnum):
     BUILD_ERROR = 'build_error'
     ITS_ERROR = 'its_error'
     RELEASE_ERROR = 'release_error'
+
+    @classmethod
+    def db_filter_values(cls) -> tuple[str, ...]:
+        """Outcome values included in KonfluxDb default and cache queries."""
+        return (
+            cls.SUCCESS.value,
+            cls.FAILURE.value,
+            cls.BUILD_ERROR.value,
+            cls.ITS_ERROR.value,
+            cls.RELEASE_ERROR.value,
+            cls.TIMEOUT.value,
+            cls.CANCELLED.value,
+        )
 
     def is_success(self) -> bool:
         return self is KonfluxBuildOutcome.SUCCESS
--- a/artcommon/artcommonlib/konflux/konflux_build_record.py
+++ b/artcommon/artcommonlib/konflux/konflux_build_record.py
@@ -25,6 +25,22 @@ class KonfluxBuildOutcome(KonfluxEnum):
     PENDING = 'pending'
     TIMEOUT = 'timeout'
     CANCELLED = 'cancelled'
+    BUILD_ERROR = 'build_error'
+    ITS_ERROR = 'its_error'
+    RELEASE_ERROR = 'release_error'
+
+    def is_success(self) -> bool:
+        return self is KonfluxBuildOutcome.SUCCESS
+
+    def is_failure(self) -> bool:
+        return self in (
+            KonfluxBuildOutcome.FAILURE,
+            KonfluxBuildOutcome.BUILD_ERROR,
+            KonfluxBuildOutcome.ITS_ERROR,
+            KonfluxBuildOutcome.RELEASE_ERROR,
+            KonfluxBuildOutcome.TIMEOUT,
+            KonfluxBuildOutcome.CANCELLED,
+        )
 
     @classmethod
     def extract_from_pipelinerun_succeeded_condition(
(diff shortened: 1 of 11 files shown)


artcommon/artcommonlib/konflux/konflux_build_record.py
```diff
@@ -34,13 +75,12 @@ def extract_from_pipelinerun_succeeded_condition(
...
-                    return cls.CANCELLED
-                if reason == 'Timeout':
-                    return cls.TIMEOUT
-                return cls.FAILURE
+            reason = succeeded_condition.reason
+            if reason == 'Cancelled':
+                return cls.CANCELLED
+            if reason == 'Timeout':
+                return cls.TIMEOUT
+            return cls.BUILD_ERROR
```
locriandev, line 83: Is this correct? not fully aware of how the KubeCondition looks like, but could a failure here be due to ITS or authz failures?

doozer/doozerlib/backend/konflux_fbc.py
```diff
@@ -2130,17 +2130,17 @@ async def _update_konflux_db(
                             'image_tag': image_pullspec.split(':')[-1],
                         }
                     )
-                case KonfluxBuildOutcome.FAILURE:
```
locriandev, line 2133: Removing this case would catch all non-success fields, is this really what we want? or, is it a mere translation of FAILURE into BUILD_ERROR?

artcommon/artcommonlib/konflux/konflux_build_record.py
lgarciaaco: Good question — this helper only maps the Tekton PipelineRun `Succeeded` condition, not the final row we write to BigQuery.

For image builds the flow is layered. `extract_from_pipelinerun_succeeded_condition` runs right after the *build* PLR finishes. EC/ITS only kicks in when that returns SUCCESS (`konflux_image_builder.build()` around the `should_run_ec` block). If EC fails we overwrite to `ITS_ERROR` there. Same idea for base-image release → `RELEASE_ERROR`.

So a failed build PLR won't land as ITS_ERROR from this line. Authz/registry/git failures during build tasks show up as the build PLR failing, and `BUILD_ERROR` is what we want for that stage.

Attestation/signature issues after a successful PLR are handled separately and also get remapped to `BUILD_ERROR`.

One nuance: we also call this on the EC PLR inside `verify_enterprise_contract`, but that return value never gets persisted — callers use `ec_failed` and set `ITS_ERROR`. Might be worth a short docstring note that this is PLR-scoped; happy to add that in a follow-up if you think it helps.

doozer/doozerlib/backend/konflux_fbc.py
lgarciaaco: Intentional, and a bit more than FAILURE → BUILD_ERROR.

The old `case KonfluxBuildOutcome.FAILURE` block only stamped `start_time`/`end_time` for generic PLR failures. Once `extract_from_pipelinerun_succeeded_condition` started returning `BUILD_ERROR` instead of `FAILURE`, that branch would have stopped matching and we'd lose timestamps on failed builds.

Moving the timestamp logic outside the `match` means we populate PLR timing whenever `status` is present, regardless of outcome — `BUILD_ERROR`, `ITS_ERROR`, `TIMEOUT`, etc. We're not reclassifying failures; we're decoupling timing metadata from outcome type.

Same pattern in `konflux_olm_bundler.py`. For `ITS_ERROR` rows the timestamps still come from the build PLR (which succeeded), and EC timing lives on `ec_pipeline_url`. That matches what we did before when EC failures were stored as `FAILURE`.

lgarciaaco:
@locriandev thanks for the review — replied inline on both threads.

Short version: no behavior change requested from your comments. `extract_*` is PLR-scoped (callers assign ITS/release outcomes), and the FBC/OLM timestamp refactor is deliberate so `BUILD_ERROR` rows still get PLR start/end times. Let me know if the inline replies don't land or you want a docstring tweak on `extract_from_pipelinerun_succeeded_condition`.

lgarciaaco:
[test build](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-konflux/36382/)

lgarciaaco:
## BigQuery validation (openshift-art.events.builds)

I queried production BigQuery for the new granular `outcome` values after the 4.22 test-assembly runs started writing them (first rows ~2026-06-01 12:28 UTC).

### What shows up in the DB

| Outcome | Count (30d, konflux) | Where |
|---------|----------------------|--------|
| `build_error` | 16 | `openshift-4.22`, `assembly=test` |
| `its_error` | 9 | same |
| `release_error` | 0 | — |

Legacy `failure` is still the bulk of failed rows (~17k in 30d). Granular outcomes only appear on **test** assembly so far; nothing on `stream` yet (expected until this ships on main and stream Jenkins picks it up).

### Stage fields line up with the taxonomy

- **`build_error`** (16/16): `build_pipeline_url` set, `ec_pipeline_url` empty, `ec_status` = `n/a`
- **`its_error`** (9/9): both `build_pipeline_url` and `ec_pipeline_url` set, `ec_status` = `n/a`

Example on **ironic** / 4.22 / test: `build_error` at 2026-06-01 12:28 (no EC PLR), then `its_error` at 13:05 (EC PLR present) on a later attempt. Same component, different stage, different outcome.

Latest `build_error` sample: `ironic-container-v4.22.0-202606021258.p2.gf934d2c.assembly.test.el9` (2026-06-02 13:02 UTC).

### Not covered in live data yet

`release_error` has no rows in the last 90 days. Unit coverage for that path (`test_trigger_base_image_release_failure_flow`) is still the only proof until a base-image release fails under this code.

### Takeaway

`build_error` and `its_error` look correct in BigQuery: stage-specific outcomes, pipeline URLs on the right fields, deprecated `ec_status` not populated on new rows. I'd treat `release_error` as validated in tests only until we get a real release failure in the DB.

locriandev (approved):
/lgtm

PR #2990, opened 2026-06-03
Fix embargo detection bypass when pull_url is None

joepvd:
## Summary

- Fix private fix detection logic that was bypassed when `pull_url` is `None`
- Add `is_fork_build` property to `SourceResolution` to properly detect fork builds
- Add comprehensive unit tests including E2E tests with realistic commit scenarios

## Problem

The condition `source.url == source.pull_url` on line 283 of `rebaser.py` evaluated to `False` when `pull_url` is `None` (the common case when `url_pull` is not specified in ocp-build-data). This caused the `is_commit_in_public_upstream` check to be completely skipped, resulting in embargoed builds being incorrectly marked as public.

## Fix

Introduced an `is_fork_build` property that returns `True` only when `pull_url` is explicitly set to a different URL (fork build). The rebaser now uses `not source.is_fork_build`:

| Scenario | `is_fork_build` | Embargo check |
|----------|-----------------|---------------|
| `pull_url=None` (normal) | `False` | ✅ Runs |
| `pull_url==url` (explicit same) | `False` | ✅ Runs |
| `pull_url!=url` (fork build) | `True` | ❌ Skipped (Dockerfile fallback) |

## Test plan

- [x] Added 3 tests for `is_fork_build` property
- [x] Added 7 unit tests for private fix detection scenarios
- [x] Added 5 E2E tests with realistic commit hashes (f1ffd09 private, abc4567 public)
- [x] All 2,595 tests pass (`make test`)

### DIFF
--- a/doozer/doozerlib/backend/rebaser.py
+++ b/doozer/doozerlib/backend/rebaser.py
@@ -280,7 +280,7 @@ class KonfluxRebaser:
 
         # Determine if this image contains private fixes
         if private_fix is None:
-            if source and source_dir and source.url == source.pull_url:
+            if source and source_dir and not source.is_fork_build:
                 # If the private org branch commit doesn't exist in the public org,
                 # this image contains private fixes
                 is_commit_in_public_upstream = await util.is_commit_in_public_upstream_async(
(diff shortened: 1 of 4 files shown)


locriandev (commented):
/lgtm

PR #2995, opened 2026-06-03
ART-19484: fix(images): skip base image release for OKD before force check

lgarciaaco:
## Summary

OKD Konflux builds no longer trigger RH base-image snapshot→release when image metadata has `base_image_release.force: true`. The OKD variant guard runs first in `should_trigger_base_image_release()`.

## Problem

**Before:** `base_image_release.force` was evaluated before `BuildVariant.OKD`, so OKD builds of openshift-5.0 images (e.g. `ose-installer` with `force: true` from ocp-build-data#10855) ran snapshot→release and persisted `release_error` in BigQuery.

**After:** OKD always returns `False` from `should_trigger_base_image_release()`; `force` applies only on OCP. Successful OKD PLRs persist as `success`.

## Implementation

- Move `if self.runtime.variant is BuildVariant.OKD: return False` to the top of `should_trigger_base_image_release()`
- Update docstring precondition order
- Test: OKD + `force: true` expects `False`

https://redhat.atlassian.net/browse/ART-19484

### DIFF
--- a/doozer/doozerlib/image.py
+++ b/doozer/doozerlib/image.py
@@ -1391,8 +1391,8 @@ class ImageMetadata(Metadata):
         Determines whether this image should trigger the base image release workflow.
 
         The method checks preconditions in the following order:
-        0. Image metadata `base_image_release.force` set to true always enables workflow
-        1. Variant must be OCP (not OKD)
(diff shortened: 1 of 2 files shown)


locriandev (approved):
/lgtm

PR #3001, opened 2026-06-03
ART-19485: preserve release_pipeline URL in BigQuery for failed base image releases

locriandev:
## Summary

Fixes an issue where the `release_pipeline` field in BigQuery was empty for failed base image releases, making it difficult to debug release failures.

## Problem

When a base image release failed (outcome = `release_error`), the `release_pipeline` URL was not being stored in BigQuery because:
1. `BaseImageHandler.snapshot_release()` returned `None` when `_wait_for_release_completion()` returned `False`
2. The caller in `KonfluxImageBuilder` then stored an empty string for `release_pipeline`

This meant we couldn't query BigQuery to find the Konflux release URL for debugging failed releases.

## Solution

**Modified `BaseImageHandler.snapshot_release()`:**
- Now returns a `BaseImageReleaseResult` even when the release fails
- Sets `release_pipeline` to the URL (for debugging)
- Sets `released_pullspec` to empty string (to indicate failure)

**Modified `KonfluxImageBuilder`:**
- Checks `released_pullspec` presence (not just `release_result` truthiness) to determine success
- Preserves `release_pipeline` URL from the result even when the release fails

## Testing
https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-konflux/36563/

### DIFF
--- a/doozer/doozerlib/backend/konflux_image_builder.py
+++ b/doozer/doozerlib/backend/konflux_image_builder.py
@@ -373,7 +373,9 @@ class KonfluxImageBuilder:
                         release_result = await self._trigger_base_image_release(
                             metadata, nvr, definitive_image_pullspec, build_repo
                         )
-                        if release_result:
+                        # Success is indicated by having a released_pullspec (empty on failure)
+                        release_succeeded = release_result and release_result.released_pullspec
+                        if release_succeeded:
                             logger.info("Base image release succeeded for %s, persisting build record", nvr)
                             record["release_pipeline"] = release_result.release_pipeline
                         else:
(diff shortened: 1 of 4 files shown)


lgarciaaco:
Separate idea for a follow-up (not asking for it in this PR): when `_trigger_base_image_release` returns `None` (snapshot failed or Release CR never created), we still persist an empty `release_pipeline` on `RELEASE_ERROR`. That was the original triage gap for rows where no Konflux Release existed at all.

Storing something like `N/A` in that bucket would make BigQuery queries easier to read ("no release to link" vs "bug dropped the URL"). Happy to open a small follow-up for [ART-19485](https://redhat.atlassian.net/browse/ART-19485) if you want that layered on after this lands.

doozer/doozerlib/backend/konflux_image_builder.py
```diff
@@ -373,7 +373,9 @@ async def build(self, metadata: ImageMetadata, git_auth_secret: Optional[str] =
                         release_result = await self._trigger_base_image_release(
                             metadata, nvr, definitive_image_pullspec, build_repo
                         )
-                        if release_result:
+                        # Success is indicated by having a released_pullspec (empty on failure)
+                        release_succeeded = release_result and release_result.released_pullspec
```
lgarciaaco, line 377: Small related nit in `_trigger_base_image_release` (~L1264): when `snapshot_release()` returns a partial result (URL set, empty `released_pullspec`), that helper still logs "Successfully triggered..." at INFO. The builder side here looks right with `release_succeeded`, but log triage might misread the failure path.

Might be worth an error-level log there when `result` is present but `not result.released_pullspec`.

locriandev: done with 9d76443162470e76298ac774870f93a1eb0ce71e

PR #3005, opened 2026-06-04
ART-18202: pass reinstall packages as upgrade targets to avoid PackagesNotAvailableError

fgallott:
### Problem

  After #2999, `base.reinstall(pkg)` crashes with `PackagesNotAvailableError` when a base image package's installed version isn't in the configured EUS repos (e.g. `sed`, `libnghttp2`).
  This happens because the base image was built from a different repo set than what's configured for lockfile resolution.

  ### Root cause

  rpm-lockfile-prototype's `base.reinstall()` handler swallows `PackagesNotAvailableError` **only** when the package is also in `upgradePackages`:

  ```python
  except dnf.exceptions.PackagesNotAvailableError:
      if pkg not in upgrade_packages:
          raise  # <-- crashes here
```
  We were passing base image packages as reinstallPackages but not as upgradePackages.

   ### Fix

  Pass all reinstallPackages as upgradePackages too. When the installed version isn't in repos, the reinstall error is swallowed (because the package IS in upgrade targets), and the
  upgrade provides a replacement version from the repos. This is the intended design of the rpm-lockfile-prototype handler.

  Also makes remaining_reinstall mutable in the retry loop so unavailable packages are filtered on retry (same as remaining_packages and remaining_update_targets).

### DIFF
--- a/doozer/doozerlib/lockfile_prototype/generator.py
+++ b/doozer/doozerlib/lockfile_prototype/generator.py
@@ -660,14 +660,22 @@ class RpmLockfilePrototypeGenerator:
         # use the no-auth registry proxy instead.
         resolver_pullspec = ContainerImageHelper._proxy_pullspec(image_pullspec) if image_pullspec else None
 
+        # When reinstall_packages is set, also pass them as upgrade targets.
+        # base.reinstall() raises PackagesNotAvailableError when the installed
+        # version isn't in the configured repos — but rpm-lockfile-prototype
+        # swallows that error when the package is also in upgradePackages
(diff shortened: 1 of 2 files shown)


lgarciaaco:
Makes sense to me. Passing reinstall targets through upgradePackages matches how rpm-lockfile-prototype handles PackagesNotAvailableError when the installed NVR is not in the configured repos, and the retry tests cover the filtering path.

Optional follow-up (not blocking): the "no packages remaining" skip might want to check remaining_reinstall too if install/arch lists are empty but base reinstall targets are still left.

/lgtm

PR #3007, opened 2026-06-04
fix(doozer): pass registry_config to oc image info during upstream RHEL version detection

thegreyd:
## Summary
- Pass `registry_config` through to `oc image info` in `extract_builder_info_from_pullspec()` so that base image inspection works on the buildvm where `registry.ci.openshift.org` requires explicit auth
- Without this, images whose upstream Dockerfile uses ambiguous base tags (e.g. `:cli`, `:base`) fail RHEL version detection on the buildvm — the `oc image info` fallback silently fails and the code picks up a `rhel-8` compat builder tag instead of the correct `rhel-9` base, causing a spurious mismatch error during rebase
- The `@lru_cache` on `extract_builder_info_from_pullspec` is removed because adding `registry_config` as a parameter would create redundant cache entries for the same pullspec (`registry_config` doesn't affect the result, only whether the call succeeds). The underlying `oc_image_info__cached__lru` already provides equivalent caching with `registry_config` correctly excluded from its cache key

## Test plan
- [ ] Verify `kube-compare-artifacts` rebase succeeds on buildvm for openshift-5.0
- [ ] Run `uv run pytest doozer/tests/test_image.py` — all 81 tests pass

### DIFF
--- a/doozer/doozerlib/backend/rebaser.py
+++ b/doozer/doozerlib/backend/rebaser.py
@@ -2302,7 +2302,10 @@ class KonfluxRebaser:
             self._logger.debug('Retrieving image info for image %s', original_parent)
 
             # Use the cached function to extract builder info
-            el_version, golang_version = extract_builder_info_from_pullspec(original_parent)
+            el_version, golang_version = extract_builder_info_from_pullspec(
+                original_parent,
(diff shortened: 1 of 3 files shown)


joepvd (approved):
/lgtm

PR #3015, opened 2026-06-05
fix(doozer): skip shell cmds in Konflux injected lines for scratch stream images

ashwindasr:
## Summary

- When an image uses `from: stream: scratch` (e.g. `agentic-skills`), the rebased Dockerfile has `FROM scratch` which has no shell
- Doozer was injecting `RUN go clean -cache || true` and `USER 0` into the Konflux-specific block, causing the build to fail
- Auto-detect `from.stream == 'scratch'` and treat it as `no_shell=True` so these shell-dependent lines are skipped

Related ocp-build-data PR: https://github.com/openshift-eng/ocp-build-data/pull/11034

Test build: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Focp4-konflux/36769/console

### DIFF
--- a/doozer/doozerlib/backend/rebaser.py
+++ b/doozer/doozerlib/backend/rebaser.py
@@ -1273,7 +1273,9 @@ class KonfluxRebaser:
         df_lines = filtered_content
 
         # ART-8476 assert rhel version equivalence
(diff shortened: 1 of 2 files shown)


joepvd (approved):
/lgtm

PR #3019, opened 2026-06-05
feat(pyartcd): add GA readiness check in prepare-release-konflux

ashwindasr:
## Summary

- Adds `check_bug_config_for_ga()` to `PrepareReleaseKonfluxPipeline`, called early in `_run_pipeline()` as a fail-fast check for standard (GA/z-stream) assemblies
- Validates that `group.yml`'s `software_lifecycle.phase` is `'release'` and that `bug.yml`'s `target_release` includes the `.z` entry (e.g. `4.22.z`)
- Both conditions are checked independently and all errors are reported at once, so they can be fixed in a single ocp-build-data PR
- This check would have caught the missed configuration in ocp-build-data PR #11046 (4.22 GA transition)

## Test plan

- [ ] New `TestCheckBugConfigForGa` test class with 7 unit tests covering all cases (pass, phase wrong, z-stream missing, both wrong, non-standard assembly skipped, bug.yml missing)
- [ ] All 7 tests pass

### DIFF
--- a/pyartcd/pyartcd/pipelines/prepare_release_konflux.py
+++ b/pyartcd/pyartcd/pipelines/prepare_release_konflux.py
@@ -18,6 +18,7 @@ from urllib.parse import urlparse
 import aiohttp
 import asyncstdlib as a
 import click
+import ruamel.yaml
 import semver
 from artcommonlib import exectools
 from artcommonlib.assembly import (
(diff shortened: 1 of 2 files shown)


joepvd (approved):
/lgtm

PR #3021, opened 2026-06-05
WIP fix(doozer,pyartcd): capture parent-failure message in record and exc…

locriandev:
…lude from counters

When a parent image fails to build, its child images raise an IOError before the build loop starts. Previously this left record['message'] as 'Unknown failure', and update_build_fail_counters() couldn't distinguish these children from real build failures. This caused erroneous counter increments for images that were never attempted.

Changes:
- doozer/konflux_image_builder.py: add except clause before finally to capture the actual exception message in record['message'] when it would otherwise remain 'Unknown failure'
- pyartcd/ocp4_konflux.py: filter out entries whose message contains 'parent images failed to build' before categorizing failures and incrementing Redis counters
- Tests for both fixes in their respective test files

### DIFF
--- a/doozer/doozerlib/backend/konflux_image_builder.py
+++ b/doozer/doozerlib/backend/konflux_image_builder.py
@@ -434,6 +434,16 @@ class KonfluxImageBuilder:
                 record["message"] = str(error)
                 raise error
 
+        except Exception as exc:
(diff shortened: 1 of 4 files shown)


locriandev:
superseded by https://github.com/openshift-eng/art-tools/pull/3022

PR #3043, opened 2026-06-09
Update golang builder CVE component name to openshift4/openshift-golang-builder

ashwindasr:
## Summary
- Update `GOLANG_BUILDER_CVE_COMPONENT` from `openshift-golang-builder-container` to `openshift4/openshift-golang-builder` to match the new JIRA pscomponent name for golang builder security trackers.
- Introduce `GOLANG_BUILDER_BREW_COMPONENT` (`openshift-golang-builder-container`) to preserve the Brew/Konflux package name which remains unchanged.
- Fix filtering logic in `attach_cve_flaws_cli` and `find_bugs_sweep_cli` to correctly handle the new `openshift4/` prefix format.
- Add `is_golang_builder_component()` helper that accepts both old and new pscomponent names for backward compatibility during migration.
- Fix `is_first_fix_for_tracker` / `is_first_fix_any` to normalize tracker `whiteboard_component` before comparing against Hydra unfixed components, preventing mismatches when the tracker uses a delivery-repo-format pscomponent.
- Add fallback in `get_cve_unfixed_components` to preserve raw delivery-repo names when translation fails.
- Extend `is_valid()` filter in `find_bugs_golang_cli` to also skip delivery-repo-format image trackers (`openshift4/...`), not just `-container` suffixed ones, while preserving the golang builder exemption.

## Test plan
- [x] All existing unit tests pass (71/71 in affected test files)
- [x] New tests for `is_golang_builder_component()` helper (4 cases)
- [x] New tests for `is_first_fix` normalization and fallback (2 cases)
- [x] Lint and format checks pass (`ruff check` clean)
- [ ] Verify golang bug sweep works correctly with new pscomponent in a dry-run against a stream assembly
- [ ] Verify `update-golang` pipeline correctly sweeps builder bugs with the new component name

### DIFF
--- a/elliott/elliottlib/bzutil.py
+++ b/elliott/elliottlib/bzutil.py
@@ -35,6 +35,7 @@ from elliottlib.util import (
     chunk,
     get_component_by_delivery_repo,
     isolate_timestamp_in_release,
+    normalize_component_by_ocp_delivery_repo,
 )
 
 logger = logutil.get_logger(__name__)
@@ -1730,6 +1731,7 @@ def get_cve_unfixed_components(runtime, cve_alias: str) -> Dict:
                     logger.warning(
                         f"Could not find component name for {pkg_name}! is it an art component? is the delivery repo defined?"
                     )
+                    unfixed_components.append(pkg_name)
                 else:
                     unfixed_components.append(comp_name)
             else:
@@ -1751,7 +1753,8 @@ def is_first_fix_for_tracker(runtime, flaw_bug: BugzillaBug, tracker_bug: JIRABu
         )
     alias = flaw_bug.alias[0]
     unfixed_components = get_cve_unfixed_components(runtime, alias)
-    first_fix = tracker_bug.whiteboard_component in unfixed_components
+    tracker_component = normalize_component_by_ocp_delivery_repo(runtime, tracker_bug.whiteboard_component)
+    first_fix = tracker_component in unfixed_components
     logger.info(
         f"Flaw bug {flaw_bug.id} is {'first-fix' if first_fix else 'second-fix'} for tracker bug {tracker_bug.id}"
     )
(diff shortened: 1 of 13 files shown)


elliott/elliottlib/constants.py
```diff
@@ -22,7 +22,14 @@
 }
 
 # Golang builder needs special treatment when associating security tracking bugs with builds:
-GOLANG_BUILDER_CVE_COMPONENT = 'openshift-golang-builder-container'
+GOLANG_BUILDER_CVE_COMPONENT = 'openshift4/openshift-golang-builder'
+GOLANG_BUILDER_BREW_COMPONENT = 'openshift-golang-builder-container'
```
thegreyd: i would prefer if we don't use the word brew here
since we compute component names for konflux the same way

Maybe just golang_builder_component ?

And `openshift4/openshift-golang-builder` can be the golang_builder_ocp4_delivery_repo? 
there will be an ocp5 one also, have a func wrap this `is_golang_builder_delivery_repo` ?

elliott/elliottlib/bzutil.py
```diff
@@ -1751,7 +1753,8 @@ def is_first_fix_for_tracker(runtime, flaw_bug: BugzillaBug, tracker_bug: JIRABu
         )
     alias = flaw_bug.alias[0]
     unfixed_components = get_cve_unfixed_components(runtime, alias)
-    first_fix = tracker_bug.whiteboard_component in unfixed_components
+    tracker_component = normalize_component_by_ocp_delivery_repo(runtime, tracker_bug.whiteboard_component)
```
thegreyd, line 1756: nice! thanks for fixing these

elliott/elliottlib/constants.py
ashwindasr: thanks - that makes sense. added in https://github.com/openshift-eng/art-tools/pull/3043/commits/80b30a91e05cadfaebb78baf857da2adfff2241d

ashwindasr:
/hold

Might need to confirm with with prodsec

elliott/elliottlib/constants.py
```diff
@@ -22,7 +22,18 @@
 }
 
 # Golang builder needs special treatment when associating security tracking bugs with builds:
-GOLANG_BUILDER_CVE_COMPONENT = 'openshift-golang-builder-container'
+GOLANG_BUILDER_COMPONENT = 'openshift-golang-builder-container'
+GOLANG_BUILDER_OCP4_DELIVERY_REPO = 'openshift4/openshift-golang-builder'
```
thegreyd, line 26: I don't know if we had discussion around handling openshift5/ here ? If not, then I'd recommend that

PR #3048, opened 2026-06-10
Filter Konflux golang-builder search by hermetic flag

thegreyd:
## Summary
- Adds `_get_effective_network_mode()` to `UpdateGolangPipeline` that reads the network mode from ocp-build-data config (CLI override > image config > group config > default "open"), matching doozer's precedence
- Modifies `get_existing_builders_konflux()` to filter by `hermetic=True/False` based on the effective network mode, preventing reuse of non-hermetic builds when config requires hermetic (or vice versa)
- Fixes boolean handling in `search_builds_by_fields` to use `Column(col_name, Boolean)` instead of `String` for bool values in the `where` dict

## Test plan
- [x] All 78 existing + new unit tests pass (`uv run pytest pyartcd/tests/pipelines/test_update_golang.py`)
- [x] Lint passes (`make lint`)
- [ ] Verify with a dry-run of `update_golang` against an OCP version with hermetic golang-builder config

### DIFF
--- a/artcommon/artcommonlib/metadata.py
+++ b/artcommon/artcommonlib/metadata.py
@@ -21,6 +21,27 @@ CONFIG_MODES = [
 
 CONFIG_MODE_DEFAULT = CONFIG_MODES[0]
 
+VALID_NETWORK_MODES = ["hermetic", "internal-only", "open"]
+
+
+def resolve_network_mode(
+    cli_override: str | None = None,
+    image_config_network_mode: str | None = None,
+    group_config_network_mode: str | None = None,
+) -> str:
+    """Resolve the effective Konflux network mode from multiple config sources.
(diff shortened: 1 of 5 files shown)


artcommon/artcommonlib/metadata.py
```diff
@@ -33,6 +33,10 @@ def resolve_network_mode(
 
     Precedence: cli_override > image_config > group_config > "open"
     """
+    cli_override = cli_override if isinstance(cli_override, str) else None
```
fgallott: mmm.... 

The function signature says `str | None`. Passing a MagicMock is a test bug, not a function bug. 
The right fix would be to set `runtime.network_mode_override = None` in those 4 tests IMHO

PR #3051, opened 2026-06-11
fix: respect skip_nvr_check stream config in scan-sources-konflux

joepvd:
## Summary
- The `skip_nvr_check` field was defined in the streams schema but never read by `scan_sources_konflux`, causing spurious `OSError` when NVR resolution fails for streams (e.g. `rhel-coreos-base`) that don't have a corresponding Brew build
- Now checks `skip_nvr_check` on the stream config before raising, and skips the builder with an info log when set

## Test plan
- [x] Existing scan_sources tests pass (45 passed)
- [ ] Verify with a stream that has `skip_nvr_check: true` (e.g. `rhel-coreos-base`) that the error is no longer raised


locriandev (approved):
/lgtm

PR #3062, opened 2026-06-15
Hoist _get_microshift_rpm_commit() out of _rebase_and_build_bootc

fgallott:
- Move _get_microshift_rpm_commit() call to _run_pipeline so it runs once before parallel variant builds, avoiding concurrent elliott calls clashing on the shared working directory
- Pass upstream_commit as a parameter to _rebase_and_build_bootc
- Update tests to pass upstream_commit directly instead of mocking _get_microshift_rpm_commit

rh-pre-commit.version: 2.3.2

## Summary by CodeRabbit

* **Refactor**
  * Optimized the microshift bootc build pipeline to extract the upstream RPM source commit once during pipeline initialization instead of repeatedly for each variant, improving build efficiency while ensuring all variants are locked to the same commit source.


rayfordj (approved):
/lgtm
/approve

PR #3063, opened 2026-06-15
fix: use upgrade semantics for update-only lockfile stages

adobes1:
## Problem

4.23 was stuck in an infinite rebuild loop: ~252 images flagged as outdated every scan cycle, rebuilt, then flagged again. Root cause: the lockfile generator was pinning base image package versions instead of upgrading them, so `yum update -y` in hermetic builds had nothing to install.

Specifically, `openssl-3.5.5-3` (from `rhel9-8-els/rhel-minimal:latest`) was being locked into the lockfile even though `openssl-3.5.5-4` was available in `rhel-9-baseos-rpms`.

## Root Cause

For **update-only** stages (no explicit `dnf install`, just `yum update -y`), `_handle_update_only_stage` correctly sets base image packages as `upgradePackages` so DNF fetches the latest versions. However, the final-stage reinstall block then unconditionally called `_get_base_image_packages` again and set those same packages as `reinstallPackages`.

In `rpm-lockfile-prototype`, `base.reinstall()` runs **after** `base.upgrade()` and overrides it. Since the old version (`3.5.5-3`) is still present in the repos, reinstall succeeds and pins the old version. The `PackagesNotAvailableError` fallback to upgrade is never triggered.

The hermetic build then pre-fetches `3.5.5-3` via cachi2, `yum update -y` sees nothing to upgrade, the scan detects the outdated package, and the cycle repeats.

## Fix

Skip setting `reinstallPackages` when the final stage is update-only. The `_handle_update_only_stage` path already populates `packages` and `upgradePackages` with all base image packages, so reinstall is redundant for lockfile coverage and only adds harmful version-pinning.

Non-update-only final stages are unaffected — the `reinstall_pkgs` block still runs for stages with explicit package installs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unnecessary base-image package reinstall/extra package expansion for update-only final stages.

* **Tests**
  * Strengthened assertions to ensure update-only stages skip reinstallPackages behavior and that installed-package lookups occur with the expected frequency.

### DIFF
--- a/doozer/doozerlib/lockfile_prototype/generator.py
+++ b/doozer/doozerlib/lockfile_prototype/generator.py
@@ -412,7 +412,7 @@ class RpmLockfilePrototypeGenerator:
                 if extra:
                     packages = list(packages) + extra
 
-            is_update_only = not packages
+            is_update_only = stage_info.has_update and not packages
             if is_update_only:
                 packages, upgrade_targets, image_pullspec = await self._handle_update_only_stage(
                     stage_num, image_pullspec, distgit_key
--- a/doozer/doozerlib/lockfile_prototype/generator.py
+++ b/doozer/doozerlib/lockfile_prototype/generator.py
@@ -412,7 +412,8 @@ class RpmLockfilePrototypeGenerator:
                 if extra:
                     packages = list(packages) + extra
 
-            if not packages:
+            is_update_only = not packages
+            if is_update_only:
                 packages, upgrade_targets, image_pullspec = await self._handle_update_only_stage(
                     stage_num, image_pullspec, distgit_key
                 )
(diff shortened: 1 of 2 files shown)


adobes1:
[openshift-base-rhel9 test build](https://art-build-history-art-build-history.apps.artc2023.pc3z.p1.openshiftapps.com/build?nvr=openshift-base-rhel9-container-v4.23.0-202606151259.p2.g7d2e767.assembly.test.el9&record_id=4402a4ff-4fb9-488a-03f5-d9345550c855&group=openshift-4.23&outcome=success&type=image) correctly installed `openssl-3.5.5-4.el9_8`
[base-rhel9 test build from logging-6.0](https://art-build-history-art-build-history.apps.artc2023.pc3z.p1.openshiftapps.com/build?nvr=base-rhel9-container-6.0.15-202606151450.p2.g7d2e767.assembly.test.el9&record_id=55ffbd31-c357-55d8-95d2-ee676fb53985&group=logging-6.0&outcome=success&type=image) also correctly installed `openssl-3.5.5-4.el9_8
`

fgallott (approved):

PR #3066, opened 2026-06-16
fix(doozer): fail rebase when art-images-base tag is unreachable

lgarciaaco:
## Summary

Remove the legacy quay.io Konflux pullspec fallback when late-resolving base-image-release parents. Rebase now fails with an explicit error if the `registry.redhat.io/openshift/art-images-base` tag is not reachable.

## Problem

**Before:** Late-resolved base-image-release parents silently fell back to the Konflux `quay.io` pullspec from the build DB when the art-images-base tag was missing.

**After:** Rebase raises `IOError` naming the member and expected art-images-base pullspec. The unused `base_image_release.quay_fallback` config knob and validator schema entries are removed.

## Test plan

- [x] `pytest doozer/tests/backend/test_rebaser.py::TestRebaserResolveMemberParentRegistryRedhat`
- [x] `pytest doozer/tests/test_image.py -k base_image_release`

## Summary by CodeRabbit

* **Bug Fixes**
  * Late-resolving member parents now fail with an error when the required `art-images-base` pullspec is unreachable, instead of falling back to an alternate pullspec.
* **Breaking Changes**
  * Removed the `base_image_release.quay_fallback` option from build-data validation schemas; fallback behavior is no longer configurable.
* **Tests**
  * Updated rebaser and image metadata tests to reflect the new “fail when art-base is missing/unreachable” expectations and removed obsolete fallback coverage.

### DIFF
--- a/doozer/doozerlib/backend/rebaser.py
+++ b/doozer/doozerlib/backend/rebaser.py
@@ -562,9 +562,7 @@ class KonfluxRebaser:
                     rh_pullspec = util.rh_art_images_base_pullspec(build.nvr)
                     if await self._registry_pullspec_exists(rh_pullspec):
                         return rh_pullspec, build.embargoed
-                    raise IOError(
-                        f"Late-resolved parent {member}: art-images-base tag unreachable at {rh_pullspec}"
-                    )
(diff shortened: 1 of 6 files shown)


joepvd (approved):
/lgtm

PR #3067, opened 2026-06-16
fix: use runtime working_dir for rpm-lockfile temp files instead of /tmp

fgallott:
should fix issues like 

```
While writing extension cache /tmp/rpm-lockfile-cache-u63g9s5h/rhel-8-for-aarch64-baseos-rpms-filenames.solvx.wxMvEi (0): repowriter write failed: -1, error: write error blob: No space left on device
```

## Summary by CodeRabbit

* **Refactor**
  * Improved control over temporary file and cache directory placement by introducing a configurable working directory parameter throughout the RPM lockfile generation pipeline. Temporary files and caches are now created under the specified working directory instead of system defaults, enabling better isolation and resource management.

* **Tests**
  * Updated test fixtures to use isolated working directories for resolver and generator initialization.

### DIFF
--- a/doozer/doozerlib/backend/rebaser.py
+++ b/doozer/doozerlib/backend/rebaser.py
@@ -1024,6 +1024,7 @@ class KonfluxRebaser:
             base_dir=self._base_dir,
             downstream_parents=downstream_parents,
             parent_members=parent_members,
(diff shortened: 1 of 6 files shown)


joepvd (approved):
/lgtm

PR #3070, opened 2026-06-17
fix: skip scratch base image in scan-sources builder checks

joepvd:
## Summary
- Images using `from: scratch` have no real parent image, so `oc image info` fails when scan-sources tries to check builder NVR
- Skip the builder change check for `scratch`-based images, since there is no upstream image to compare against

## Test plan
- [ ] Verify scan-sources no longer errors on images with `from: scratch`
- [ ] Verify scan-sources still correctly detects builder changes for normal images


locriandev (approved):
/lgtm

PR #3073, opened 2026-06-18
fix: prevent lockfile-related perpetual rebuild loops and empty lockfile build failures

fgallott:
Four fixes for hermetic build issues with rpm-lockfile-prototype:

1. dockerfile_transforms.py: Add microdnf to the strip_bare_updates regex so bare microdnf update commands are stripped when strip_updates=True.

2. generator.py: Skip reinstallPackages when a stage has a bare update (e.g. microdnf update -y) even when combined with install commands. Same root cause as https://github.com/openshift-eng/art-tools/pull/3063 — reinstall overrides upgrade in rpm-lockfile-prototype, pinning old base image versions and preventing the update from picking up latest RPMs from repos.

3. generator.py: For mixed install+bare-update stages, populate upgrade targets from base image packages. Without this, base image RPMs like `which` don't appear in the lockfile at all — _resolve_bare_update_targets returns empty for bare updates, so inherited packages are never upgraded and the scan keeps detecting them as outdated.

4. rebaser.py: Force strip_updates=True when the lockfile has no packages. When all base image RPMs are already at repo versions, rpm-lockfile-prototype returns an empty lockfile. Without stripping, the bare update tries to refresh unreachable external repos (e.g. cdn-ubi.redhat.com) in hermetic mode and fails with DNS errors.

### DIFF
--- a/doozer/doozerlib/backend/rebaser.py
+++ b/doozer/doozerlib/backend/rebaser.py
@@ -974,9 +974,10 @@ class KonfluxRebaser:
             ):
                 from doozerlib.lockfile_prototype.rebaser_hooks import apply_dockerfile_transforms
 
+                lockfile_has_packages = bool(getattr(metadata, "lockfile_packages", None))
                 apply_dockerfile_transforms(
                     dest_dir,
-                    strip_updates=not downstream_parents,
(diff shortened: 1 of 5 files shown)


rayfordj (approved):
/lgtm
/approve

PR #3080, opened 2026-06-19
ART-19729: Comment on invalid bugs instead of crashing in find-bugs

kopero2000:
## Summary
Implements [ART-19729](https://redhat.atlassian.net/browse/ART-19729): Elliott find-bugs command now comments on invalid bugs instead of crashing.

### Added commenting infrastructure in `bzutil.py`:
- `add_comment_once()` - Base method for idempotent commenting with markers
- `get_comments()` - Fetch existing comments (implemented for both JIRA and Bugzilla)
- `add_invalid_summary_comment()` - Comment on tracker bugs with invalid summary format
- `add_fake_tracker_comment()` - Comment on bugs that look like trackers but aren't valid

### Invalid bug detection:
The command detects three types of invalid bugs:
1. **Fake trackers** - Bugs that look like CVE trackers but are missing required attributes
2. **Invalid summary trackers** - Trackers without proper version tag in summary
3. **Missing builds** - Trackers without matching builds in advisories (TODO)

### Comment format:
- Uses markers like `[ART-AUTOMATION: invalid-summary]` for idempotency
- Markers placed at bottom after horizontal rule to be less obtrusive
- Uses Jira {noformat} tags for proper formatting
- Comments explain what's wrong and how to fix it

## Status
**DRAFT** - Work in progress

### Completed:
- ✅ Commenting infrastructure
- ✅ JIRA comment support
- ✅ Bugzilla comment support  
- ✅ Invalid summary comment handler
- ✅ Fake tracker comment handler

### TODO:
- ⬜ Implement missing builds comment handler
- ⬜ Integrate commenting into categorize_bugs_by_type()
- ⬜ Add --dry-run support
- ⬜ Add tests
- ⬜ Update documentation

## Testing
Tested manually with test JIRA ticket OCPBUGS-90093

## Related
Jira: https://redhat.atlassian.net/browse/ART-19729

## Summary by CodeRabbit

* **New Features**
  * Added `--comment-on-invalid-bugs` to `find-bugs:sweep` to post explanatory comments for invalid tracker bugs (auto-disabled in permissive mode).
* **Improvements**
  * Tracker validation now returns structured results with detailed reasons (e.g., CVE-in-summary, missing keywords/components, missing linked flaws).
  * Invalid-summary tracker bugs and fake trackers now produce per-bug, marker-based comments to prevent duplicates; failures are handled per bug.
* **Tests**
  * Updated unit tests to reflect structured validation results and reason propagation.

### DIFF
--- a/elliott/elliottlib/bzutil.py
+++ b/elliott/elliottlib/bzutil.py
@@ -55,6 +55,12 @@ def get_jira_bz_bug_ids(bug_ids):
 class Bug:
     def __init__(self, bug_obj):
         self.bug = bug_obj
+        self.invalid_bug = {
+            "comment": {
+                "bug_properties": [],
+                "missing_components": []
+            }
+        }
 
     def __str__(self):
         return str(self.id)
@@ -264,7 +270,6 @@ class BugzillaBug(Bug):
         return None
 
     def is_tracker_bug(self):
-        self.invalid_bug.comment.missing_components = []
         has_keywords = set(constants.TRACKER_BUG_KEYWORDS).issubset(set(self.keywords))
         if not has_keywords:
             self.invalid_bug.missing_components.append(f"- This bugs lacks the required keywords: ${set(constants.TRACKER_BUG_KEYWORDS)}")
@@ -279,7 +284,6 @@ class BugzillaBug(Bug):
         if 'WeaknessTracking' in self.keywords:
             # See e.g. https://bugzilla.redhat.com/show_bug.cgi?id=2092289. This bug is not a CVE tracker
             return False
-        self.invalid_bug.comment.bug_properties = []
         has_cve_in_summary = bool(re.search(r'CVE-\d+-\d+', self.summary))
         if has_cve_in_summary:
             self.invalid_bug.comment.bug_properties.append("- Has CVE identifier in summary")
@@ -336,7 +340,6 @@ class JIRABug(Bug):
             return None
 
     def is_tracker_bug(self):
-        self.invalid_bug.comment.missing_components = []
         if self.is_type_vulnerability():
             return True
         has_keywords = set(constants.TRACKER_BUG_KEYWORDS).issubset(set(self.keywords))
@@ -362,7 +365,6 @@ class JIRABug(Bug):
             # Context in this thread: https://redhat-internal.slack.com/archives/C04SCM5AYE4/p1685524912511489?thread_ts=1685489306.568039&cid=C04SCM5AYE4
             # This is likely not the end state, but at least for the time being.
             return False
-        self.invalid_bug.comment.bug_properties = []
         has_cve_in_summary = bool(re.search(r'CVE-\d+-\d+', self.summary))
         if has_cve_in_summary:
             self.invalid_bug.comment.bug_properties.append("- Has CVE identifier in summary")
--- a/elliott/elliottlib/bzutil.py
+++ b/elliott/elliottlib/bzutil.py
@@ -55,12 +55,6 @@ def get_jira_bz_bug_ids(bug_ids):
 class Bug:
     def __init__(self, bug_obj):
         self.bug = bug_obj
-        self.invalid_bug = {
-            "comment": {
-                "bug_properties": [],
-                "missing_components": []
-            }
-        }
 
     def __str__(self):
         return str(self.id)
@@ -143,7 +137,7 @@ class Bug:
     @property
     def product(self):
         raise NotImplementedError
-    
+
     @property
     def bug_class(self) -> str:
         raise NotImplementedError
@@ -248,7 +242,7 @@ class BugzillaBug(Bug):
     @property
     def corresponding_flaw_bug_ids(self):
         return self.bug.blocks
-    
+
     @property
     def bug_class(self):
         return 'bugzilla'
@@ -270,27 +264,44 @@ class BugzillaBug(Bug):
         return None
 
     def is_tracker_bug(self):
+        from elliottlib.cli.find_bugs_sweep_cli import BugValidiationResult
+
         has_keywords = set(constants.TRACKER_BUG_KEYWORDS).issubset(set(self.keywords))
-        if not has_keywords:
-            self.invalid_bug.missing_components.append(f"- This bugs lacks the required keywords: ${set(constants.TRACKER_BUG_KEYWORDS)}")
         has_whiteboard_component = bool(self.whiteboard_component)
-        if not has_whiteboard_component:
-            self.invalid_bug.missing_components.append(f"- Missing Whiteboard component")
-        return has_keywords and has_whiteboard_component
+
+        if not (has_keywords and has_whiteboard_component):
+            missing = []
+            if not has_keywords:
+                missing.append(f"- This bug lacks the required keywords: {set(constants.TRACKER_BUG_KEYWORDS)}")
+            if not has_whiteboard_component:
+                missing.append("- Missing Whiteboard component")
+            return BugValidiationResult(ok=False, reason="\n".join(missing))
+
+        return BugValidiationResult(ok=True, reason="")
 
     def is_invalid_tracker_bug(self):
-        if self.is_tracker_bug():
-            return False
+        from elliottlib.cli.find_bugs_sweep_cli import BugValidiationResult
+
+        tracker_result = self.is_tracker_bug()
+        if tracker_result.ok:
+            return BugValidiationResult(ok=False, reason="")
+
         if 'WeaknessTracking' in self.keywords:
             # See e.g. https://bugzilla.redhat.com/show_bug.cgi?id=2092289. This bug is not a CVE tracker
-            return False
+            return BugValidiationResult(ok=False, reason="")
+
         has_cve_in_summary = bool(re.search(r'CVE-\d+-\d+', self.summary))
-        if has_cve_in_summary:
-            self.invalid_bug.comment.bug_properties.append("- Has CVE identifier in summary")
         has_keywords = set(constants.TRACKER_BUG_KEYWORDS).issubset(set(self.keywords))
-        if has_keywords:
-            self.invalid_bug.comment.bug_properties.append(f"- Has SecurityTracking keywords: {constants.TRACKER_BUG_KEYWORDS}")
-        return has_keywords or has_cve_in_summary
+
+        if not (has_keywords or has_cve_in_summary):
+            return BugValidiationResult(ok=False, reason="")
+
+        properties = []
+        if has_cve_in_summary:
+            properties.append("- Has CVE identifier in summary so this is registered as a bug")
+        properties.append(tracker_result.reason)
+
+        return BugValidiationResult(ok=True, reason="\n".join(properties))
 
     def all_advisory_ids(self):
         return ErrataBug(self.id).all_advisory_ids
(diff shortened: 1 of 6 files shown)


elliott/elliottlib/bzutil.py
```diff
@@ -138,6 +138,10 @@ def component(self):
     def product(self):
         raise NotImplementedError
 
+    @property
+    def bug_class(self) -> str:
+        raise NotImplementedError
+
```
thegreyd, line 156: I'm not sure we need this? isinstance method for this?

elliott/elliottlib/bzutil.py
```diff
@@ -256,19 +264,44 @@ def whiteboard_component(self):
         return None
 
     def is_tracker_bug(self):
+        from elliottlib.cli.find_bugs_sweep_cli import BugValidiationResult
```
thegreyd: let's keep imports on the starting level

elliott/elliottlib/bzutil.py
```diff
@@ -256,19 +264,44 @@ def whiteboard_component(self):
...
+            if not has_whiteboard_component:
+                missing.append("- Missing Whiteboard component")
+            return BugValidiationResult(ok=False, reason="\n".join(missing))
+
+        return BugValidiationResult(ok=True, reason="")
 
     def is_invalid_tracker_bug(self):
-        if self.is_tracker_bug():
-            return False
+        from elliottlib.cli.find_bugs_sweep_cli import BugValidiationResult
```
thegreyd: let's not keep imports here

elliott/elliottlib/bzutil.py
```diff
@@ -318,29 +351,56 @@ def security_level(self):
...
-            return False
+            return BugValidiationResult(ok=False, reason="")
+
         has_cve_in_summary = bool(re.search(r'CVE-\d+-\d+', self.summary))
         has_keywords = set(constants.TRACKER_BUG_KEYWORDS).issubset(set(self.keywords))
         has_linked_flaw = bool(self.corresponding_flaw_bug_ids)
-        return has_keywords or has_cve_in_summary or has_linked_flaw
+
+        if not (has_keywords or has_cve_in_summary or has_linked_flaw):
+            return BugValidiationResult(ok=False, reason="")
```
thegreyd: whenever ok=False, we should pass a reason?

elliott/elliottlib/bzutil.py
```diff
@@ -318,29 +351,56 @@ def security_level(self):
...
             # See e.g. https://issues.redhat.com/browse/OCPBUGS-5804. This is not to be regarded a tracking bug.
-            return False
+            return BugValidiationResult(ok=False, reason="")
         if 'art:cloned-kernel-bug' in self.keywords:
             # Bugs for advance-shipped kernel builds should not be regarded as a tracker. They might look like one,
             # but they are not invalid.
             # Context in this thread: https://redhat-internal.slack.com/archives/C04SCM5AYE4/p1685524912511489?thread_ts=1685489306.568039&cid=C04SCM5AYE4
             # This is likely not the end state, but at least for the time being.
-            return False
+            return BugValidiationResult(ok=False, reason="")
```
thegreyd: same here, ok=False, reason should not be empty?

elliott/elliottlib/bzutil.py
```diff
@@ -597,6 +661,89 @@ def attach_bugs(self, bugids: List, advisory_id: int = 0, advisory_obj: Erratum
...
+        - [openshift-{major_version}.{minor_version}.z]
+        - [openshift-{major_version}.{minor_version}.0]
+
+        Current summary: {bug.summary}
+
+        Please update the bug summary to include the correct target version tag.
+
+        ----
+        {marker}
+        """
```
thegreyd: this is too long of a comment, let's make it concise - 1/2 liner

elliott/elliottlib/bzutil.py
```diff
@@ -597,6 +661,89 @@ def attach_bugs(self, bugids: List, advisory_id: int = 0, advisory_obj: Erratum
...
+Bug evaluation:
+{validation_reason if validation_reason else "This bug has tracker-like properties but is missing required tracker bug attributes."}
+
+Please either:
+1. Complete the tracker bug setup with all required fields, OR
+2. Remove the CVE/SecurityTracking indicators if this is not meant to be a tracker bug
+
+----
+{marker}
+"""
```
thegreyd: Make it as concise as possible

elliott/elliottlib/bzutil.py
```diff
@@ -989,6 +1136,20 @@ def add_comment(self, bugid: str, comment: str, private: bool, noop=False):
...
+        :param bugid: The JIRA bug ID
+        :return: List of comment body strings
+        """
+        try:
+            issue = self._client.issue(bugid)
+            comments = issue.fields.comment.comments
+            return [comment.body for comment in comments]
+        except Exception as e:
+            logger.warning(f"Failed to fetch comments for {bugid}: {e}")
+            return []
```
thegreyd: why catch this exception, why not fail?

elliott/elliottlib/bzutil.py
```diff
@@ -1319,8 +1480,26 @@ def _update_bug_status(self, bugid, target_status):
         return self._client.update_bugs([bugid], self._client.build_update(status=target_status))
 
     def add_comment(self, bugid, comment: str, private, noop=False):
+        if noop:
+            logger.info(f"Would have added a private={private} comment to {bugid}: {comment}")
+            return
         self._client.update_bugs([bugid], self._client.build_update(comment=comment, comment_private=private))
 
+    def get_comments(self, bugid) -> List[str]:
```
thegreyd, line 1454: don't need to implement this right now, you can keep it notimplemented 
because nothing should use it, and we can let it fail

elliott/elliottlib/bzutil.py
kopero2000: I am not sure about it because, that returns `false` in the `is_invalid_tracker_bug` function. It means that if it is false it means the bug is valid, so we can ignore reasons.

elliott/elliottlib/bzutil.py
kopero2000: same as previous comment

elliott/elliottlib/bzutil.py
```diff
@@ -319,28 +360,51 @@ def security_level(self):
...
             # See e.g. https://issues.redhat.com/browse/OCPBUGS-5804. This is not to be regarded a tracking bug.
-            return False
+            return BugValidiationResult(ok=False, reason="")
         if 'art:cloned-kernel-bug' in self.keywords:
             # Bugs for advance-shipped kernel builds should not be regarded as a tracker. They might look like one,
             # but they are not invalid.
             # Context in this thread: https://redhat-internal.slack.com/archives/C04SCM5AYE4/p1685524912511489?thread_ts=1685489306.568039&cid=C04SCM5AYE4
             # This is likely not the end state, but at least for the time being.
-            return False
+            return BugValidiationResult(ok=False, reason="")
```
thegreyd: Validation instead of Validiation

elliott/elliottlib/bzutil.py
thegreyd: we should move all the reason logic inside of is_tracker_bug - and here is_invalid_tracker_bug should either simply be ! is_tracker_bug or it should be deprecated

elliott/elliottlib/bzutil.py
```diff
@@ -597,6 +639,40 @@ def attach_bugs(self, bugids: List, advisory_id: int = 0, advisory_obj: Erratum
     def add_comment(self, bugid, comment: str, private: bool, noop=False):
         raise NotImplementedError
 
+    def get_comments(self, bugid):
+        """Get all comments for a bug. Must be implemented by subclasses."""
+        raise NotImplementedError
+
+    def add_fake_tracker_comment(self, bugid: str, validation_reason: str, private: bool = False, noop=False):
+        """Add comment explaining bug looks like CVE tracker but is invalid."""
+        marker = "[ART-AUTOMATION: fake-tracker]"
```
thegreyd, line 693: add info on how to contact art team, #forum-ocp-art and instead of fake, let's call it invalid publically, to be more polite

elliott/elliottlib/bzutil.py
```diff
@@ -317,30 +350,35 @@ def security_level(self):
...
-        has_keywords = set(constants.TRACKER_BUG_KEYWORDS).issubset(set(self.keywords))
-        has_linked_flaw = bool(self.corresponding_flaw_bug_ids)
-        return has_keywords or has_cve_in_summary or has_linked_flaw
+
+        if not (has_keywords and has_whiteboard_component and has_linked_flaw):
+            invalid_comment = []
+            if not has_cve_in_summary:
+                invalid_comment.append("- Has CVE identifier in summary so this is registered as a bug")
+            if not has_keywords:
+                invalid_comment.append(f"- This bug lacks the required keywords: {set(constants.TRACKER_BUG_KEYWORDS)}")
```
thegreyd: here, include the diff - so what exactly is missing, only that keyword

has_keywords is a bool
constants.TRACKER_BUG_KEYWORDS includes all the keywords, say 3 or 4
if only 1 of the keyword is missing, we should say in the comment which exact one.

elliott/elliottlib/bzutil.py
```diff
@@ -317,30 +350,35 @@ def security_level(self):
...
+            if not has_whiteboard_component:
+                invalid_comment.append("- Missing Whiteboard component")
+            if not has_linked_flaw:
+                invalid_comment.append("- Missing flaw bug links")
```
thegreyd: let's improve the message here - refer to corresponding_flaw_bug_ids method - we look for a label like `flaw:bz#(\d+)` - so we can say - flaw bug label example.. not found

elliott/elliottlib/bzutil.py
kopero2000: if we completely ignore  Bugzilla then we can revert this.

elliott/elliottlib/cli/find_bugs_sweep_cli.py
```diff
@@ -525,12 +535,16 @@ def categorize_bugs_by_type(
 
     # Categorize into tracker and non-tracker bugs
     # while also collecting fake trackers
+    fake_tracker_results = {}  # Store validation results for fake trackers
```
thegreyd, line 538:
```
fake_tracker_results = {b.id for b in fake_trackers}
```

elliott/elliottlib/cli/find_bugs_sweep_cli.py
```diff
@@ -561,7 +575,16 @@ def categorize_bugs_by_type(
             logger.warning(f"{message} Ignoring them.")
             issues.append(message)
         else:
-            raise ElliottFatalError(f"{message} Please fix.")
+            for t in fake_trackers:
```
thegreyd, line 578: why not loop directly over fake_tracker_results here?

elliott/elliottlib/cli/find_bugs_sweep_cli.py
```diff
@@ -561,7 +575,16 @@ def categorize_bugs_by_type(
...
+                # Add comment with the validation result reason
+                validation_result = fake_tracker_results.get(t.id)
+                if validation_result:
```
thegreyd: i don't think we need if here, because if a bug id is in the set, then it is a fake tracker

elliott/elliottlib/cli/find_bugs_sweep_cli.py
```diff
@@ -561,7 +575,16 @@ def categorize_bugs_by_type(
...
+                    try:
+                        runtime.get_bug_tracker(t.bug_class).add_invalid_tracker_comment(
+                            t.id, validation_result.reason, noop=not comment_on_invalid_bugs
+                        )
+                    except Exception as e:
+                        logger.error(f"Failed to create comment for bug {t.id}: {e}")
```
thegreyd: this whole block maybe can be in a class method, since we'll be calling it in several places

elliott/elliottlib/cli/find_bugs_sweep_cli.py
```diff
@@ -588,7 +611,13 @@ def categorize_bugs_by_type(
             logger.warning(f"{message} Ignoring them.")
             issues.append(message)
         else:
-            raise ElliottFatalError(f"{message} Please fix.")
+            for t in invalid_summary_trackers:
+                try:
+                    runtime.get_bug_tracker(t.bug_class).add_invalid_summary_comment(
+                        t, major_version, minor_version, noop=not comment_on_invalid_bugs
```
thegreyd, line 616: before calling the method, check `if comment_on_invalid_bugs` , and then noop should be noop

elliott/elliottlib/cli/find_bugs_sweep_cli.py
thegreyd: i see that they are separate methods

elliott/elliottlib/cli/find_bugs_sweep_cli.py
kopero2000: we do not want to store the id-s we want to store the reason:
```
invalid_result = b.is_invalid_tracker_bug()
if invalid_result:
  fake_trackers.add(b)
  fake_tracker_results[b.id] = invalid_result
```

elliott/elliottlib/cli/find_bugs_sweep_cli.py
kopero2000: we would not know the bug class. If we limit this only to Jira then we can iterate over the results

PR #3119, opened 2026-07-01
switch to cyclonedx

ashwindasr:
rh-pre-commit.version: 2.3.2

## Summary by CodeRabbit

* **Bug Fixes**
  * Build job settings now only include optional overrides when they are actually provided.
  * Fixed an issue where an extra build parameter was being sent unconditionally, helping keep build behavior more consistent with the selected options.


ashwindasr:
this is a no-op since konflux default is now spdx: https://github.com/konflux-ci/build-definitions/blob/main/pipelines/docker-build-oci-ta/README.md

PR #3123, opened 2026-07-02
Redirect rpm-lockfile-prototype RPMDB cache to workspace in Jenkins

joepvd:
## Summary
- When running in Jenkins (`JENKINS_HOME` is set), sets `XDG_CACHE_HOME` in the rpm-lockfile-prototype subprocess env so the RPMDB cache lands under the working directory instead of `~/.cache`
- Avoids filling up the small root volume on Jenkins agents with cached RPMDBs
- Outside Jenkins the behavior is unchanged — the cache stays in `~/.cache`

## Test plan
- [x] Existing unit tests updated and passing
- [x] New tests verify `XDG_CACHE_HOME` is set when `JENKINS_HOME` is present and absent otherwise
- [x] `make test` passes (lint + all 2937 unit tests)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RPM database cache handling by redirecting cache to a persistent workspace location when running under Jenkins.
  * Kept the standard cache behavior for non-Jenkins runs.
  * Made cache cleanup safer and more reliable by validating the computed cache directory before removing stale entries.
* **Tests**
  * Added coverage to verify whether subprocess `XDG_CACHE_HOME` is set when Jenkins is present.
  * Updated cache cleanup tests to validate against the resolver’s computed cache directory rather than a global setting.

### DIFF
--- a/doozer/doozerlib/lockfile_prototype/constants.py
+++ b/doozer/doozerlib/lockfile_prototype/constants.py
@@ -37,10 +37,15 @@ RPM_PSEUDO_PACKAGES = frozenset({"gpg-pubkey"})
 VALID_PKG_NAME = re.compile(r"^[a-zA-Z0-9*][a-zA-Z0-9._+\-*]*$")
 
 
-# rpm-lockfile-prototype stores extracted RPMDBs here. There is no env var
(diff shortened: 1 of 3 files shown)


joepvd:
/hold
not happy with lack of clarity where on disk this ends up.

ashwindasr (approved):

PR #3126, opened 2026-07-06
ART-21515: Don't reinstall Dockerfile packages that are also base image packages

fgallott:
[ironic-agent](https://konflux-ui.apps.kflux-ocp-p01.7ayg.p1.openshiftapps.com/ns/ocp-art-tenant/applications/openshift-4-19/pipelineruns/ose-4-19-ironic-agent-jshx2)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved package handling during lockfile generation so packages already included in the base image are no longer unnecessarily reinstalled.
  * Preserved intended upgrade targets during fallback resolution, even when some packages may not be installed, reducing the risk of missing updates.
* **Tests**
  * Added coverage for overlapping base-image and Dockerfile package scenarios, including fallback behavior under retry exhaustion.


locriandev (commented):
/lgtm

PR #3131, opened 2026-07-08
Fix image filtering to use distgit_key instead of name

fgallott:
The {name} format field in doozer strips dot suffixes (e.g. "image.rhel9" becomes "image"), causing images with RHEL variant suffixes to be incorrectly filtered out as non-existent. Add {distgit_key} format field to images:print and use it in both OCP and OKD images-health pipelines so Redis keys match correctly.

example https://redhat-internal.slack.com/archives/GTDLQU9LH/p1783500150370479

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image filtering to use full distgit keys, which helps match reported failures against the correct image metadata.
  * `images:print` now supports a `{distgit_key}` placeholder, making output more consistent for downstream tooling.
* **Documentation**
  * Clarified the description of valid image output to better reflect the identifiers returned.


joepvd (approved):
/lgtm

PR #3133, opened 2026-07-08
Fix cross-arch RPM version reconciliation to pin max EVR instead of min

ashwindasr:
## Summary

Fixes the `rpm_packages.unique_version` Conforma violation on the golang builder image (and any other image with arch-conditional Dockerfile installs) by fixing how cross-arch RPM version mismatches are reconciled.

## Root cause

The golang builder Dockerfile has an x86_64-only conditional block that installs `libxml2-devel`, `gcc-c++`, etc. Their transitive dependencies (`libxml2`, `libstdc++`, `libsemanage`) resolve to newer versions on x86_64 than on the other arches, since those arches never run that block.

The lockfile generator's `_resolve_with_reconciliation()` correctly *detects* this mismatch, but `_compute_version_pins()` pinned to the **minimum (oldest)** EVR across arches. Pinning down to the older version creates an unsatisfiable dependency on x86_64 (which needs the newer version), so the re-resolution pass fails with `RuntimeError`, and the code silently falls back to the original, mismatched lockfile. The image builds successfully, but the release later fails Conforma:

```
✕ [Violation] rpm_packages.unique_version
  Reason: Mismatched versions of the "libxml2" RPM were found across different arches.
  Platform linux/x86_64 has libxml2-2.9.13-13.el9_4. Platforms linux/arm64, linux/ppc64le, linux/s390x have libxml2-2.9.13-6.el9_4.
```

(same pattern for `libsemanage` and `libstdc++`)

## Fix

Pin to the **maximum (newest)** EVR instead of the minimum in `_compute_version_pins()`. The newest version satisfies the arch-specific dependency's requirement and is available for all arches in the repos, so the re-resolution succeeds and all arches converge on the same version.

## Risk

Low risk — this only changes behavior on the reconciliation path that already runs when a cross-arch mismatch is detected (most images never hit it). If the max-pinned re-resolution still fails for some reason, the existing `except RuntimeError` safety net falls back to the original mismatched lockfile — i.e. the current (broken) behavior — so there's no new failure mode introduced.

## Test plan

- [x] Added `pick_maximum_evr()` to `utils.py` with unit tests
- [x] Updated `_compute_version_pins()` to use it, updated existing reconciliation tests to reflect max-EVR pinning
- [x] `doozer/tests/lockfile_prototype/test_utils.py` and `test_generator.py` pass (119/119)
- [x] `ruff check` / `ruff format --check` pass
- [ ] Re-run the golang-builder Konflux release and confirm the `rpm_packages.unique_version` Conforma violation is resolved

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved package version reconciliation so mismatched cross-architecture builds now align to the newest available version, reducing unexpected downgrades during dependency resolution.
  * Updated resolution behavior to better handle version differences across architectures and epochs.

* **Tests**
  * Expanded coverage for version selection and reconciliation scenarios to reflect the new highest-version matching behavior.

### DIFF
--- a/doozer/doozerlib/lockfile_prototype/generator.py
+++ b/doozer/doozerlib/lockfile_prototype/generator.py
@@ -41,7 +41,7 @@ from doozerlib.lockfile_prototype.models import (
 )
 from doozerlib.lockfile_prototype.resolver import RpmResolver
 from doozerlib.lockfile_prototype.shell_parser import resolve_bash_expansion
-from doozerlib.lockfile_prototype.utils import format_version_pin, pick_minimum_evr
+from doozerlib.lockfile_prototype.utils import format_version_pin, pick_maximum_evr
 from doozerlib.repos import Repos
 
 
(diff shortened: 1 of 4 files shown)


ashwindasr:
/hold

Testing with https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fgolang-builder/627/console

PR #3136, opened 2026-07-09
ART-19542: Remove deprecated separate golang branch logic, always use monobranch

kopero2000:
## Summary

This PR removes the deprecated separate golang branch logic from the `update-golang` pipeline and makes it always use the unified `golang` monobranch layout.

### Update Golang Pipeline (`update_golang.py`)

- **Added deprecation warning**: When `--use-new-golang-branch` flag is not set (default), a warning is logged directing users to the golang monobranch
- **Removed `get_golang_branch()` method**: No longer needed since we always use `'golang'` monobranch
- **Simplified `_get_doozer_var_args()`**: Always returns MAJOR/MINOR OCP version variables (needed for monobranch template substitution)
- **Simplified `_get_doozer_group_and_image()`**: Always uses `GOLANG_DATA_BRANCH = 'golang'` and versioned image keys
- **Simplified `verify_golang_builder_repo()`**: Always checks the `'golang'` monobranch
- **Simplified `should_sign_golang_rpm()`**: Always reads from `'golang'` monobranch group.yml
- **Updated `_build_golang_plashets()`**: Now uses monobranch group name
- **Updated CLI help text**: Marked `--use-new-golang-branch` as deprecated

### Test Updates

- Removed `test_get_golang_branch` (method no longer exists)
- Updated `test_get_doozer_var_args` to expect vars are always returned
- Updated `test_rebase_brew` to expect monobranch group name and --var arguments
- Updated `test_returns_true_when_config_set` and related tests to expect `ref='golang'` instead of separate branch names
- Updated plashet tests to expect monobranch group name

## Backward Compatibility

The `--use-new-golang-branch` parameter is kept to:
1. Show deprecation warnings to users still using the old default
2. Maintain CLI compatibility (no breaking changes)

The flag is now effectively ignored - the pipeline always uses the monobranch approach.

Unused `--var` arguments passed to doozer are harmless - they're simply ignored if the group.yml doesn't have corresponding template placeholders.

## Dependencies

This PR depends on the plashet support for golang groups being merged first. See related PR for adding version parameter support to build-plashets pipeline.

## Testing

- [x] All unit tests pass
- [x] Verified deprecation warning appears when flag is not set
- [x] Confirmed monobranch group name is used throughout

## Related

- ocp-build-data golang monobranch: https://github.com/openshift-eng/ocp-build-data/tree/golang

### DIFF
--- a/pyartcd/pyartcd/pipelines/update_golang.py
+++ b/pyartcd/pyartcd/pipelines/update_golang.py
@@ -189,7 +189,6 @@ class UpdateGolangPipeline:
         skip_pr: bool = False,
         external_golang_rpms: bool = False,
         network_mode: str | None = None,
-        use_new_golang_branch: bool = False,
         major_bump: bool = False,
     ):
         self.runtime = runtime
@@ -209,20 +208,11 @@ class UpdateGolangPipeline:
         self.skip_pr = skip_pr
         self.external_golang_rpms = external_golang_rpms
         self.network_mode = network_mode
-        self.use_new_golang_branch = use_new_golang_branch
         self.major_bump = major_bump
         self._slack_client = self.runtime.new_slack_client()
         self._doozer_working_dir = self.runtime.working_dir / "doozer-working"
         self._doozer_env_vars = os.environ.copy()
         self._branch_content = None
-
-        # Deprecation warning for separate golang branches
-        if not use_new_golang_branch:
-            _LOGGER.warning(
-                "Using unique golang version branches is deprecated. "
-                "Will use golang mono branch instead. "
-                "See: https://github.com/openshift-eng/ocp-build-data/tree/golang"
-            )
 
         # Get kubeconfig from environment variable if not provided
         if not kubeconfig:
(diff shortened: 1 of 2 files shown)


pyartcd/pyartcd/pipelines/build_plashets.py
```diff
@@ -32,6 +32,8 @@ def __init__(
         self.data_path = data_path
         self.data_gitref = data_gitref
         self.copy_links = copy_links
+        self.version = version
+        self.override_ocp_version = override_ocp_version
```
thegreyd: i don't think the override_ocp_version param is needed. the artcd command should accept both version and group params and then the python code can validate and complain if both are passed for anything other than the golang branch. the golang branch is special since right now it is the only branch that uses the pattern of major.minor variables being passed at runtime. so we won't be using these params for anything else, and it is okay for the python code to know about this and raise error in any other case

pyartcd/pyartcd/pipelines/update_golang.py
```diff
@@ -1172,7 +1165,7 @@ def get_module_tag(self, nvr, el_v) -> str:
     '--use-new-golang-branch',
     is_flag=True,
     default=False,
-    help='Use the unified "golang" branch layout in ocp-build-data instead of per-variant branches (e.g. rhel-9-golang-1.26).',
+    help='[DEPRECATED] This flag is now ignored. The pipeline always uses the unified "golang" monobranch layout.',
```
thegreyd: we can remove this flag entirely - as it can create confusion just by existing

PR #3137, opened 2026-07-09
ART-21561: Fix bundle CSV namespace references for OCP 5.0

tomasdavidorg:
Auto-derive delivery namespace from MAJOR version in group.yml
  instead of using hardcoded 'openshift4' default.

  When building OLM bundles, placeholder images (e.g., origin-*) are
  not found in delivery_namespace_map and fall back to the default.
  Previously hardcoded 'openshift4', this caused OCP 5.0 bundles to
  reference wrong namespace, resulting in stage-testing timeouts.

  Solution:
  - Read MAJOR from group.yml vars
  - For OCP 5.x: use 'openshift5'
  - For OCP 4.x: use 'openshift4' (backward compatible)
  - Future: automatically use 'openshift{MAJOR}'

  After bundle rebuild, CSV will correctly reference:
    registry.redhat.io/openshift5/operator-name@sha256:... Instead of: registry.redhat.io/openshift4/operator-name@sha256:...

  Fixes: https://redhat.atlassian.net/browse/ART-21561
  Parent: https://redhat.atlassian.net/browse/ART-14465

rh-pre-commit.version: 2.4.0

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the OpenShift delivery-namespace fallback to derive `openshift{N}` from the configured platform major version, defaulting to `4` when unset.
  * Improved operator image reference rewriting to target the correct Red Hat registry pullspec format for major versions 4 vs 5 (including digest formatting differences).
  * Added clear validation and error reporting when the major version value cannot be parsed.
* **Tests**
  * Added async unit tests verifying major-version-specific image rewrite behavior for both 4.x and 5.x.

### DIFF
--- a/doozer/doozerlib/backend/konflux_olm_bundler.py
+++ b/doozer/doozerlib/backend/konflux_olm_bundler.py
@@ -427,7 +427,7 @@ class KonfluxOlmBundleRebaser:
                 f"Invalid MAJOR version in group config for {metadata.runtime.group}: "
                 f"{self._group_config.vars.get('MAJOR')}"
             ) from e
-        default_delivery_namespace = f'openshift{major}' if major >= 5 else 'openshift4'
+        default_delivery_namespace = f'openshift{major}'
         # Build a map of image short name -> delivery override short name for images that have
         # delivery_repo_name_override set. This allows operators to reference images using a
         # versioned internal name (e.g. ose-csi-livenessprobe-4.18-rhel9) while having them
--- a/doozer/doozerlib/backend/konflux_olm_bundler.py
+++ b/doozer/doozerlib/backend/konflux_olm_bundler.py
@@ -462,8 +462,11 @@ class KonfluxOlmBundleRebaser:
             if not metadata.runtime.group.startswith("openshift-"):
                 new_namespace = namespace
             else:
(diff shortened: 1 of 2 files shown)


doozer/doozerlib/backend/konflux_olm_bundler.py
```diff
@@ -418,6 +418,16 @@ async def _replace_image_references(self, old_registry: str, content: str, engin
...
+        # Derive default delivery namespace from MAJOR version (openshift5 for OCP 5.x, openshift4 for older)
+        try:
+            major = int(self._group_config.vars.get('MAJOR', 4))
+        except (ValueError, TypeError) as e:
+            # MAJOR should always be a valid integer in group.yml; if not, this is a config error
+            raise ValueError(
+                f"Invalid MAJOR version in group config for {metadata.runtime.group}: "
+                f"{self._group_config.vars.get('MAJOR')}"
+            ) from e
+        default_delivery_namespace = f'openshift{major}' if major >= 5 else 'openshift4'
```
fgallott: I think `if major >= 5 else 'openshift4'` is useless here, given 
 
`major = int(self._group_config.vars.get('MAJOR', 4))`

tomasdavidorg: Thanks. Fixed

fgallott (approved):

PR #3142, opened 2026-07-10
ART-21571: Auto-drop empty RPM/RHCOS advisories in promote pipeline

joepvd:
## Summary

When an advisory has no builds attached (e.g. z-stream releases with no RPM or RHCOS changes), the promote pipeline previously attempted to move it to QE state, which failed with "No builds in advisory". This required manual cleanup via the `drop_advisories` Jenkins job.

This change automatically detects and drops empty advisories during the promote pipeline, before the QE state change loop.

Based on #3141, with the following improvements:
- **No code duplication**: Extracts the repair-bugs + advisory-drop logic from `advisory_drop.py` into a reusable `drop_advisory()` function, called from both the CLI command and the promote pipeline
- **Handles RHCOS too**: Checks both `rpm` and `rhcos` advisories (both can legitimately be empty in z-streams)
- **Retries with tenacity**: `_check_and_drop_empty_advisory` retries 3 times with 30s wait for transient errata API failures
- **No broad error swallowing**: Uses `logger.exception` for full tracebacks; tenacity handles transient failures at the inner level, outer method catches per-advisory so one failure doesn't block others

### `pyartcd/pyartcd/pipelines/advisory_drop.py`
- Extracted core logic into reusable `drop_advisory()` async function with `group`, `advisory`, `env`, `dry_run` params
- CLI command (`advisory-drop`) now delegates to `drop_advisory()`

### `pyartcd/pyartcd/pipelines/promote.py`
- `_drop_empty_advisories()`: iterates droppable impetuses (`rpm`, `rhcos`), catches errors per-advisory
- `_check_and_drop_empty_advisory()`: checks builds via `get_builds()`, drops if empty via `drop_advisory()`, decorated with `@retry`
- Inserted call before QE state change loop in `_run_pipeline()`

### `pyartcd/tests/pipelines/test_promote.py`
- 9 new tests across 3 test classes covering: `drop_advisory()` function, per-advisory check, and orchestrator logic

## Test plan
- [x] `make test` passes (2975 tests across all packages)
- [ ] Verify in staging with a release that has an empty RPM advisory
- [ ] Verify dry-run mode logs correctly without dropping

Fixes: https://redhat.atlassian.net/browse/ART-21571

## Summary by CodeRabbit

* **New Features**
  * During standard promotion, the system now detects empty RPM and RHCOS advisories and drops them automatically.
  * Advisory dropping is now centralized and reused by both the promotion flow and the `advisory-drop` command, with a consistent default comment.
  * Dry-run mode previews drops without making external changes.
* **Bug Fixes**
  * Promotion continues toward QE even if checking or dropping an individual advisory fails; errors are reported without stopping the rest.
  * Advisories that still have associated builds are no longer dropped.

### DIFF
--- a/pyartcd/pyartcd/pipelines/advisory_drop.py
+++ b/pyartcd/pyartcd/pipelines/advisory_drop.py
@@ -1,4 +1,6 @@
+import logging
 import os
+from typing import Optional, Union
 
 import click
 from artcommonlib import exectools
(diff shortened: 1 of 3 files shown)


pyartcd/pyartcd/pipelines/advisory_drop.py
```diff
@@ -36,13 +63,31 @@ async def advisory_drop(runtime: Runtime, group: str, advisory: str, comment: st
...
+@click.option(
+    '--comment',
+    required=False,
+    default=DEFAULT_DROP_COMMENT,
+    help='Comment will add to the bug attached on the advisory to explain the reason',
+)
+@pass_runtime
+@click_coroutine
+async def advisory_drop_cli(runtime: Runtime, group: str, advisory: str, comment: str):
+    await drop_advisory(group=group, advisory=int(advisory), comment=comment)
```
fgallott: do we really need 

`advisory=int(advisory)` here, and then 

`adv_str = str(advisory)` in drop_advisory?

joepvd: Good catch. Changed `drop_advisory` to accept `Union[int, str]` and convert to `str` once internally. The CLI now passes the string straight through without the `int()` round-trip.

fgallott (approved):

PR #3159, opened 2026-07-15
ART-21639: preserve reinstall commands in Konflux Dockerfiles

fgallott:
Previously, strip_reinstall_commands() unconditionally removed all dnf/microdnf/yum reinstall commands from Dockerfiles during Konflux builds. This caused semantically important re-extractions (e.g. reinstall tzdata for timezone data) to be silently dropped, breaking images that depend on the reinstalled package files.

Two fixes:
- Parse reinstall as a recognized action in the shell parser so reinstall packages are included in the lockfile resolution
- Transform reinstall commands to (cmd || true) instead of stripping them, so they succeed when the NEVRA matches and degrade gracefully when it does not

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

rh-pre-commit.version: 2.4.0

## Summary by CodeRabbit

* **Bug Fixes**
  * Reinstall package commands are no longer removed during hermetic builds; they are rewritten to fail safely when referenced packages aren’t available.
  * Package discovery now recognizes `reinstall` for `dnf`, `yum`, and `microdnf`, ensuring those packages are included in processing.
* **Tests**
  * Updated and expanded coverage to validate the new reinstall-command behavior, including single, multiple, and mixed install/reinstall flows.

### DIFF
--- a/doozer/doozerlib/lockfile_prototype/dockerfile_transforms.py
+++ b/doozer/doozerlib/lockfile_prototype/dockerfile_transforms.py
@@ -76,13 +76,18 @@ def transform_reinstall_commands(df_content: str) -> str:
         str: Transformed Dockerfile text with reinstall commands wrapped.
     """
     reinstall_re = re.compile(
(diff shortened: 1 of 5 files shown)


rayfordj (approved):
/lgtm
/approve

PR #3194, opened 2026-07-23
Fix CoreOS image mirroring to use ocp-priv namespace

fgallott:
Test Platform moved from ocp-private to ocp-priv, causing build-sync-konflux to fail when tagging CoreOS images for private CI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated private CI image tagging to use the correct `ocp-priv` image stream and namespace.
  * Improved related logging to accurately reflect the updated destination.


joepvd (approved):
/lgtm

PR #3093, opened 2026-06-24
sync-ci-images: fix --stream handling, add --images filter

joepvd:
## Summary

Two fixes for the `sync-ci-images` pipeline:

1. **Fix `--only-streams` → `--stream`**: The pipeline passed `--only-streams` as a global doozer option, but doozer has no such flag. Replaced with `--stream` at the subcommand level where doozer actually accepts it.

2. **Add `--images` filter**: New `--image` option on doozer's `images:streams` subcommands and a corresponding `--images` parameter on the pyartcd pipeline. Allows targeting specific distgit keys (e.g. `ci-openshift-base.rhel10`) without running a full sync.

## Test plan
- [x] All existing unit tests pass (`make test`)
- [x] New unit tests for `--image` filter parsing and selection behavior

PR #3196, opened 2026-07-23
fix(sync-ci-images): stop assuming oc image mirror preserves manifest digests

fgallott:
## Summary

- Reverts the `source_digest` optimization from PR #3093 that assumed `oc image mirror` preserves manifest digests across registries
- When manifest format conversion occurs (OCI↔Docker v2s2), the digest changes, so the source digest cannot be used as the QCI digest
- Restores the original behavior of always querying QCI for the actual digest after mirroring

## Root cause

PR #3093 added an optimization: extract the digest from Konflux `@sha256:` source pullspecs and reuse it as the QCI digest for GC-prevention tagging, skipping the `get_image_digest()` call. This fails for `openshift-base-nodejs` because `oc image mirror` converts the manifest format, producing a different digest at QCI.

**Impact**: `sync-ci-images` job has been failing for ALL OCP versions since July 14 (~71 consecutive failures, builds #629–#699). All CI base images including `base-rhel9` are stale (last updated June 8), blocking local container builds for downstream teams (e.g., ironic team).

## Test plan

- [x] `uv run pytest doozer/tests/cli/test_images_streams.py` — 28 passed
- [ ] Trigger `sync-ci-images` Jenkins job for a single version to verify fix

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image mirroring reliability by always refreshing the image digest from the floating tag after the mirror operation.
  * Updated ARM64 mirroring to follow the same refreshed digest behavior.
  * Removed prior digest reuse logic that could lead to stale image references.

### DIFF
--- a/doozer/doozerlib/cli/images_streams.py
+++ b/doozer/doozerlib/cli/images_streams.py
@@ -172,7 +172,7 @@ def images_streams_mirror(
     elif runtime.registry_config_dir is not None:
         registry_config_file = get_docker_config_json(runtime.registry_config_dir)
 
-    def mirror_image(cmd_start: str, upstream_dest: str, source_digest: Optional[str] = None):
+    def mirror_image(cmd_start: str, upstream_dest: str):
         if upstream_dest.startswith('registry.ci.openshift.org/'):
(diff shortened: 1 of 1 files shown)


rayfordj (approved):
/lgtm
/approve

PR #3208, opened 2026-07-27
Fix monobranch Konflux builder lookup

thegreyd:
## What

- resolve the Konflux builder-record name through the same group/image helper
  used by the rebase and build steps
- query versioned image keys such as
  `openshift-golang-builder-1-26.rhel8` for the Golang monobranch
- recognize Golang builder records by their stable
  `openshift-golang-builder-container` NVR component, independent of the
  metadata key layout
- retain the existing unversioned `openshift-golang-builder` lookup for legacy
  per-version branches
- handle an empty installed-Golang mapping without leaking `StopIteration`

## Why

After a successful Golang monobranch Konflux build and base-image release,
`update-golang` could not rediscover the build record it had just stored:

```text
Failed to find existing builder(s) for rhel version(s): Konflux: {8}
```

Doozer correctly stores Konflux records under the metadata `distgit_key`. On
the monobranch that key is versioned, but the post-build lookup was still
hard-coded to query the legacy unversioned key.

After correcting that query, Elliott exposed a second layout assumption: it
only treated a record as the Golang builder when `build.name` was exactly
`openshift-golang-builder`. A versioned record was treated as an ordinary child
image, producing an empty Golang-NVR mapping and:

```text
coroutine raised StopIteration
```

Using `_get_doozer_group_and_image()` aligns the database lookup with the
selected build layout. Using the NVR component identifies the builder
consistently across both layouts and remains compatible with the planned
universal monobranch behavior in #3136.

Related: #3206

## Impact

Successful monobranch builders can be rediscovered, validated against their
installed Golang RPM, and proceed to streams updates and the remaining
pipeline steps. Legacy branch lookups are unchanged.

## Verification

- focused Elliott regression: 1 passed
- all Elliott utility tests: 6 passed
- all `update-golang` tests: 92 passed
- `make format`
- `make test`: 3,084 tests passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of existing Go builder images for each platform and Go version in Konflux.
  * Updated Go-version extraction for relevant Konflux records to rely on parsed container identifiers, improving accuracy for monobranch builds.
  * Added test coverage to validate monobranch-specific lookup behavior, including cases where no matching builder metadata is found.

### DIFF
--- a/elliott/elliottlib/util.py
+++ b/elliott/elliottlib/util.py
@@ -565,7 +565,7 @@ def get_golang_container_nvrs_for_konflux_record(
         # because we need to look at included golang rpms instead of parent images.
         # Unlike `get_golang_container_nvrs_brew`, we don't accept 'go-toolset*' containers
         # because they don't exist in KonfluxDB.
-        if build.name == GOLANG_BUILDER_IMAGE_NAME:
+        if nvr_dict['name'] == constants.GOLANG_BUILDER_CVE_COMPONENT:
             go_package = next(
                 (
                     pkg_nvr
--- a/elliott/tests/test_util.py
+++ b/elliott/tests/test_util.py
@@ -1,4 +1,5 @@
 import unittest
+from unittest.mock import Mock
 
 from elliottlib import brew, util
 from elliottlib.bzutil import Bug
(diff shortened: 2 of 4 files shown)


thegreyd:
Tested at https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fgolang-builder/679/console

PR #3209, opened 2026-07-28
ART-21518: Fix non-latest Golang builder stream updates

thegreyd:
## Summary

- Skip RPM tagging, buildroot availability checks, and plashet builds for configured non-`GO_LATEST` lines.
- Discover Konflux builders under both versioned and legacy record names, regardless of the Golang branch-layout flag, and require an exact installed compiler NVR match.
- Reuse existing non-`GO_LATEST` builders; fail before rebase when a required builder is missing.
- Preserve missing-builder rebase/build behavior with `--external-golang-rpms`.
- Update `GO_EXTRA` streams using either templated aliases or literal versioned keys, including streams sharing the selected builder image.

## Root cause

The pipeline accepted `GO_EXTRA` but still unconditionally processed its RPMs. After that was skipped, a monobranch lookup searched only versioned Konflux record names, so legacy builders were treated as missing and the normal missing-image path rebased them.

## External RPM behavior

`--external-golang-rpms` still permits a missing builder to be rebased and built. It assumes the selected golang-builder metadata enables the external repository containing the requested RPM; after the build, the pipeline rediscovers the builder and verifies its installed Golang NVR exactly.

## Validation

- Focused update-golang tests: 96 passed
- Ruff lint and formatting passed
- `git diff --check` passed
- Global rh-pre-commit and commit-message hooks passed

### DIFF
--- a/pyartcd/pyartcd/pipelines/update_golang.py
+++ b/pyartcd/pyartcd/pipelines/update_golang.py
@@ -403,6 +403,18 @@ class UpdateGolangPipeline:
         )
 
         if brew_missing or konflux_missing:
+            if not process_rpm_builds and not self.external_golang_rpms:
+                missing = []
+                if brew_missing:
+                    missing.append(f'Brew: {sorted(brew_missing)}')
+                if konflux_missing:
+                    missing.append(f'Konflux: {sorted(konflux_missing)}')
+                raise ValueError(
+                    "Cannot build missing non-GO_LATEST golang builder image(s) without "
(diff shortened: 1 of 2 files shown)


thegreyd:
Tested at https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fgolang-builder/690

thegreyd:
Merging as this is fixing a critical pipeline bug.

PR #3214, opened 2026-07-31
Add bundle_name_override config field to allow custom OLM bundle short names

ashwindasr:
## Summary

Adds a top-level `bundle_name_override` field to `ocp-build-data` image config, allowing an operator's OLM bundle short name to differ from the default `{distgit_key}-bundle` derivation. This is useful when an operator's bundle needs a name that doesn't follow the standard convention (e.g. to match an existing delivery/component name).

- Added `bundle_name_override` (+ assembly override variants `!`, `?`, `-`) to the `image_config.base.schema.json` schema.
- `ImageMetadata.get_olm_bundle_short_name()` now returns the override when set, falling back to `{distgit_key}-bundle` otherwise.
- Updated all downstream consumers to go through this method instead of re-deriving the name themselves:
  - Legacy `OLMBundle.bundle_name` (Brew path)
  - `doozer/doozerlib/cli/fbc.py` and `cli/images_konflux.py` bundle build lookups
  - `KonfluxFbcRebaser._find_component_in_shipment` (tries the override before the default `{name}-bundle` pattern)
- Extended the `doozer olm-bundle:list-olm-operators` CLI with a new `--output-format=bundle-name` mode (emits `{distgit_key}\t{bundle_short_name}` pairs) so pyartcd pipelines (which shell out to doozer rather than importing it) can resolve the correct bundle name without duplicating the derivation logic.
- Updated `pyartcd/pipelines/scan_operator.py` and `build_conforma_verify.py` to use that mapping instead of hardcoding `{name}-bundle`, and to track operator name via the query result rather than reverse-deriving it from the bundle name with `removesuffix('-bundle')` (which breaks under an override).

Alongside https://github.com/openshift-eng/ocp-build-data/pull/12079

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring custom OLM bundle names through image metadata.
  * Bundle discovery, component matching, and build verification now honor custom names while retaining fallback naming.
  * Added a `bundle-name` output format to the OLM operator listing command.
  * Bundle names are derived from operator metadata when available.

* **Bug Fixes**
  * Improved matching for operators with non-standard bundle names.
  * Corrected the product key name for Multicluster Engine.

* **Documentation**
  * Documented bundle-name configuration and validation requirements.

### DIFF
--- a/artcommon/artcommonlib/constants.py
+++ b/artcommon/artcommonlib/constants.py
@@ -90,7 +90,7 @@ GOLANG_NVR_ENV = '__doozer_golang_nvr'
 # Product-based mappings for Konflux tenant namespaces and kubeconfigs
 PRODUCT_NAMESPACE_MAP = {
     "acm": "art-acm-tenant",
-    "mce": "art-acm-tenant",
+    "multicluster-engine": "art-acm-tenant",
     "cert-manager": "art-oap-tenant",
     "external-secrets": "art-oap-tenant",
     "installer-ove-ui": "art-installer-agent-tenant",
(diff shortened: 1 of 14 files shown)


ashwindasr:
/hold

Test build triggered: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Flayered-products/11898/

ashwindasr:
testing with stream: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Flayered-products/11962/

PR #3226, opened 2026-08-05
Skip Konflux component update when spec already matches

ashwindasr:
## Summary

- `ensure_component()` previously always did a full PUT replace of the Kubernetes Component resource, even when it already had the correct values
- When multiple Jenkins jobs build different OCP versions concurrently, concurrent PUTs all hit the namespace ResourceQuota admission controller simultaneously, causing 409 Conflicts on the quota object: _\"Operation cannot be fulfilled on resourcequotas \\\"konflux\\\": the object has been modified\"_
- Add `_component_matches()` to compare desired spec fields against the existing component; if they match, return the existing resource without issuing a replace
- Annotations are excluded from the comparison because Konflux controllers (e.g. the PaC controller) may extend annotation values after creation, which would cause spurious mismatches and defeat the optimization

## Test plan

- [ ] `uv run pytest --verbose doozer/tests/backend/test_konflux_client.py` — new `TestComponentMatches` and `TestEnsureComponent` classes cover: match→skip, mismatch→replace, not-found→create, and extended-annotation tolerance
- [ ] `uv run pytest --verbose doozer/tests/backend/test_konflux_image_builder.py doozer/tests/backend/test_konflux_olm_bundler.py doozer/tests/backend/test_konflux_fbc.py` — no regressions

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved component reconciliation by preserving existing components when their configuration already matches.
  * Automatically replaces components when application, source, image, revision, or annotation settings differ.

* **Tests**
  * Added coverage for matching, mismatched, optional, and missing component configurations.

### DIFF
--- a/doozer/doozerlib/backend/konflux_client.py
+++ b/doozer/doozerlib/backend/konflux_client.py
@@ -838,6 +838,36 @@ class KonfluxClient:
         """
         return await self._get__caching(API_VERSION, KIND_COMPONENT, name, strict=strict)
 
+    @staticmethod
+    def _component_matches(
+        existing: resource.ResourceInstance,
+        application: str,
+        component_name: str,
+        image_repo: Optional[str],
+        source_url: Optional[str],
+        revision: Optional[str],
+    ) -> bool:
+        """Return True if the existing component already has the desired spec values.
+
+        Only spec fields are compared. Annotations are intentionally excluded because
+        Konflux controllers (e.g. the PaC controller) may extend annotation values with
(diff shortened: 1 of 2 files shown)


lgarciaaco:
/lgtm

This will cut most of the ResourceQuota 409s we saw on the 5.1 mass rebuild. Skipping the replace when the Component spec already matches is the right fix for the common path.

One leftover race: if two jobs both miss the first GET and then one creates, the other can still replace inside `_create_or_replace` even when the specs match. Narrower than before and fine as a follow-up if you want to keep this PR small.

PR #3239, opened 2026-08-06
fix: images-health reports exact failure count for new images

fgallott:
## Summary

- `FAILING_AT_LEAST_FOR` code (which prints "Failing for at least 100 attempts") was set whenever no successful build was found in the query results, regardless of how many builds actually exist
- For new groups like `openshift-5.1` with only 5 builds, this produced a misleading "at least 100 attempts" message (spotted in #art-release-bot)
- Fix: only use `FAILING_AT_LEAST_FOR` when results are truncated at the query limit (truly unknown depth); use `LATEST_ATTEMPT_FAILED` with the exact count otherwise

## Test plan

- [ ] New unit tests cover both branches: fewer-than-limit → `LATEST_ATTEMPT_FAILED`, at-limit → `FAILING_AT_LEAST_FOR`
- [ ] All 12 existing tests still pass (`uv run pytest doozer/tests/cli/test_images_health.py`)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved image health reporting when no successful build is available.
  - Reports whether the failure count is exact or may exceed the available results.
  - Provides the precise number of failed builds when all attempts are available.
  - Indicates an unknown minimum count when results reach the query limit.

### DIFF
--- a/doozer/doozerlib/cli/images_health.py
+++ b/doozer/doozerlib/cli/images_health.py
@@ -152,11 +152,19 @@ class ImagesHealthPipeline:
             )
 
         elif latest_success_idx == -1:
-            # No success record was found: add a concern
+            # No success record was found: distinguish between hitting the query limit (truly unknown
+            # how far back failures go) vs. knowing the exact count (fewer results than the limit)
+            if len(builds) == self.limit:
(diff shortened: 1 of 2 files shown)


locriandev (approved):
/lgtm

PR #3251, opened 2026-08-07
Support stream and test assemblies in update-golang

thegreyd:
## Summary

- add an assembly option to update-golang, defaulting new runs to stream and accepting stream or test
- propagate the assembly through Golang image rebase, build, plashet creation, repository lookup, and notifications
- preserve legacy unqualified-build lookup for existing stream builds while requiring assembly-qualified new builds
- make test builds Konflux-only; reject brew and both because successful Brew builds update floating tags
- keep release-facing operations on stream only: test skips published pullspec checks, streams.yml updates and PRs, and bug updates
- reject named/non-stream-type assemblies in this pipeline

## Behavior

stream is the default for all new pipeline runs. test provides isolated Konflux test rebases and builds. Brew test-assembly builds are rejected before pipeline work because Brew floating tags are updated after every successful build. Existing builds without an assembly suffix remain discoverable as stream builds for backward compatibility; the pipeline does not offer a new unqualified-build mode.

## Validation

- 126 focused update-golang and rebuild-golang-rpms tests passed, including Brew and both rejection subtests
- Ruff check and formatting passed
- Doozer base-image release assembly test passed
- live Konflux no-push rebases succeeded for stream and test, producing assembly-qualified NVRs
- Brew stream and Konflux stream/test build command paths were exercised with --dry-run; execution reached the external services, where local credentials/network access blocked submission
- git diff --check

## Dependency

Requires openshift-eng/ocp-build-data#12202.

## Summary by CodeRabbit

- **New Features**
  - Added support for selecting `stream` or `test` assemblies when updating Golang.
  - Assembly selection is reflected in builds, repository locations, commands, and related update operations.
  - Improved compatibility when locating existing Golang builds.

- **Bug Fixes**
  - Added validation for supported assembly configurations and unified-branch requirements.
  - Test assemblies now require the appropriate branch and platform configuration.
  - Test assemblies skip production-only publishing, stream, and bug-tracking updates.
  - Improved cached-content and dry-run behavior across assemblies.

### DIFF
--- a/pyartcd/pyartcd/pipelines/update_golang.py
+++ b/pyartcd/pyartcd/pipelines/update_golang.py
@@ -21,7 +21,7 @@ from artcommonlib.constants import (
 from artcommonlib.github_auth import get_github_client_for_org
 from artcommonlib.konflux.konflux_build_record import ArtifactType, Engine, KonfluxBuildOutcome, KonfluxBuildRecord
 from artcommonlib.konflux.konflux_db import KonfluxDb
-from artcommonlib.release_util import isolate_el_version_in_release
+from artcommonlib.release_util import isolate_assembly_in_release, isolate_el_version_in_release
 from artcommonlib.rpm_utils import parse_nvr
 from artcommonlib.util import new_roundtrip_yaml_handler
 from doozerlib.cli.config_plashet import KNOWN_SIGNING_KEYS
@@ -39,6 +39,8 @@ from pyartcd.util import default_release_suffix, kinit
 _LOGGER = logging.getLogger(__name__)
 yaml = new_roundtrip_yaml_handler()
 PUBLISHED_GOLANG_BUILDER_REPO = f"{REGISTRY_REDHAT_IO}/openshift/{ART_IMAGES_BASE_APPLICATION}"
+GOLANG_ASSEMBLIES = ("stream", "test")
+DEFAULT_GOLANG_ASSEMBLY = "stream"
 
 
 def is_latest(ocp_version: str, el_v: int, nvr: str, koji_session) -> bool:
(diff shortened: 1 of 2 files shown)


thegreyd:
tested at https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fgolang-builder/705/consoleFull

pyartcd/pyartcd/pipelines/update_golang.py
```diff
@@ -726,47 +788,61 @@ async def get_existing_builders_konflux(
...
-                    _LOGGER.warning("Could not determine installed Golang RPM for %s", build_record.nvr)
-                    continue
-                actual_go_nvr, _ = next(iter(go_nvr_map.items()))
-                if actual_go_nvr != expected_go_nvr:
-                    _LOGGER.warning(
-                        "Installed golang rpm in %s is different from the expected one: %s != %s",
-                        build_record.nvr,
-                        actual_go_nvr,
-                        expected_go_nvr,
+                        limit=50 if assembly_filter is None else 1,
```
lgarciaaco: /lgtm

Not sure if this would actually bite us in practice, but worth running by:

With `limit=1` on the assembly-qualified path in `get_existing_builders_konflux`, a golang RPM mismatch on the newest record makes `find_builder` return `None` and we kick off a full rebuild even when an older matching build is already in the DB. The unqualified legacy path already uses `limit=50` for this.

Say build #1 (older) is assembly=`stream`, golang RPM = `go1.22.0`. Build #2 (newest) is assembly=`stream`, golang RPM = `go1.22.1`. Pipeline expects `go1.22.0`, so we only fetch build #2, hit the RPM mismatch, `continue`, and return `None`. Unnecessary full rebuild. With `limit=50` we walk back, find build #1, and reuse it.

I'd drop the `else 1` and use `limit=50` unconditionally so the assembly path matches the legacy behavoir.

PR #3256, opened 2026-08-10
ART-18061: Filter operand NVR resolution by assembly

fgallott:
Explicit assembly filtering in bundle rebase operand resolution to prevent test/stream assembly mismatches. When vector was built in test assembly and CLO's bundle rebase queried for latest vector without assembly filtering, it pulled in the test build, causing Conforma violations ([ART-21925](https://redhat.atlassian.net/browse/ART-21925)).

Changes:
- Pass assembly=self.assembly to get_latest_konflux_build() calls
- Same for embargo-substitution fallback build
- Assert resolved build's assembly matches expected assembly, fail fast if not

Prevents silent assembly pollution in bundle builds.

## Summary by CodeRabbit

* **Bug Fixes**
  * Operand build resolution now verifies that builds belong to the requested assembly.
  * Prevents mismatched-assembly builds from being used during OLM bundle rebasing.
  * Applies assembly validation to both standard and embargoed build lookups.

### DIFF
--- a/doozer/doozerlib/backend/konflux_olm_bundler.py
+++ b/doozer/doozerlib/backend/konflux_olm_bundler.py
@@ -681,6 +681,12 @@ class KonfluxOlmBundleRebaser:
                         f"embargoed, and no public (non-embargoed) build of '{meta.distgit_key}' "
                         f"is available to substitute. Cannot safely rebase this operator bundle."
                     )
+                if public_build.assembly != self.assembly:
+                    raise KonfluxOlmBundleRebaseError(
+                        f"Assembly mismatch: public substitute operand {public_build.nvr} for operator "
+                        f"{metadata.distgit_key} has assembly '{public_build.assembly}', "
+                        f"expected '{self.assembly}'."
+                    )
                 logger.warning(
                     "Operand %s referenced by operator %s is embargoed; substituting with public build %s",
                     build.nvr,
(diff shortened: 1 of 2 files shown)


doozer/doozerlib/backend/konflux_olm_bundler.py
rayfordj, line 690: leaving `/hold` for comment consideration; remove to proceed as-is.

Should we have a [similar](https://github.com/openshift-eng/art-tools/pull/3256/changes#diff-af07b11f87daf487432e8fc37e3efcc3c71c59d2445510bf1d9155991a5979ebR654) `if public_build.assembly != self.assembly:` here?  Or is it because we are already passing in `assembly=self.assembly` it should always be correct assembly?  Is there any way this might produce a wrong assembly result? 
🤔🥴 

I'm thinking "safe" [given](https://github.com/openshift-eng/art-tools/blob/d3ea445ab351dc281444d18a746aa7086378f903/artcommon/artcommonlib/metadata.py#L457-L459)
```
:param assembly: A non-default assembly name to search relative to. If not specified, runtime.assembly
                 will be used. If runtime.assembly is also None, the search will return true latest.
                 If the assembly parameter is set to '', this search will also return true latest.
``` 


/lgtm
/approve

fgallott: let me add a check, better be safe here :)

rayfordj (approved):
/unhold
/lgtm
/approve

PR #3266, opened 2026-08-11
fix(okd): stop overwriting scos-5.0-art stream-coreos with 4.22 image

fgallott:
see https://github.com/openshift/release/blob/main/ci-operator/config/openshift/os/openshift-os-master__okd-scos.yaml#L54-L55

[thread](https://redhat-internal.slack.com/archives/CB95J6R4N/p1786427518087249)

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated CoreOS image mirroring to run only for version 4.23.
  * Version 4.23 now mirrors both CoreOS image tags from version 5.0.
  * Versions 4.21, 4.22, 5.0, and other versions no longer perform this mirroring.
  * Preserved support for custom namespaces, dry runs, and successful or failed mirroring outcomes.


locriandev (approved):
/lgtm

PR #3269, opened 2026-08-11
ART-21515: detect RHEL version from image path, not just tag

fgallott:
When a builder stage uses a generic tag like "latest" (e.g. ubi9/ubi:latest), _extract_rhel_version_from_pullspec only searched the tag and returned None, causing _has_rhel_version_mismatch to fail-open and allow el8 repos to be used against a UBI9 platform context — crashing DNF module resolution with "nothing provides module(platform:el8)" errors.

Fix: search the full image path (before digest) so image names like ubi9/ubi or rhel8/buildah are matched even when the tag is generic.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of RHEL-family versions in image references, including versions specified in image paths and digest-only references.
  * Preserved fallback support for version identifiers included in image tags.

### DIFF
--- a/doozer/doozerlib/lockfile_prototype/generator.py
+++ b/doozer/doozerlib/lockfile_prototype/generator.py
@@ -512,12 +512,20 @@ class RpmLockfilePrototypeGenerator:
             int | None: RHEL major version (e.g. 8, 9), or None if
                 not detectable.
         """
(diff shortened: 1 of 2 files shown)


rayfordj (approved):
/unhold
/lgtm
/approve

PR #3275, opened 2026-08-12
ART-21515: restore upgrade-over-reinstall for base image packages

fgallott:
[ART-21309](https://redhat.atlassian.net/browse/ART-21309) rewrite (612950052) lost the fix from PR #3126 (055086c65) that kept base image packages out of reinstallPackages. Reinstalling pins to the installed EVR and overrides DNF's upgrade intent — e.g. python3-setuptools 53.0.0 from the e4s baseos repo silently wins over 67.6.1 in the ironic plashet, causing prepare-image.sh's >= 64.0.0 requirement to fail at build time.

Pass base image packages as upgradePackages instead of reinstallPackages for the final stage so repos can provide newer versions. Also collapses the duplicate bare-update branch (final stage already populates upgrade_pkgs) to avoid a redundant _get_base_image_packages call.

Adds a regression test that fails with the old reinstall behavior.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved final-stage package updates so newer repository versions can replace packages inherited from the base image.
  * Ensured bare updates consistently use the correct package targets across build stages.

* **Tests**
  * Added regression coverage for selecting newer repository versions over installed base-image packages.

### DIFF
--- a/doozer/doozerlib/lockfile_prototype/generator.py
+++ b/doozer/doozerlib/lockfile_prototype/generator.py
@@ -387,14 +387,21 @@ class RpmLockfilePrototypeGenerator:
             extra_packages: list[str] = list(cat_packages.get(stage_num, []))
 
             reinstall_pkgs: list[str] | None = None
+            upgrade_pkgs: list[str] | None = None
             if stage_num == final_stage_num:
                 if image_pullspec:
                     base_pkgs = await self._get_base_image_packages(stage_num, image_pullspec, distgit_key)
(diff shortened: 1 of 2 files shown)


fgallott:
[test](https://konflux-ui.apps.kflux-ocp-p01.7ayg.p1.openshiftapps.com/ns/ocp-art-tenant/applications/openshift-4-18/pipelineruns/ose-4-18-ironic-999tt)

PR #3304, opened 2026-08-17
ART-21958: Rebuild and sync CI golang-builder images when their golang parent changes

kopero2000:
## Summary
- Add a standing reconciliation check in `update-golang` (`_reconcile_ci_golang_builder_images`) that detects when a `ci-openshift-golang-builder-*` CI image is stale relative to its golang parent's latest Konflux build, and rebuilds it directly (bypassing streams.yml/PRs) via `beta:images:konflux:rebase`/`build`.
- After a successful rebuild, immediately mirror the new image to CI (`_sync_ci_golang_builder_images`, using the same `images:streams mirror` mechanism and credential env vars as the scheduled sync-ci-images job) instead of waiting for that job's next run to pick it up.
- Add an ocp-build-data-validator check requiring `GO_LATEST`/`GO_EXTRA`/`GO_PREVIOUS` group.yml vars to resolve to distinct major.minor versions, so the reconciliation logic can assume at most one matching variant per golang build instead of checking every variant that could share a version.

## Test plan
- [x] `uv run pytest pyartcd/tests/pipelines/test_update_golang.py` (116 passed)
- [x] `uv run pytest ocp-build-data-validator/tests/test_schema/test_group_schema.py` (13 passed)
- [x] `make lint`

## Summary by CodeRabbit

* **New Features**
  * Added validation for Go version variables, including format checks and duplicate major/minor version detection.
  * Added automatic reconciliation of stale CI Go builder and build-root images.
  * Test assemblies now synchronize rebuilt CI images using test-specific image tags.
  * Added shipment warnings and per-platform pullspec details to build-data updates.

* **Bug Fixes**
  * Prevented invalid or conflicting Go version configurations from passing validation.

* **Tests**
  * Added coverage for Go version validation, image staleness detection, rebuilding, synchronization, and failure handling.

### DIFF
--- a/doozer/doozerlib/cli/images_streams.py
+++ b/doozer/doozerlib/cli/images_streams.py
@@ -386,6 +386,8 @@ def images_streams_mirror(
             # upstream_image to all the upstream_image_mirror destinations so they all get the same version.
             if config.upstream_image_mirror is not Missing:
                 for upstream_image_mirror_dest in config.upstream_image_mirror:
+                    if live_test_mode:
+                        upstream_image_mirror_dest += '.test'
                     # Mirror to each destination only if not in only-if-missing mode OR destination doesn't exist
                     if not only_if_missing or not destinations_to_check.get(upstream_image_mirror_dest, False):
                         priv_cmd = f'oc image mirror {config.upstream_image}'
--- a/pyartcd/pyartcd/jenkins.py
+++ b/pyartcd/pyartcd/jenkins.py
@@ -53,7 +53,7 @@ class Jobs(Enum):
     BUILD_CONFORMA_VERIFY = 'aos-cd-builds/build%2Fbuild-conforma-verify'
     SCAN_OPERATOR = 'aos-cd-builds/build%2Fscan-operator'
     # TODO TEST
-    SYNC_CI_IMAGES = 'hack/bvizi/add-ci-config-to-golang-builder'
+    SYNC_CI_IMAGES = 'hack/bvizi/add-load-disabled'
     OPEN_RECONCILIATION_PRS = 'aos-cd-builds/build%2Fopen-reconciliation-prs'
     OPEN_RECONCILIATION_PRS_LAYERED = 'aos-cd-builds/build%2Fopen-reconciliation-prs-layered-products'
 
--- a/pyartcd/pyartcd/jenkins.py
+++ b/pyartcd/pyartcd/jenkins.py
@@ -52,7 +52,8 @@ class Jobs(Enum):
     SCAN_PLASHET_RPMS = 'scanning/scanning%2Fplashet-rpms'
     BUILD_CONFORMA_VERIFY = 'aos-cd-builds/build%2Fbuild-conforma-verify'
     SCAN_OPERATOR = 'aos-cd-builds/build%2Fscan-operator'
-    SYNC_CI_IMAGES = 'aos-cd-builds/build%2Fsync-ci-images'
+    # TODO TEST
+    SYNC_CI_IMAGES = 'hack/bvizi/add-ci-config-to-golang-builder'
     OPEN_RECONCILIATION_PRS = 'aos-cd-builds/build%2Fopen-reconciliation-prs'
     OPEN_RECONCILIATION_PRS_LAYERED = 'aos-cd-builds/build%2Fopen-reconciliation-prs-layered-products'
 
@@ -531,10 +532,25 @@ def start_rhcos(build_version: str, new_build: bool, job_name: str = 'build', **
     )
 
 
-def start_sync_ci_images(version: str, **kwargs) -> Optional[str]:
+def start_sync_ci_images(
+    version: str,
+    assembly: str = 'stream',
+    image_list: list = None,
+    dry_run: bool = False,
+    load_disabled: bool = False,
+    live_test_mode: bool = False,
+    **kwargs,
+) -> Optional[str]:
     params = {
         'VERSION': version,
     }
+    if image_list:
+        params['IMAGE_LIST'] = ','.join(image_list)
+    params['ART_TOOLS_COMMIT'] = 'kopero2000@ci-golang-builder-from-update-golang-ART-21958'
+    params['DRY_RUN'] = dry_run
+    params['LOAD_DISABLED'] = load_disabled
+    params['LIVE_TEST_MODE'] = live_test_mode
+    params['ASSEMBLY'] = assembly
     return start_build(
         job=Jobs.SYNC_CI_IMAGES,
         params=params,
--- a/pyartcd/pyartcd/jenkins.py
+++ b/pyartcd/pyartcd/jenkins.py
@@ -545,7 +545,7 @@ def start_sync_ci_images(
         'VERSION': version,
     }
     if image_list:
-        params['IMAGE_LIST'] = ','.join(image_list)
+        params['IMAGES'] = ','.join(image_list)
     params['ART_TOOLS_COMMIT'] = 'kopero2000@ci-golang-builder-from-update-golang-ART-21958'
     params['DRY_RUN'] = dry_run
     params['LOAD_DISABLED'] = load_disabled
--- a/pyartcd/pyartcd/pipelines/sync_ci_images.py
+++ b/pyartcd/pyartcd/pipelines/sync_ci_images.py
@@ -57,6 +57,8 @@ class SyncCIImagesPipeline:
         skip_waits: bool = False,
         force_run: bool = False,
         update_images_only_when_missing: bool = False,
+        load_disabled: bool = False,
+        live_test_mode: bool = False,
     ) -> None:
         """
         Initialize sync-ci-images pipeline.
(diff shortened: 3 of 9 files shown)


pyartcd/pyartcd/pipelines/update_golang.py
```diff
@@ -45,6 +51,7 @@
     "Brew builds for the test assembly are not supported because Brew floating tags are updated after every "
     "successful build. Use --build-system konflux for test assembly builds."
 )
+CI_GOLANG_BUILDER_IMAGE_PREFIX = "ci-openshift-golang-builder-"
```
thegreyd, line 44: Here we want to target the buildroot images as well so you can add "ci-openshift-build-root-"

pyartcd/pyartcd/pipelines/update_golang.py
```diff
@@ -1170,6 +1181,286 @@ async def _rebase_and_build_konflux(self, el_v, go_version, go_nvr: str):
...
+            if content.name.startswith(CI_GOLANG_BUILDER_IMAGE_PREFIX) and content.name.endswith(".yml")
+        )
+        return image_keys
+
+    async def _find_latest_ci_golang_builder_build(self, image_key: str, el_v: int) -> KonfluxBuildRecord | None:
+        """Look up the latest successful Konflux build record for a CI golang builder image."""
+        return await anext(
+            self.konflux_db.search_builds_by_fields(
+                where={
+                    "name": image_key,
```
thegreyd: here we can have assembly=stream as well?

pyartcd/pyartcd/pipelines/update_golang.py
```diff
@@ -1170,6 +1181,286 @@ async def _rebase_and_build_konflux(self, el_v, go_version, go_nvr: str):
...
+            f"--working-dir={self._doozer_working_dir}-ci-{image_key}",
+            "--build-system=konflux",
+        ]
+        if self.data_path:
+            cmd.append(f"--data-path={self.data_path}")
+        cmd.extend(
+            [
+                "--group",
+                group,
+                "-i",
```
thegreyd: be sure to include assembly=stream in every doozer command - or i don't think it's a bad idea to do test assembly builds also. and then skip the sync for test assembly

pyartcd/pyartcd/pipelines/update_golang.py
kopero2000: I dont want to resolve it yet, but I put there the feature

pyartcd/pyartcd/pipelines/update_golang.py
kopero2000: I think we should put assembly = self.assembly in case of test, shouldn't we?

thegreyd: yes that sounds good

pyartcd/pyartcd/pipelines/update_golang.py
thegreyd: self.assembly as well

pyartcd/pyartcd/pipelines/update_golang.py
```diff
@@ -1335,13 +1357,13 @@ async def _sync_ci_golang_builder_images(self, image_keys: List[str]):
...
 
-    CI_GOLANG_BUILDER_VARIANT_BY_GROUP_VAR = {
+    CI_VARIANT_BY_GROUP_VAR = {
         "GO_LATEST": "latest",
         "GO_EXTRA": "extra",
         "GO_PREVIOUS": "previous",
     }
 
-    async def _reconcile_ci_golang_builder_images(
+    async def _reconcile_ci_images(
```
thegreyd: can we use another word for "reconcile" here. the reason is we have a process / job called reconciliation PRs job - so it can be confusing with that whole concept/flow.

pyartcd/pyartcd/pipelines/update_golang.py
```diff
@@ -1170,6 +1180,332 @@ async def _rebase_and_build_konflux(self, el_v, go_version, go_nvr: str):
...
+                "No CI golang builder images found for variant %s on openshift-%s; "
+                "skipping CI golang builder/build-root reconciliation",
+                variant,
+                self.ocp_version,
+            )
+            return
+
+        builder_records = await self.get_existing_builders_konflux(
+            {el_v: el_nvr_map_for_images[el_v] for el_v, _ in builder_targets}, go_version
+        )
```
thegreyd: try to reuse this code from scan-sources https://github.com/openshift-eng/art-tools/blob/ff9c3bd603d9beef9f119e820476c87ee8098b0b/doozer/doozerlib/cli/scan_sources_konflux.py#L1020

pyartcd/pyartcd/pipelines/update_golang.py
```diff
@@ -1163,6 +1177,332 @@ async def _rebase_and_build_konflux(self, el_v, go_version, go_nvr: str):
...
+            _LOGGER.info(
+                "All CI golang builder/build-root images for openshift-%s already use their current parent image;",
+                self.ocp_version,
+            )
+
+        # Sync every image considered this run, not just what was just rebuilt -- mirroring an
+        # already-current image is a no-op cost-wise (a handful of images at most), and it keeps CI
+        # in sync with the latest successful build even on runs where nothing needed rebuilding.
+        try:
+            await self._sync_ci_images(scan_keys)
```
kopero2000: use jenkins job instead

pyartcd/pyartcd/pipelines/update_golang.py
```diff
@@ -1163,6 +1177,332 @@ async def _rebase_and_build_konflux(self, el_v, go_version, go_nvr: str):
...
+        scan_keys = [image_key for _, image_key in scan_targets]
+        stale_image_keys = set(await self._scan_stale_ci_images(scan_keys))
+        rebuilt_image_keys = [image_key for image_key in scan_keys if image_key in stale_image_keys]
+
+        if rebuilt_image_keys:
+            await self._slack_client.say_in_thread(
+                f":construction: Rebuilding CI golang builder/build-root image(s) for "
+                f"{self.ocp_version}: {', '.join(rebuilt_image_keys)}"
+            )
+            await self._rebase_and_build_ci_images(rebuilt_image_keys)
```
kopero2000: use ocp4-konflux jenkins job

pyartcd/pyartcd/jenkins.py
```diff
@@ -53,7 +53,8 @@ class Jobs(Enum):
     SCAN_PLASHET_RPMS = 'scanning/scanning%2Fplashet-rpms'
     BUILD_CONFORMA_VERIFY = 'aos-cd-builds/build%2Fbuild-conforma-verify'
     SCAN_OPERATOR = 'aos-cd-builds/build%2Fscan-operator'
-    SYNC_CI_IMAGES = 'aos-cd-builds/build%2Fsync-ci-images'
+    # TODO TEST
+    SYNC_CI_IMAGES = 'hack/bvizi/add-load-disabled'
```
thegreyd: This looks like it’s still pointing at the test Jenkins job and feature branch. We’ll want to restore the production routing before merging, since scheduled syncs use this helper too.

pyartcd/pyartcd/pipelines/update_golang.py
```diff
@@ -1285,6 +1296,203 @@ async def _rebase_and_build_konflux(self, el_v, go_version, go_nvr: str):
...
+            _LOGGER.info(
+                "All CI golang builder/build-root images for openshift-%s already use their current parent image;",
+                self.ocp_version,
+            )
+
+        # Sync every image considered this run, not just what was just rebuilt -- mirroring an
+        # already-current image is a no-op cost-wise (a handful of images at most), and it keeps CI
+        # in sync with the latest successful build even on runs where nothing needed rebuilding.
+
+        sync_result = jenkins.start_sync_ci_images(
```
thegreyd: Maybe we need to force the sync here. A rebuilt image won’t necessarily change the build-data SHA, so the job could skip mirroring while we still report success.

doozer/doozerlib/cli/images_streams.py
```diff
@@ -442,6 +442,8 @@ def mirror_image(cmd_start: str, upstream_dest: str):
             # upstream_image to all the upstream_image_mirror destinations so they all get the same version.
             if config.upstream_image_mirror is not Missing:
                 for upstream_image_mirror_dest in config.upstream_image_mirror:
+                    if live_test_mode:
+                        upstream_image_mirror_dest += '.test'
```
thegreyd: Could this source tag need the `.test` suffix as well? It looks like we’re copying the production public image into the private test tag, rather than the test image we just mirrored.

pyartcd/pyartcd/pipelines/sync_ci_images.py
```diff
@@ -547,6 +552,8 @@ async def _mirror_images_to_ci(self, doozer_opts: str, auth_file: str) -> None:
             mirror_args += "--only-if-missing "
         if self.runtime.dry_run:
             mirror_args += "--dry-run"
+        if self.live_test_mode:
+            mirror_args += "--live-test-mode"
```
thegreyd: These seem to run together as `--dry-run--live-test-mode` when both options are enabled. Appending them as separate arguments would avoid the parsing error.

PR #3313, opened 2026-08-19
Add referenced-image signing to sign_existing_releases hack

thegreyd:
Tested at https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/hack/job/sidsharm-aos-cd-jobs/job/build%252Fsign-existing-release/4/console

Extends `pyartcd/hack/sign_existing_releases.py` to optionally sign the component images referenced by a release payload, not just the release images.

The signing library (`SigstoreSignatory`) already supports this — the `sigstore-sign` pipeline uses it — this just wires it into the hack script.

Add `--sign-release yes|no|only`: both (default) / release images only / components only.

## Summary by CodeRabbit

* **New Features**
  * Added options to sign release images, referenced component images, or both.
  * Component images are signed by digest, while release images remain tag-only by default.
  * Added an option to enable digest signatures for release images.
  * Added support for expanding multi-architecture component images and avoiding duplicate signing.
* **Bug Fixes**
  * Improved signing reliability with per-image status tracking.
  * Commands now return a nonzero exit code when signing fails.
* **Documentation**
  * Updated command-line help and examples for the available signing modes.

### DIFF
--- a/pyartcd/hack/sign_existing_releases.py
+++ b/pyartcd/hack/sign_existing_releases.py
@@ -6,33 +6,24 @@ images they reference) with Sigstore/cosign.
 This tool can sign release payloads (manifest lists or single manifests) that
 already exist in quay.io. It reuses the SigstoreSignatory class for signing logic.
 
-For release images, only TAG-BASED signatures are created by default (digest
-signatures are skipped). This is appropriate for retroactive signing where digest
-signatures already exist. Use --sign-digest to also create digest signatures.
+By default, only TAG-BASED signatures are created (digest signatures are skipped).
+This is appropriate for retroactive signing where digest signatures already exist.
 
-Use --sign-release to control what gets signed:
-    yes  (default) sign the release image(s) and the components they reference
-    only          sign only the release image(s)
-    no            sign only the referenced component images
-
-Referenced component images are discovered by spidering each release payload with
-`oc adm release info -o json` and are always signed with digest identity only.
-NOTE: `oc adm release info` on a `-multi` pullspec only returns one arch's
-references, so to sign all referenced images across every architecture, pass the
-per-arch release pullspecs (e.g. via --file), not just the `-multi` pullspec.
+Use --sign-release to control what is signed: "yes" (default) signs the release
+image(s) and the component images they reference, "only" signs just the release
+image(s), and "no" signs only the referenced component images. Referenced component
+images are discovered with `oc adm release info -o json` and are always signed with
+digest identity only (multi-arch references are expanded to and signed for every
+architecture).
 
 Usage:
     # Sign a single release image and its referenced components (from art-tools directory)
     uv run pyartcd/hack/sign_existing_releases.py --dry-run \
         quay.io/openshift-release-dev/ocp-release:4.16.1-multi
 
-    # Sign only the release images (skip components) from a file
+    # Sign multiple release images (only, no components) from a file
     uv run pyartcd/hack/sign_existing_releases.py --dry-run --sign-release only \
         --file pullspecs.txt
-
-    # Sign only the referenced component images
-    uv run pyartcd/hack/sign_existing_releases.py --dry-run --sign-release no \
-        quay.io/openshift-release-dev/ocp-release:4.16.1-x86_64
 
     # Real signing (requires KMS credentials)
     KMS_CRED_FILE=/path/to/creds KMS_KEY_ID=key-id REKOR_URL=https://... \
@@ -49,10 +40,10 @@ import asyncio
 import logging
 import os
 import sys
-from typing import Dict, List, Optional, Set
+from typing import List, Optional, Set
 
 import click
-from pyartcd.signatory import ReleaseImageInfo, SigstoreSignatory
+from pyartcd.signatory import SigstoreSignatory
 
 # Configure logging
 logging.basicConfig(level=logging.INFO, format='%(asctime)s - %(levelname)s - %(message)s', datefmt='%Y-%m-%d %H:%M:%S')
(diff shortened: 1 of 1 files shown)


pyartcd/hack/sign_existing_releases.py
```diff
@@ -1,22 +1,39 @@
...
-By default, only TAG-BASED signatures are created (digest signatures are skipped).
-This is appropriate for retroactive signing where digest signatures already exist.
+For release images, only TAG-BASED signatures are created by default (digest
+signatures are skipped). This is appropriate for retroactive signing where digest
+signatures already exist. Use --sign-digest to also create digest signatures.
+
+Use --sign-release to control what gets signed:
+    yes  (default) sign the release image(s) and the components they reference
+    only          sign only the release image(s)
+    no            sign only the referenced component images
```
wking: This seems workable, although personally `--sign-release no` seems a bit awkward as a way to say "sign only the referenced component images.  Maybe pivot to `--sign (release|release-image|component-images)` or some such that avoids going boolean-ish?  Or just leave it as you have it, because I expect folks to run this command very rarely, so there's not much value in polishing its interface.

pyartcd/hack/sign_existing_releases.py
```diff
@@ -1,22 +1,39 @@
...
+
+Referenced component images are discovered by spidering each release payload with
+`oc adm release info -o json` and are always signed with digest identity only.
+NOTE: `oc adm release info` on a `-multi` pullspec only returns one arch's
+references, so to sign all referenced images across every architecture, pass the
```
wking: This isn't true, multi referenced images are themselves are themselves manifest lists.  So pick one arch, get the referenced manifest lists, and then head out to the single-arch shards to sign.

```console
$ oc adm release info -o json quay.io/openshift-release-dev/ocp-release:4.20.21-multi | jq -r '.references.spec.tags[] | .name + " " + .from.name' | head -n3
agent-installer-api-server quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:538d13386a5849c0316b2c8b81cd8d926d9905089a9d772235dd8e53d1cc4e3e
agent-installer-csr-approver quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:c30f11f320c06be1fc5e257ce3220c24ff5686fbdd041cd730074b9e6a01b9cb
agent-installer-node-agent quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:e10014a7a9f8e83f37fbafe2125fa4ac76fd9c58a58954bddcf8017622a901eb
$ oc image info quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:538d13386a5849c0316b2c8b81cd8d926d9905089a9d772235dd8e53d1cc4e3e
error: the image is a manifest list and contains multiple images - use --filter-by-os to select from:

  OS            DIGEST
  linux/amd64   sha256:b9f2059776e64d8f25f041ac8742c4610267ae68d6c302250eb7ca1a81919209
  linux/arm64   sha256:2ec9ebe0000be90e7aa4f02ea47ef618e8779549a1ddb72ad7e76e6392d35dcb
  linux/s390x   sha256:1289d310ea5ffc72752e588feb677ec42630422f4ce785b8d34d91ae02ed4cb3
  linux/ppc64le sha256:870d49e712a92e0a6873686c23164f16ad983d8f79d9b9b918c639272230ac44

```

pyartcd/hack/sign_existing_releases.py
```diff
@@ -166,37 +132,96 @@ async def main_async(
...
+    if component_images:
+        logger.info("Signing %d component image(s) [digest only]", len(component_images))
+        if errors := await signatory.sign_component_images(component_images):
+            for ps, err in errors.items():
+                logger.error("Component image signing error for %s: %s", ps, err)
+            return 1
 
     # Summary
     logger.info("=" * 60)
-    logger.info("Signing complete: %d successful, %d errors", success_count, error_count)
```
wking, line 197: Why drop the `error_count`?  I don't see motivation for that change discussed in the commit message.

pyartcd/hack/sign_existing_releases.py
```diff
@@ -166,37 +132,96 @@ async def main_async(
...
+                release_name="",  # Skip release name validation
+                verify_legacy_sig=False,
+            )
+            release_images.append(release_info)
+            all_errors.update(errors)
+
+    # --- Phase 2: Discover referenced component images from each payload ---
+    component_images: Set[str] = set()
+    if do_sign_components:
+        for pullspec in cleaned:
```
wking: This all-discovery-first approach surprises me, although it can clearly work.  It might be easier to think about if we keep `sign_release_pullspec` (and generalize the name to `sign_release`?) and pass through a mutable [set](https://docs.python.org/3/tutorial/datastructures.html#sets) of already-signed-this-round referenced images.  Then that per-release function can get that release all signed up, without needing to wait on discovery having walked all the _other_ releases that we were planning to sign.  And if walking a later release turned up a referenced image we'd already signed when processing an earlier release, we'd see the entry in the shared, mutable set, and realize we didn't need to double-up on the signature.

pyartcd/hack/sign_existing_releases.py
wking: Pinning this down more exactly, by explicitly comparing two shards of that release image:

```console
$ oc image info quay.io/openshift-release-dev/ocp-release:4.20.21-multi 
error: the image is a manifest list and contains multiple images - use --filter-by-os to select from:

  OS            DIGEST
  linux/amd64   sha256:5a55ef5c98fa4e0bfea201652cb8e779285202dbc36f1723f8553d294fe852a7
  linux/arm64   sha256:5debfb941fbad45596b5665529d77377a8cb0d9a103be0a3c11ddf652ef557fe
  linux/s390x   sha256:bfd0f80bdfe9f6387f8092b132ac96e1afcc06240bb0aacd366ed9e8b2e437fd
  linux/ppc64le sha256:9fb326aacc7daa7043c2be9f6b2048eaddedf266e6927525ad9a6378621b9d74

$ diff -u1 <(oc adm release info -o json quay.io/openshift-release-dev/ocp-release@sha256:5a55ef5c98fa4e0bfea201652cb8e779285202dbc36f1723f8553d294fe852a7 | jq -r '.references.spec.tags[] | .name + " " + .from.name') <(oc adm release info -o json quay.io/openshift-release-dev/ocp-release@sha256:5debfb941fbad45596b5665529d77377a8cb0d9a103be0a3c11ddf652ef557fe | jq -r '.references.spec.tags[] | .name + " " + .from.name')
...no difference...
```

pyartcd/hack/sign_existing_releases.py
thegreyd: Yeah, I agree it is a bit awkward. This being the existing choice in sigstore pipeline makes it a familiar pattern - https://github.com/openshift-eng/art-tools/blob/ea7f4aebc9c053ec34776f9f3708b584f6ff683a/pyartcd/pyartcd/pipelines/sigstore_sign.py#L218 so I'll leave it as is for now.

pyartcd/hack/sign_existing_releases.py
thegreyd: Thanks for catching this!

pyartcd/hack/sign_existing_releases.py
thegreyd: Yeah that was an overreach, restored now

pyartcd/hack/sign_existing_releases.py
thegreyd: `signatory.discover_component_images` is an established SigstoreSignatory pattern so I felt comfortable using it

PR #3328, opened 2026-08-27
fix(konflux-db): pass group from where clause to get_latest_build in get_build_records_by_nvrs

lgarciaaco:
## Summary

- `get_build_records_by_nvrs` was silently ignoring the `group` key in the `where` parameter, causing `get_latest_build` to fall back to `extract_group_from_nvr()` for NVR-only lookups
- For golang builder NVRs, this extracts the old per-version group (`rhel-9-golang-1.26`) instead of the unified `golang` group used since the branch migration
- `GolangBuilderShipmentHandler._create_snapshot_via_elliott` calls `elliott snapshot new --group golang`, which passes `where={"group": "golang"}`, but the DB query was searching under `rhel-9-golang-1.26` and returning 0 results

## Fix

Extract `group` from the `where` param alongside `assembly`, `el_target`, etc., and forward it to `get_latest_build`.

## Test plan

- [ ] Verify `elliott snapshot new --group golang --nvr openshift-golang-builder-container-v1.26.5-...` finds the build record
- [ ] Confirm `base-image-release` Jenkins job with `BUILD_VERSION=golang SKIP_LEGACY_RELEASE=true` succeeds end-to-end

## Summary by CodeRabbit

* **Bug Fixes**
  * Build record searches now correctly apply the optional group filter when retrieving the latest build information.

* **Changes**
  * The Golang update command no longer supports the `--skip-pr` option.
  * Golang stream updates no longer create or update build-data branches, files, or pull requests.
  * Pinned-NVR updates are skipped after stream processing.

### DIFF
--- a/artcommon/artcommonlib/konflux/konflux_db.py
+++ b/artcommon/artcommonlib/konflux/konflux_db.py
@@ -1354,6 +1354,7 @@ class KonfluxDb:
                 )
 
         # Extract additional filters from where if provided
+        group = where.get('group') if where else None
         assembly = where.get('assembly') if where else None
         el_target = where.get('el_target') if where else None
         artifact_type = where.get('artifact_type') if where else None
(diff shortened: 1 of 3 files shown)


lgarciaaco:
Closing: this PR mixed two unrelated changes (the konflux_db.py group fix and the update_golang PR creation removal). Reopening as separate focused PRs.

PR #3339, opened 2026-09-01
Resolve dnf --installroot packages with --bare in rpm-lockfile-prototype

fbladilo:
## Summary
- `dnf --installroot` installs into an empty root, not the stage image. Resolving those names with `--image` (e.g. golang-builder) saw them as already installed, produced an empty RPM lockfile, and broke hermetic Konflux (no yum repos).
- Split resolve: installroot package names go through `--bare`; stages whose only RPM work is `--installroot` skip the `--image`/pin path entirely (CLO builder). Mixed stages keep `--image` and exclude those names from pin.
- Normal `dnf install` into the stage image is unchanged.

## Failed job
Konflux pipeline run `logging-6-7-cluster-logging-operator-lwx4l` (`art-logging-tenant` / `logging-6-7`). Hermetic `dnf --installroot` failed with no yum repos after an empty RPM lockfile.

## Test plan
- [x] `uv run pytest --verbose --color=yes doozer/tests/lockfile_prototype/test_generator.py` (89 passed)
- [ ] Re-run `build/layered-products` for logging-6.7 `cluster-logging-operator` with `ART_TOOLS_COMMIT` pointing at this branch
- [ ] Confirm no pin-pass `rsync` warnings; `rpms.lock.yaml` contains `rsync`/`xz`/deps; Hermeto RPM prefetch downloads; hermetic `dnf --installroot` sees a yum repo

### DIFF
--- a/doozer/doozerlib/lockfile_prototype/generator.py
+++ b/doozer/doozerlib/lockfile_prototype/generator.py
@@ -58,14 +58,44 @@ _BARE_UPDATE_RE = re.compile(
     r"\b(?:microdnf|dnf|yum)\s+(?:-y\s+)?(?:update|upgrade)(?:\s+-y)?\s*(?:\\\n\s*&&\s*|&&\s*|;\s*|\n|(?=$))"
 )
 
+# Flags between the manager and ``install`` must not cross ``&&`` / ``;``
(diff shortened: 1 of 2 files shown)


fbladilo:
Final solution here : https://github.com/openshift-eng/art-tools/pull/3343

PR #3357, opened 2026-09-07
ART-23298: fix(go:report): handle floating golang builder tags in streams.yml

lgarciaaco:
## Summary

`elliott go:report` and `find-bugs:golang` crash when a golang builder stream in `streams.yml` uses a floating tag (e.g. `v1.22-rhel9`) instead of a full NVR-style tag. `parse_nvr()` needs an `X.Y.Z` version and `split_el_suffix_in_release` needs `.elN` — both fail on `-rhelN` floating tags. This blocks the `streams.yml` migration to floating tags ([ART-23148](https://redhat.atlassian.net/browse/ART-23148)).

## Changes

**`elliott/elliottlib/cli/get_golang_report_cli.py`**
- Add `is_floating_golang_builder_tag()` to detect `vX.Y-rhelN` tags before the existing NVR code paths
- `go_version_from_floating_tag()` — non-exact mode: extracts `X.Y.elN` (or `X.Y`) from the tag string directly, no network call
- `go_version_from_floating_tag_exact()` — exact mode: calls `oc image info` to resolve the floating tag to an actual build, reads golang package NVR from OCI labels, delegates to `get_golang_container_nvrs`
- All three existing NVR tag formats are unaffected (legacy, registry NVR-style, quay Konflux)
- `find-bugs:golang` inherits the fix through the shared `golang_report_for_version` function
- Add defensive `len==1` assertions before indexing `go_builder_nvr_map` in both floating-tag and NVR exact-mode paths

**`elliott/tests/test_get_golang_report_cli.py`** (new file)
- `(a)` floating-tag detection helper — True for floating, False for all three NVR formats
- `(b)` non-exact extraction — correct `X.Y.elN` from floating tag
- `(c)` exact-mode resolution via mocked `oc image info`
- `(d)` regression tests for all three legacy NVR formats
- `(e)` `golang_report_for_version(exact=True)` with a floating-tag stream ([SC-5](https://redhat.atlassian.net/browse/SC-5) / find-bugs path)

## Test plan

```
PYTHONPATH=artcommon:elliott python3 -m pytest elliott/tests/test_get_golang_report_cli.py -v
# 18 passed
```

Full suite: 758 passed, 0 failed (verified by reviewer).

## Related

- Fixes [ART-23298](https://redhat.atlassian.net/browse/ART-23298)
- Part of [ART-23167](https://redhat.atlassian.net/browse/ART-23167) floating tag audit
- Unblocks [ART-23148](https://redhat.atlassian.net/browse/ART-23148) streams.yml migration

## Summary by CodeRabbit

* **New Features**
  * Golang reports now support floating builder tags with Go-version extraction and exact-version resolution.
  * Exact lookups can resolve versions from image metadata or package information.
  * Exact resolution validates that image metadata maps to one package version.
  * Updated CLI usage examples to use `--ocp-versions`.

* **Bug Fixes**
  * Improved handling of missing, empty, null, malformed, or ambiguous image metadata.

* **Tests**
  * Added coverage for tag handling, parsing, validation, fallback resolution, and report generation.

### DIFF
--- a/elliott/elliottlib/cli/get_golang_report_cli.py
+++ b/elliott/elliottlib/cli/get_golang_report_cli.py
@@ -6,12 +6,71 @@ from artcommonlib import logutil
 from artcommonlib.format_util import green_print
 from artcommonlib.release_util import split_el_suffix_in_release
 from artcommonlib.rpm_utils import parse_nvr
+from artcommonlib.util import oc_image_info
 
 from elliottlib.cli.common import cli
 from elliottlib.runtime import Runtime
 from elliottlib.util import get_golang_container_nvrs
 
 _LOGGER = logutil.get_logger(__name__)
+
+# Matches floating golang-builder tags such as:
+#   openshift-golang-builder-container-v1.22-rhel9
+#   openshift-golang-builder-container-v1.22-rhel8
+# These lack the X.Y.Z patch version present in full NVR tags.
+_FLOATING_TAG_RE = re.compile(r'v(\d+\.\d+)-rhel(\d+)$')
+
+
+def is_floating_golang_builder_tag(nvr_like: str) -> bool:
+    """Return True when *nvr_like* is a floating tag (vX.Y-rhelN) rather than a full NVR string."""
+    return bool(_FLOATING_TAG_RE.search(nvr_like))
+
+
+def go_version_from_floating_tag(nvr_like: str, ignore_rhel: bool) -> str:
+    """Extract a go-version string from a floating tag like
+    ``openshift-golang-builder-container-v1.22-rhel9``.
+
+    Returns ``X.Y.elN`` normally, or just ``X.Y`` when *ignore_rhel* is True.
+    """
+    m = _FLOATING_TAG_RE.search(nvr_like)
+    if not m:
+        raise ValueError(f"Not a floating golang-builder tag: {nvr_like!r}")
+    major_minor = m.group(1)
+    rhel_version = m.group(2)
+    if ignore_rhel:
+        return major_minor
+    return f"{major_minor}.el{rhel_version}"
+
+
+def go_version_from_floating_tag_exact(image_pullspec: str) -> str:
+    """Resolve a floating-tag pullspec to the exact golang package NVR.
+
+    Calls ``oc image info`` to read the OCI labels from the resolved image,
+    constructs the builder NVR, then delegates to ``get_golang_container_nvrs``
+    (exact mode) to return the golang package NVR string (e.g.
+    ``golang-1.22.5-1.el9``).
+    """
+    _LOGGER.info(f"Resolving floating tag via oc image info: {image_pullspec}")
+    image_data = oc_image_info(image_pullspec, '--filter-by-os=amd64')
+    labels = image_data.get('config', {}).get('config', {}).get('Labels', {})
+    component = labels.get('com.redhat.component')
+    version = labels.get('version')
+    release = labels.get('release')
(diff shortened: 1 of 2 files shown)


elliott/elliottlib/cli/get_golang_report_cli.py
```diff
@@ -6,13 +6,72 @@
...
 from elliottlib.runtime import Runtime
 from elliottlib.util import get_golang_container_nvrs
 
 _LOGGER = logutil.get_logger(__name__)
 
+# Matches floating golang-builder tags such as:
+#   openshift-golang-builder-container-v1.22-rhel9
+#   openshift-golang-builder-container-v1.22-rhel8
+# These lack the X.Y.Z patch version present in full NVR tags.
+_FLOATING_TAG_RE = re.compile(r'v(\d+\.\d+)-rhel(\d+)$')
```
locriandev: can you add an example of string matching the regex for reference?

elliott/elliottlib/cli/get_golang_report_cli.py
```diff
@@ -6,13 +6,72 @@
...
+
+
+def is_floating_golang_builder_tag(nvr_like: str) -> bool:
```
locriandev, line 34: can you improve the docstring by providing an example `nvr_like` string?

elliott/elliottlib/cli/get_golang_report_cli.py
```diff
@@ -6,13 +6,72 @@
...
+def go_version_from_floating_tag_exact(image_pullspec: str) -> str:
+    """Resolve a floating-tag pullspec to the exact golang package NVR.
+
+    Calls ``oc image info`` to read the OCI labels from the resolved image,
+    constructs the builder NVR, then delegates to ``get_golang_container_nvrs``
+    (exact mode) to return the golang package NVR string (e.g.
+    ``golang-1.22.5-1.el9``).
+    """
+    _LOGGER.info(f"Resolving floating tag via oc image info: {image_pullspec}")
+    image_data = oc_image_info(image_pullspec, '--filter-by-os=amd64')
```
locriandev: We already have `oc_image_info_for_arch()` as a helper. There is also near-identical NVR-label extraction in doozer/doozerlib/release_inspector.py:72-101 via `extract_nvr_from_pullspec()`, we could reuse that

elliott/elliottlib/cli/get_golang_report_cli.py
```diff
@@ -6,13 +6,72 @@
...
+    if not m:
+        raise ValueError(f"Not a floating golang-builder tag: {nvr_like!r}")
+    major_minor = m.group(1)
+    rhel_version = m.group(2)
+    if ignore_rhel:
+        return major_minor
+    return f"{major_minor}.el{rhel_version}"
+
+
+def go_version_from_floating_tag_exact(image_pullspec: str) -> str:
```
locriandev, line 61: The PR reads the builder image’s own NVR labels:
- com.redhat.component
- version
- release

 Those identify the container build, not necessarily the exact golang RPM being reported. It then queries Brew/Konflux to derive the RPM NVR. However, builder images already carry the exact RPM NVR in:

```
io.openshift.build.golang-nvr=golang-1.22.12-2.el9
```
so we should read that label directly and return it. The database lookup should only be a fallback for older images without the label. This is both simpler and avoids an unnecessary external lookup.

locriandev (changes requested):

locriandev (commented):
/lgtm

PR #3358, opened 2026-09-07
ART-23300: fix(release_util): handle -rhelN suffix in split_el_suffix_in_release

lgarciaaco:
## Summary

`isolate_el_version_in_release()` was returning `None` for floating golang-builder tags like `golang-builder-v1.22-rhel9` because `split_el_suffix_in_release()` only matched `.elN`/`+elN`/`.scosN` suffixes, not the `-rhelN` form used by floating tags introduced in [ART-23148](https://redhat.atlassian.net/browse/ART-23148).

This caused the rebaser's upstream-to-stream matching (`rebaser.py:2409`) to silently fail the EL-version comparison and fall back to default stream resolution, losing the `# ART matched upstream parent` Dockerfile annotation.

## Problem

**Before:** `isolate_el_version_in_release("golang-builder-v1.22-rhel9")` returned `None`. The rebaser could not match floating-tag stream entries by RHEL version.

**After:** Returns `9` (or `8`, `10`, etc.). The rebaser correctly pairs upstream CI builder images to the right stream entry.

## Implementation Details

- Added a `-rhelN` fallback regex in `split_el_suffix_in_release()` (`artcommonlib/release_util.py`) that normalises matches to the canonical `elN` form. The primary `.elN`/`.scosN` branch is unchanged and always wins for real RPM NVRs.
- Delegated `pyartcd.util.isolate_el_version_in_release()` to the artcommonlib canonical implementation. The delegation is safe — `pyartcd/util.py` already imports from `artcommonlib.release_util`. This also extends pyartcd callers to handle `+elN`/`.scosN` patterns they previously missed.
- New tests cover `rhel8`/`rhel9`/`rhel10`, multiple `-rhelN` segments (rightmost wins due to greedy `.*`), and a malformed `-rhel` (no digits) case. All existing patterns pass.

Note: the `-rhelN` regex is a fallback that only fires when the primary `.elN`/`.scosN` branch does not match. Real RPM NVRs always carry a dist tag, so there is no false-positive risk for existing callers.

Fixes: [ART-23300](https://redhat.atlassian.net/browse/ART-23300)
Related: ART-23167, [ART-23148](https://redhat.atlassian.net/browse/ART-23148)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved release parsing for floating image tags with `-rhelN` suffixes.
  * Normalized supported suffixes to the corresponding EL version, including tags containing multiple suffixes.
  * Added consistent handling for `.elN`, `+elN`, `.scosN`, and `-rhelN` formats.
  * Improved handling of malformed suffixes that lack a numeric version.

* **Tests**
  * Expanded coverage for valid, multiple, and malformed release suffixes across supported image tag formats.

### DIFF
--- a/artcommon/artcommonlib/release_util.py
+++ b/artcommon/artcommonlib/release_util.py
@@ -22,6 +22,10 @@ def split_el_suffix_in_release(release: str) -> Tuple[str, Optional[str]]:
     is None if there .el### or .scos### is not detected.
     For OKD builds, this returns the scos suffix (e.g., 'scos9').
     For OCP builds, this returns the el suffix (e.g., 'el9').
+
+    Also handles floating image tags that use a '-rhelN' suffix (e.g.,
+    'golang-builder-v1.22-rhel9'), normalising the result to the canonical
+    'el9' form so that callers behave identically for both tag styles.
     """
 
     el_suffix_match = re.match(r'(.*)[.+]((?:el|scos)\d+(?:_\d+)?)(?:.*|$)', release)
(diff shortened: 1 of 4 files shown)


locriandev (approved):
/lgtm

PR #3359, opened 2026-09-07
ART-23299: update_golang floating-tag support in _get_builder_pullspec

lgarciaaco:
## Summary

Fix `_get_builder_pullspec()` so it detects whether the target branch's `streams.yml` uses floating tags ([ART-23148](https://redhat.atlassian.net/browse/ART-23148) migration) and emits the appropriate pullspec format. Previously it always emitted NVR-pinned pullspecs, which would overwrite floating refs on migrated branches.

## Changes

**`pyartcd/pipelines/update_golang.py`**

- Add `_parse_pullspec_tuple(pullspec)` normalises both floating-tag (`golang-builder-v1.22-rhel9`) and NVR formats to `(major, minor, rhel_version)` int-tuples
- Add `_branch_uses_floating_tags(streams_content)` returns True when any golang-builder stream entry uses the floating-tag format
- Add `_pullspecs_match(a, b)` compares pullspecs by tuple; falls back to string equality for legacy formats
- Update `_get_builder_pullspec(builder_nvr, streams_content=None)` auto-reads streams from the memoized `_branch_content`; returns floating tag for floating branches; falls back to NVR on parse failure
- Replace all five `info['image'] == ref` comparisons in `update_golang_streams()` with `_pullspecs_match()`-based equivalents covering minor-bump and major-bump paths

**`pyartcd/tests/pipelines/test_update_golang.py`**

New test classes: `TestParsePullspecTuple`, `TestPullspecsMatch`, `TestBranchUsesFloatingTags`, `TestGetBuilderPullspec`, `TestUpdateGolangStreamsFloatingTags`

## Test plan

```
uv run python -m pytest pyartcd/tests/pipelines/test_update_golang.py -v
# 137 passed, 0 failed
```

## Backwards compatibility

NVR-pinned branches: no change. `streams_content=None` default preserves single-argument callers. `_pullspecs_match` falls back to string equality for unrecognised formats.

Fixes: [ART-23299](https://redhat.atlassian.net/browse/ART-23299)

## Summary by CodeRabbit

- **Bug Fixes**
  - Pullspec comparisons now safely reject missing values and require matching repositories when comparing floating-tag and NVR formats.
  - Go image handling now reliably supports floating-tag and NVR references, including exact matching for distinct patch and build releases.
  - Missing or image-less `GO_EXTRA` streams are skipped with warnings instead of causing update failures.
  - Floating tags retain complete repository-qualified references.

- **Tests**
  - Added comprehensive coverage for pullspec parsing, image comparisons, version transitions, stream updates, and missing or malformed inputs.

### DIFF
--- a/pyartcd/pyartcd/pipelines/update_golang.py
+++ b/pyartcd/pyartcd/pipelines/update_golang.py
@@ -43,6 +43,54 @@ BREW_TEST_ASSEMBLY_UNSUPPORTED = (
 )
 
 
+# Floating-tag format: registry.redhat.io/openshift/golang-builder:golang-builder-v1.22-rhel9
+_FLOATING_TAG_RE = re.compile(r'golang-builder-v(\d+)\.(\d+)-rhel(\d+)$')
+# NVR-tag format: registry.redhat.io/openshift/golang-builder:openshift-golang-builder-container-v1.22.12-...el9
+_NVR_TAG_RE = re.compile(r'openshift-golang-builder[^:]*-v(\d+)\.(\d+)\.\d+[^:]*\.el(\d+)')
+
+
+def _parse_pullspec_tuple(pullspec: str) -> tuple[int, int, int]:
+    """Normalise a golang-builder pullspec (floating or NVR) to (major, minor, rhel_version).
+
+    Raises ValueError for unrecognised formats.
+    """
+    tag = pullspec.split(':')[-1]
+    m = _FLOATING_TAG_RE.search(tag)
+    if m:
+        return int(m.group(1)), int(m.group(2)), int(m.group(3))
+    m = _NVR_TAG_RE.search(tag)
+    if m:
+        return int(m.group(1)), int(m.group(2)), int(m.group(3))
+    raise ValueError(f"Cannot parse golang-builder pullspec into (major, minor, rhel) tuple: {pullspec!r}")
+
+
+def _branch_uses_floating_tags(streams_content: dict) -> bool:
+    """Return True when any golang-builder stream entry uses a floating tag format."""
+    if not streams_content:
+        return False
+    for info in streams_content.values():
+        image = info.get('image', '') if isinstance(info, dict) else ''
+        if _FLOATING_TAG_RE.search(image.split(':')[-1]):
+            return True
+    return False
+
+
+def _pullspecs_match(a: str, b: str) -> bool:
+    """Compare two golang-builder pullspecs via tuple normalisation.
+
+    When both strings represent recognisable golang-builder pullspecs (floating or NVR),
+    compares by (major, minor, rhel_version) so a floating tag and an NVR for the same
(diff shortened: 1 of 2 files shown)


pyartcd/pyartcd/pipelines/update_golang.py
```diff
@@ -806,14 +854,42 @@ async def find_builder(el_v: int) -> KonfluxBuildRecord | None:
...
+                streams_content = cached.get('streams')
+
+        if streams_content and _branch_uses_floating_tags(streams_content):
+            # Extract go major.minor and RHEL version from the NVR, e.g.
+            # openshift-golang-builder-container-v1.22.12-...el9 → (1, 22, 9)
+            published_nvr = f'{component_name}-{parsed_nvr["version"]}-{parsed_nvr["release"]}'
+            full_pullspec = rh_art_images_base_pullspec(published_nvr)
+            try:
+                major, minor, el_v = _parse_pullspec_tuple(full_pullspec)
+                return f"golang-builder-v{major}.{minor}-rhel{el_v}"
```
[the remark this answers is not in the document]
locriandev: @lgarciaaco this sounds legit. Here we only return `golang-builder-v1.22-rhel9`, which is written directly into streams.yml, breaking pulls of the intended builder; the new tests incorrectly assert the incomplete value.

pyartcd/pyartcd/pipelines/update_golang.py
```diff
@@ -43,6 +43,54 @@
...
+
+    When both strings represent recognisable golang-builder pullspecs (floating or NVR),
+    compares by (major, minor, rhel_version) so a floating tag and an NVR for the same
+    builder version are treated as equal.
+
+    Falls back to string equality when either pullspec cannot be parsed (e.g. unrecognised
+    registry/format in test fixtures or legacy entries), preserving the old behaviour.
+    """
+    try:
+        return _parse_pullspec_tuple(a) == _parse_pullspec_tuple(b)
```
locriandev: `_parse_pullspec_tuple()` drops the Go patch and build release, so v1.22.10-1.el9 and v1.22.11-2.el9 both become (1, 22, 9). Because this comparison selects which streams to update, bumping one stream would also  overwrite another stream intentionally pinned to a different build.

pyartcd/pyartcd/pipelines/update_golang.py
locriandev: We could reuse `doozerlib.util.konflux_golang_builder_component_name(published_nvr)` here. It already generates this floating tag and is used by the publishing code, so this introduces a second implementation of the same naming rules.

After importing the helper, these lines could become:
```
  tag = konflux_golang_builder_component_name(published_nvr)
  repository = full_pullspec.rsplit(":", 1)[0]
  return f"{repository}:{tag}"
```

locriandev (commented):
/lgtm

PR #3396, opened 2026-09-14
Avoid unsafe multi-arch RPM rebuilds

fgallott:
## What does this PR do?

  Prevent `ocp4-scan-konflux` from triggering RPM rebuilds that would update only some architectures and potentially create a Conforma `rpm_packages.unique_version` violation.

  RPM-triggered rebuilds are now allowed only when:

  - The RPM update is available for every target architecture.
  - The candidate RPM version is consistent across all architectures.

  An existing architecture mismatch does not suppress a rebuild when a consistent update is available to repair it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RPM update handling for multi-architecture images.
  * Rebuilds now occur only when matching RPM updates are available across all architectures.
  * Prevented unnecessary rebuilds when updates are available for only some architectures.
  * Prevented rebuilds when architectures resolve to conflicting candidate RPM versions.
  * Consistent updates can correct existing architecture version mismatches.
  * Single-architecture builds continue to follow their existing RPM update behavior.


rayfordj (approved):
/hold ... to consider coderabbit feedback
/lgtm
/approve

PR #3413, opened 2026-09-16
ART-21314: Store Build Variant in BigQuery for All Build Records

fgallott:
### Summary

  Store the build variant in every newly created BigQuery build record, for both Konflux and Brew builds.

  This allows build records from OCP, OKD, MicroShift, and layered products to be reliably distinguished without inferring the variant from the group or product name.

  ### Motivation

  Build records share the same BigQuery model across multiple build systems and artifact types. 

  The build variant is now taken from the active Doozer runtime (`runtime.variant` or `metadata.runtime.variant`), which is the authoritative value selected through the explicit `--variant` option.

  ### Changes

  #### Build variant model

  - Added `BuildVariant.MICROSHIFT`.
  - Kept MicroShift next to OCP in the enum because it is OCP-derived but has a distinct build variant.
  - Grouped layered-product variants separately in the enum.
  - Simplified product resolution by using the canonical enum values directly.
  - Unsupported products continue to return `None`.

  #### Explicit Doozer variants

  Relevant Doozer commands now receive an explicit `--variant` value instead of relying on the default:

  - Standard OCP builds use `--variant=ocp`.
  - Layered-product builds use the product-specific variant.
  - FBC and OLM bundle Konflux builds receive the resolved variant.
  - MicroShift RPM builds use `--variant=microshift`.
  - MicroShift bootc Konflux builds use `--variant=microshift`.
  - Existing OKD handling remains unchanged.

  #### BigQuery build records

  `build_variant` is now stored for:

  - Konflux image records.
  - Konflux bundle records.
  - Konflux FBC records.
  - Brew image records.
  - Brew RPM records.
  - Brew OLM bundle records.

  The database now rejects any newly written build record that does not include `build_variant`, regardless of whether the engine is Brew or Konflux.

  Historical BigQuery records remain unchanged and may still contain a `NULL` variant.

  ### Testing

  Added and updated coverage for:

  - Complete `BuildVariant` enum definitions.
  - Product-to-variant resolution and normalization.
  - Rejection of missing variants for both Konflux and Brew records.
  - Brew image, RPM, and OLM bundle records storing the runtime variant.
  - Explicit variant arguments in OCP, layered-product, FBC, OLM bundle, and MicroShift build commands.

---
I ran a couple of test builds with this change

Screenshot From 2026-09-16 11-55-02

## Summary by CodeRabbit

* **New Features**
  * Added support for resolving product names to build variants, including MicroShift and layered products.
  * Build pipelines now pass the appropriate variant to build commands.
  * Persisted build records and release follow-ups now retain or derive runtime variant information.

* **Bug Fixes**
  * Records without a build variant are rejected before insertion.
  * Missing product configuration defaults to the OpenShift variant.


pyartcd/pyartcd/pipelines/build_layered_products.py
```diff
@@ -189,25 +190,29 @@ def _group_param(self) -> str:
...
             f"--assembly={self.assembly}",
             f"--working-dir={self.runtime.doozer_working}",
             f"--data-path={self._doozer_env_vars['DOOZER_DATA_PATH']}",
             "--build-system=konflux",
             self._group_param(),
-            "--latest-parent-version",
         ]
+        if build_variant is not None:
+            command.append(f"--variant={build_variant.value}")
+        command.append("--latest-parent-version")
```
locriandev: only cosmetic, but why not leave line #200 intact then?

locriandev (approved):
/lgtm

PR #3435, opened 2026-09-18
Use the Go built-in FIPS module for golang 1.26 and later

zaneb:
Go 1.26 ships a built-in, FIPS 140-3 validated cryptographic module, selected with GOFIPS140. It is pure Go, so the workarounds the shim needs in order to steer earlier toolchains onto the libopenssl backend no longer apply.

Detect the compiler version with "go.real env GOVERSION" and, from 1.26 onwards, set GOFIPS140=v1.26.0 and inject the companion no_openssl build tag rather than stripping it out. Since cgo is only needed to reach libopenssl, stop forcing CGO_ENABLED=1; and since forced cgo is what makes static linking fail, stop rewriting -extldflags "-static" to "-lc" as well.

GOEXPERIMENT and the strictfipsruntime build tag are retained. Those are orthogonal to the choice of backend: strictfipsruntime makes a binary die at startup on a FIPS host when it was not built against a compliant backend, whichever backend that is.

Add GO_COMPLIANCE_BUILTIN_FIPS_INCLUDE and its EXCLUDE counterpart for components that cannot use the built-in module yet. Setting the latter to '.*' reverts to the libopenssl mechanism everywhere. They can only ever turn the built-in module off, since there is no way to conjure one out of an older toolchain.

A GOVERSION that cannot be parsed falls back to libopenssl.

## Summary by CodeRabbit

* **New Features**
  * Added improved handling for selecting and configuring Go FIPS modes across supported builds.
  * Added diagnostics that report the active FIPS configuration and related Go settings.

* **Bug Fixes**
  * Improved compatibility between built-in FIPS mode, strict FIPS builds, and OpenSSL-based builds.
  * Corrected build-tag handling to prevent conflicting FIPS options from being applied.


doozer/doozerlib/golang_builder_FIPS_wrapper.sh
```diff
@@ -117,10 +138,50 @@ fi
...
+GO_VER="${FOUND_GO_VERSION#go}"   # "go1.26.0" -> "1.26.0"
+GO_MAJOR="${GO_VER%%.*}"          #            -> "1"
+GO_REST="${GO_VER#*.}"            #            -> "26.0"
+GO_MINOR="${GO_REST%%.*}"         #            -> "26"
+# GOVERSION can carry a suffix (e.g. "go1.24.6 X:strictfipsruntime") which the
+# parsing above tolerates, or be a devel string with no version at all. Only
+# trust the result when both fields came out as numbers; anything else falls
+# back to the openssl behaviour, which is the safe default.
+if [[ "${GO_MAJOR}" =~ ^[0-9]+$ && "${GO_MINOR}" =~ ^[0-9]+$ ]]; then
+  if [[ "${GO_MAJOR}" -gt 1 ]] || [[ "${GO_MAJOR}" -eq 1 && "${GO_MINOR}" -ge 26 ]]; then
```
zaneb: The 5.0 builder image also contains 1.26, so we need to also find another way to limit this to 5.1 and onward for now.
/hold

PR #3417, opened 2026-09-16
fix(lockfile): ignore repository-only SBOM RPM entries

fgallott:
rh-pre-commit.version: 2.4.0
rh-pre-commit.check-secrets: ENABLED

## Summary by CodeRabbit

- **Bug Fixes**
  - SBOM RPM parsing now recognizes packages with upstream metadata or both checksum and repository metadata.
  - RPM entries lacking supported package metadata continue to be excluded as repository metadata.
  - Valid installed RPM packages continue to appear correctly in generated package lists.

- **Tests**
  - Added coverage for supported and unsupported RPM metadata combinations.
  - Updated SPDX attestation and attachment fixtures with upstream package metadata.

PR #3365, opened 2026-09-08
ART-23302: improve RPM lockfile generation using SBOM

fgallott:
## Summary

  Improve RPM lockfile generation by using image SBOM data with architecture-specific package handling and reliable RPMDB fallbacks.

  ## Changes

  - Query installed packages per target architecture.
  - Scope discovered packages with `ArchSpecificPackage` to prevent cross-architecture leakage.
  - Prefer SPDX SBOM attestations.
  - Support legacy SPDX SBOM attachments through ORAS.
  - Fall back to extracting the image RPMDB with `oc image extract` and querying it with host-side `rpm`.
  - Remove the `podman run` dependency from package discovery.
  - Support both `src` and `nosrc` source RPM architectures.
  - Continue filtering RPM pseudo-packages such as `gpg-pubkey`.
  - Fall back to parent lockfile data when package metadata is unavailable.

  ## Testing

  - Added coverage for:
    - SPDX attestations
    - ORAS legacy SBOM attachments
    - Multi-architecture package resolution
    - RPMDB extraction fallback
    - Legacy RPMDB paths
    - `src` and `nosrc` package filtering
  - `231` lockfile prototype tests passed.
  - Ruff checks and formatting passed.

## Summary by CodeRabbit

* **New Features**
  * Lockfile generation now discovers installed packages separately for each target architecture.
  * Package specifications retain architecture information for more accurate upgrades, filtering, deduplication, and conflict detection.
  * Image file extraction supports individual files, glob patterns, directories, and optional registry authentication.
  * Package discovery first uses architecture-specific SBOM data and can fall back to extracted package databases.

* **Bug Fixes**
  * Package-query failures and empty or failed image extractions now safely return no packages instead of interrupting processing.

PR #3443, opened 2026-09-19
fix(scan-sources): add concurrency limits to prevent buildvm OOM

ashwindasr:
## Summary

- buildvm has been crashing repeatedly with OOM since ~Sep 16 due to `doozer beta:config:konflux:scan-sources` processes consuming 7-8 GB RSS each
- Root cause: `scan_images()` runs unbounded `asyncio.gather` over all images at a dependency tree level (200+ for OCP), each spawning cosign subprocesses. Combined with 27 groups now in `nonOCPGroups` (7 added Sep 10-17), concurrent memory usage exceeds the 256 GB buildvm limit
- PR #3365 was wrongly blamed and reverted — the real issue is the unbounded concurrency in scan-sources, not SBOM lockfile generation

## Changes

- **`scan_sources_konflux.py`**: Add `Semaphore(16)` to limit concurrent `scan_image()` calls per dependency tree level
- **`util.py`**: Add `@limit_concurrency(16)` to `get_konflux_data()` to limit concurrent `cosign download attestation` subprocesses

## Impact

Each scan-sources instance should drop from 7-8 GB peak RSS to ~1-2 GB by processing at most 16 images concurrently instead of 200+. This makes the system safe even with 27+ concurrent scan-sources instances.

## Test plan

- [x] All 76 scan-sources unit tests pass
- [x] All 93 artcommon/util tests pass
- [x] Lint passes
- [ ] After merge: monitor buildvm memory graph during next scan cycle — expect peak memory to stay well under 256 GB
- [ ] After stabilized: re-land PR #3365 and #3417 (reverted for the wrong reason)

## Summary by CodeRabbit

- **Bug Fixes**
  - Scan operations now apply a consistent limit to simultaneous image scans.
  - Retrieving Konflux data and scanning images are better coordinated under heavy workloads.
  - This helps reduce resource contention and supports more stable performance during large scan operations.


doozer/doozerlib/cli/scan_sources_konflux.py
```diff
@@ -575,9 +575,15 @@ async def scan_images(self, image_names: List[str]):
         # Store latest build records in a map, to reduce DB queries and execution time
         await self.find_latest_image_builds(image_names)
 
-        # Scan images for changes
+        # Scan images for changes, limiting concurrency to avoid OOM on buildvm
         scanning_image_metas = [self.runtime.image_map[image_name] for image_name in image_names]
-        await asyncio.gather(*[self.scan_image(image_meta) for image_meta in scanning_image_metas])
+        semaphore = asyncio.Semaphore(16)
```
pruan-rht: @ashwindasr can we put that into a constant with name to reference instead and share it between util.py?

